PR #2, opened 2016-01-20
Redis

wpride:


src/main/java/application/WebAppContext.java
```diff
@@ -0,0 +1,100 @@
...
+import services.XFormService;
+import services.impl.XFormServiceImpl;
+
+import java.util.Properties;
+
+@Configuration
+@EnableWebMvc
+@ComponentScan(basePackages = {
+        "net.petrikainulainen.spring.testmvc.common.controller",
+        "net.petrikainulainen.spring.testmvc.todo.controller"
```
snopoke: pretty sure these shouldn't be here

src/main/java/application/WebAppContext.java
```diff
@@ -0,0 +1,100 @@
...
+        configurer.enable();
+    }
+
+    @Bean
+    public SimpleMappingExceptionResolver exceptionResolver() {
+        SimpleMappingExceptionResolver exceptionResolver = new SimpleMappingExceptionResolver();
+
+        Properties exceptionMappings = new Properties();
+
+        exceptionMappings.put("net.petrikainulainen.spring.testmvc.todo.exception.TodoNotFoundException", "error/404");
```
snopoke: this too doesn't belong

src/main/java/application/WebAppContext.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Bean
+    public RedisTemplate redisTemplate(){
+        RedisTemplate redisTemplate =  new RedisTemplate();
+        redisTemplate.setConnectionFactory(jedisConnFactory());
+        return redisTemplate;
+    }
+
+    @Bean
+    public SessionRepo sessionRepo(){
```
snopoke: some of these are duplicated in `Application.java`

src/main/java/beans/NewSessionBean.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+    @JsonSetter(value = "form-url")
+    public void setFormUrl(String formUrl) {
+        this.formUrl = formUrl;
+    }
+    @JsonGetter(value = "hq_auth")
+    public Map<String, String> getHqAuth() {
+        return hqAuth;
+    }
+    @JsonSetter(value = "hq_auth")
```
snopoke: Why do some have `-` and some have `_`

src/main/java/repo/impl/SessionImpl.java
```diff
@@ -0,0 +1,47 @@
...
+import repo.SessionRepo;
+
+import java.util.Map;
+
+/**
+ * Created by willpride on 1/19/16.
+ */
+public class SessionImpl implements SessionRepo{
+
+    private RedisTemplate<String, SerializableSession> redisTemplate;
```
snopoke: Could use `@Autowired` instead of the setter & getter approach

src/main/java/beans/NewSessionBean.java
wpride: Seeing only hq_auth

snopoke: `form-url`

wpride: Ah. That's just what touchforms is doing and I'm replicating at the moment. Should be changed down the road

snopoke: 👍

PR #5, opened 2016-01-21
Cleanup

wpride:


src/main/java/application/Application.java
```diff
@@ -20,7 +24,7 @@
 @EnableWebMvc
 public class Application {
 
-    public static final String HOST = "localhost:8080";
+    public static final String HOST = "localhost:8000";
```
benrudolph: should this be 8080? in the futuer this probably be configurable

benrudolph: actually it needs to be for this to work

src/main/java/services/impl/RestoreServiceImpl.java
```diff
@@ -13,6 +13,8 @@
 public class RestoreServiceImpl implements RestoreService {
     @Override
     public String getRestoreXml(String host, String domain, HqAuth auth) {
+        System.out.println("restoring from: " +"http://" + host
```
benrudolph: print. there's gotta be a way to do real logging in java

PR #6, opened 2016-01-21
Update build.gradle to do 1.7

benrudolph:
just doing this in a separate PR so it's not dependent on the other stuff and staging won't break on deploy


benrudolph:
@wpride

PR #7, opened 2016-01-22
current command and tests

wpride:


src/main/java/beans/CurrentRequestBean.java
```diff
@@ -0,0 +1,43 @@
...
+        return formContext;
+    }
+    @JsonSetter(value = "form_context")
+    public void setFormContext(Map<String, String> formContext) {
+        this.formContext = formContext;
+    }
+
+    @Override
+    public String toString(){
+        return "Answer Question Bean [formContente: " + formContext + ", sessionId: " + sessionId + "]";
```
benrudolph: typo formContente

src/main/java/beans/CurrentResponseBean.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Created by willpride on 1/20/16.
+ */
+public class CurrentResponseBean {
+    private JSONArray tree;
+    private String title;
+    private String[] langs;
+    private String sessionId;
+
+    public CurrentResponseBean(){
```
benrudolph: any reason for this guy?

src/main/java/beans/CurrentResponseBean.java
```diff
@@ -0,0 +1,67 @@
...
+        langs = session.getLanguages();
+        sessionId = session.getUUID();
+    }
+
+
+    @JsonGetter(value = "session_id")
+    public String getSessionId() {
+        return sessionId;
+    }
+    @JsonSetter(value = "session-_d")
```
benrudolph: `session_id`

src/test/java/tests/FormEntryTest.java
```diff
@@ -125,5 +127,18 @@ public void testFormEntry() throws Exception {
...
         System.out.println("result3: " + response);
+
+        ObjectMapper mapper = new ObjectMapper();
+
+        CurrentRequestBean currentRequestBean = mapper.readValue
+                (FileUtils.getFile(this.getClass(), "requests/current/current_request.json"), CurrentRequestBean.class);
+        currentRequestBean.setSessionId(sessionId);
+
+        //Test Current Session
+        ResultActions currentResult = mockMvc.perform(get("/current")
```
benrudolph: not sure how spring does URL routing, but wondering if there's a way to get URLs that doesn't involve typing out the hard coded string

src/test/java/tests/FormEntryTest.java
```diff
@@ -125,5 +127,18 @@ public void testFormEntry() throws Exception {
...
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(mapper.writeValueAsString(currentRequestBean)));
+        String currentResultString = currentResult.andReturn().getResponse().getContentAsString();
+        System.out.println("Current Result: " + currentResultString);
```
benrudolph: print

src/main/java/beans/CurrentResponseBean.java
wpride: The library that handles mapping these objects to and from JSON (Jackson) requires a default constructor

wpride:
@benrudolph PR updated

src/main/java/application/SessionController.java
```diff
@@ -63,21 +76,46 @@ public SerializableSession getSession(@RequestParam(value="id") String id) {
...
+            JSONObject resp = AnswerQuestionJson.questionAnswerToJson(formEntrySession.getFormEntryController(),
+                    formEntrySession.getFormEntryModel(),
+                    answerQuestionBean.getAnswer(),
+                    answerQuestionBean.getFormIndex());
+            session.setFormXml(formEntrySession.getFormXml());
+            session.setInstanceXml(formEntrySession.getInstanceXml());
+            sessionRepo.save(session);
+            AnswerQuestionResponseBean responseBean = mapper.readValue(resp.toString(), AnswerQuestionResponseBean.class);
+            return responseBean;
+        } catch (JsonParseException e) {
```
benrudolph: i think this can be cut down to `JsonParseException|JsonMappingException|IOException e`. not blocking cause i think we're going to have a larger discussion about errors. in general we want to catch the errors we can handle/know how to better respond to them and do that. otherwise we'll let it fail hard and then have a handler return any unhandled exception (let's offline that). right now this just ignores some pretty severe errors and logs it to a file. realize my initial comment might've been misleading

benrudolph:
:boom:

PR #16, opened 2016-01-27
Repeat commands

wpride:


wpride:
@benrudolph last of the commands. don't merge

src/main/java/application/CaseController.java
```diff
@@ -29,4 +31,14 @@ public CaseFilterResponseBean filterCasesHQ(@RequestBody String body) throws Exc
         String caseResponse = CaseAPIs.filterCases(filterRequest);
         return new CaseFilterResponseBean(caseResponse);
     }
+
+    @RequestMapping("/sync_db")
+    public SyncDbResponseBean syncUserDb(@RequestBody String body) throws Exception {
+        ObjectMapper mapper = new ObjectMapper();
+        SyncDbRequestBean syncRequest = mapper.readValue(body, SyncDbRequestBean.class);
```
snopoke: You shouldn't have to do this yourself, Spring can do it for you if you've registered the [MappingJacksonHttpMessageConverter](http://docs.spring.io/spring/docs/3.0.x/javadoc-api/org/springframework/http/converter/json/MappingJacksonHttpMessageConverter.html) which I think you have via the [@EnableWebMvc](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html#mvc-config-enable) annotation:

```java
@RequestMapping("/sync_db")
public SyncDbResponseBean syncUserDb(@RequestBody SyncDbRequestBean syncRequest) throws Exception {
    syncRequest.setRestoreService(restoreService);
    .....
}
```

snopoke: Same applies to all the other places too.

src/main/java/application/SessionController.java
```diff
@@ -51,6 +56,8 @@ public NewSessionResponse newFormResponse(@RequestBody String body){
             log.error(e);
         } catch (IOException e) {
             log.error(e);
+        } catch (Exception e) {
+            e.printStackTrace();
```
snopoke: If you're going to just catch all exceptions you may as well get rid of the other 3 catch blocks. If you don't catch the exception Spring will return an error response to the client. We can also customize the error responses: http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-error-handling

src/main/java/application/SessionController.java
```diff
@@ -93,10 +100,16 @@ public AnswerQuestionResponseBean answerQuestion(@RequestBody String body){
...
             log.error(e);
         } catch (JsonMappingException e) {
+            System.out.println("ME: " + e);
+            e.printStackTrace();
             log.error(e);
         } catch (IOException e) {
+            System.out.println("ME: " + e);
+            e.printStackTrace();
             log.error(e);
```
snopoke: If you're already logging why do you need the printlines and the printStackTrace?

src/main/java/beans/DeleteRepeatResponseBean.java
```diff
@@ -0,0 +1,43 @@
...
+        return sequenceId;
+    }
+    @JsonSetter(value = "seq_id")
+    public void setSequenceId(String sequenceId) {
+        this.sequenceId = sequenceId;
+    }
+
+    @Override
+    public String toString(){
+        return "NewRepeatResponseBean: [tree=" + Arrays.toString(tree) + ", seq_id: " + sequenceId + "]";
```
snopoke: should be `DeleteRepeatResponseBean`

src/main/java/application/SessionController.java
```diff
@@ -154,4 +169,63 @@ public EvaluateXPathResponseBean evaluateXpath(@RequestBody String body) {
...
+                    formEntrySession.getFormEntryModel(),
+                    newRepeatRequestBean.getFormIndex());
+
+            serializableSession.setFormXml(formEntrySession.getFormXml());
+            serializableSession.setInstanceXml(formEntrySession.getInstanceXml());
+            sessionRepo.save(serializableSession);
+
+            JSONObject respo =  AnswerQuestionJson.getCurrentJson(formEntrySession.getFormEntryController(),
+                    formEntrySession.getFormEntryModel());
+            return mapper.readValue(respo.toString(), NewRepeatResponseBean.class);
```
snopoke: Can you not just return the `JSONObject` directly?

src/main/java/application/SessionController.java
```diff
@@ -154,4 +169,63 @@ public EvaluateXPathResponseBean evaluateXpath(@RequestBody String body) {
...
+    @RequestMapping(value = "/new_repeat", method = RequestMethod.GET)
+    @ResponseBody
+    public NewRepeatResponseBean newRepeat(@RequestBody String body) {
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            NewRepeatRequestBean newRepeatRequestBean = mapper.readValue(body, NewRepeatRequestBean.class);
+            SerializableSession serializableSession = sessionRepo.find(newRepeatRequestBean.getSessionId());
+            FormEntrySession formEntrySession = new FormEntrySession(serializableSession);
+
+            JSONObject resp = AnswerQuestionJson.descendRepeatToJson(formEntrySession.getFormEntryController(),
```
snopoke: `resp` unused

src/main/java/application/SessionController.java
```diff
@@ -154,4 +169,63 @@ public EvaluateXPathResponseBean evaluateXpath(@RequestBody String body) {
...
+    public DeleteRepeatResponseBean delete_repeat(@RequestBody String body) {
+        try {
+            System.out.println("Delete Repeat");
+            ObjectMapper mapper = new ObjectMapper();
+
+            DeleteRepeatRequestBean deleteRepeatRequestBean = mapper.readValue(body, DeleteRepeatRequestBean.class);
+            SerializableSession serializableSession = sessionRepo.find(deleteRepeatRequestBean.getSessionId());
+            FormEntrySession formEntrySession = new FormEntrySession(serializableSession);
+
+            JSONObject resp = AnswerQuestionJson.deleteRepeatToJson(formEntrySession.getFormEntryController(),
```
snopoke: `resp` unused

src/main/java/beans/NewRepeatRequestBean.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonSetter;
+
+/**
+ * Created by willpride on 1/20/16.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class NewRepeatRequestBean {
```
snopoke: Seems identical to `DeleteRepeatRequestBean`, why not just have a `RepeatRequestBean`?

snopoke: Same for `DeleteRepeatResponseBean` and `NewRepeatResponseBean`

src/main/java/application/CaseController.java
wpride: :bomb: Damn! Thanks Simon :bomb:

src/main/java/application/SessionController.java
wpride: agreed, this was cruft from testing, my bad

wpride:
@snopoke thanks for the comments. My thinking with returning DeleteRepeatResponseBean instead of JSONObject and having redundant-ish beans like delete and new repeat is just to be more explicit about what each function is returning and what each request should have. Happy to be overruled if we think this isn't worth it

snopoke:
Re the redundant beans I think that if they are identical then it isn't worth the extra code. Semantically I think it makes sense to say that these 2 endpoints respond with / accept the same type (especially since they operate on the same 'entity').

snopoke:
Build failing due to bad submodule ref

wpride:
@snopoke sorry  about that, pushed sub repo

wpride:
@snopoke @benrudolph Comments addressed, I'm still refactoring the tests a bit but this is good to review

src/main/java/application/SessionController.java
```diff
@@ -142,16 +114,62 @@ public GetInstanceResponseBean getInstance(@RequestBody String body) throws IOEx
...
+        serializableSession.setInstanceXml(formEntrySession.getInstanceXml());
+        sessionRepo.save(serializableSession);
+        JSONObject response =  AnswerQuestionJson.getCurrentJson(formEntrySession.getFormEntryController(),
+                formEntrySession.getFormEntryModel());
+        return mapper.readValue(response.toString(), RepeatResponseBean.class);
+    }
+
+    @RequestMapping(value = "/delete_repeat", method = RequestMethod.GET)
+    @ResponseBody
+    public RepeatResponseBean delete_repeat(@RequestBody RepeatRequestBean repeatRequestBean) throws Exception {
```
snopoke: camel case

src/main/java/beans/EvaluateXPathResponseBean.java
```diff
@@ -17,14 +17,13 @@
...
     private String status;
 
-    public EvaluateXPathResponseBean(){
-
-    }
+    //Jackson requires the default constructor
+    public EvaluateXPathResponseBean(){}
 
     public EvaluateXPathResponseBean(FormEntrySession formEntrySession, String xpath) throws IOException, XPathSyntaxException {
+        // TODO: this
```
snopoke: ?

src/main/java/hq/RestoreUtils.java
```diff
@@ -25,11 +34,60 @@ public static UserSqlSandbox restoreUser(String username, String restorePayload)
...
+        TreeElement templateRoot =
+                formDef.getInstance().getRoot().deepCopy(true);
+
+        entryModel = new FormEntryModel(formDef);
+
+        // weak check for matching forms
+        if (!savedRoot.getName().equals(templateRoot.getName()) ||
+                savedRoot.getMult() != 0) {
+            System.out.println("Instance and template form definition don't match");
+            return null;
```
snopoke: Should maybe raise an exception instead since this is an error condition

snopoke:
didn't review very closely but generally looks OK

src/main/java/application/SessionController.java
```diff
@@ -142,16 +114,62 @@ public GetInstanceResponseBean getInstance(@RequestBody String body) throws IOEx
...
+        sessionRepo.save(serializableSession);
+
+        JSONObject response =  AnswerQuestionJson.getCurrentJson(formEntrySession.getFormEntryController(),
+                formEntrySession.getFormEntryModel());
+
+        return mapper.readValue(response.toString(), RepeatResponseBean.class);
+    }
+
+    @RequestMapping("/filter_cases")
+    public CaseFilterResponseBean filterCasesHQ(@RequestBody CaseFilterRequestBean filterRequest) throws Exception {
```
benrudolph: is this for backwards compatibility? or do we still need to filter cases sent from HQ?

src/main/java/beans/EvaluateXPathResponseBean.java
```diff
@@ -17,14 +17,13 @@
...
         status = "success";
```
benrudolph: is this used?

src/main/java/hq/RestoreUtils.java
benrudolph: +1

benrudolph:
i think OK to merge since this PR is getting really big. maybe just follow up on the comments later

src/main/java/application/SessionController.java
wpride: This is needed for now, won't be once we have installs figured out

PR #18, opened 2016-02-18
Installs

wpride:
WIP


src/main/java/application/SessionController.java
```diff
@@ -84,74 +95,87 @@ public AnswerQuestionResponseBean answerQuestion(@RequestBody AnswerQuestionRequ
...
     public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitRequestBean) throws Exception {
-        SerializableSession serializableSession = sessionRepo.find(submitRequestBean.getSessionId());
-        FormEntrySession formEntrySession = new FormEntrySession(serializableSession);
+        log.info("Submit form with bean: " + submitRequestBean);
+        SerializableFormSession serializableFormSession = sessionRepo.find(submitRequestBean.getSessionId());
+        FormEntrySession formEntrySession = new FormEntrySession(serializableFormSession);
         FormRecordProcessorHelper.processXML(formEntrySession.getSandbox(), formEntrySession.submitGetXml());
-        return new SubmitResponseBean(formEntrySession);
+        SubmitResponseBean submitResponseBean = new SubmitResponseBean(formEntrySession);
+        log.info("Submit response bean: " + submitResponseBean);
```
snopoke: This kind of logging can be done with a Servlet Filter:
* http://stackoverflow.com/questions/2158647/logging-response-body-html-from-httpservletresponse-using-spring-mvc-handlerin
* https://mdeinum.wordpress.com/2015/07/01/spring-framework-hidden-gems/

src/main/java/hq/CaseAPIs.java
```diff
@@ -51,4 +57,28 @@ public static String filterCases(CaseFilterRequestBean request) throws Exception
         }
         return "null";
     }
+
+    public static CaseBean getFullCase(String caseId, String username){
+        UserSqlSandbox mSandbox = new UserSqlSandbox(username);
+        SqliteIndexedStorageUtility<Case> caseStorage = mSandbox.getCaseStorage();
+        Case cCase = caseStorage.getRecordForValue("case_id", caseId);
+        CaseBean ret = new CaseBean(cCase);
+        return ret;
```
snopoke: Seems like quite a lot of work to be doing inside a tight loop especially since the 1st 2 lines are always going to be the same for each loop iteration (in `filterCasesFull`)

wpride: agreed, that was silly. thanks

src/main/java/application/SessionController.java
wpride: Interesting stuff. Looked into this today and added a filter to get some basic request/response information. Unfortunately it seems like reading the request and response body (what we're really interested in) is more difficult since these are only meant to be read once. So people suggest building a wrapper around this that basically lets you read the input stream and make a copy. This seems like a lot of overhead for logging - especially since we're already doing this with our POJO's which we have access to at the Controller level (which is what a lot of people recommend doing). Not sure if I'm missing some other method here?

build.gradle
```diff
@@ -57,4 +58,6 @@ dependencies {
 
 task wrapper(type: Wrapper) {
     gradleVersion = '2.3'
+    def currentJvm = org.gradle.internal.jvm.Jvm.current()
+    println currentJvm
```
benrudolph: print

src/main/java/application/Application.java
```diff
@@ -27,7 +24,37 @@
...
+    @Bean
+    public CommonsRequestLoggingFilter requestLoggingFilter() {
+        CommonsRequestLoggingFilter crlf = new CommonsRequestLoggingFilter();
+        crlf.setIncludeClientInfo(true);
+        crlf.setIncludeQueryString(true);
+        crlf.setIncludePayload(true);
+        return crlf;
+    }
+
+    public class ResponseLoggingFilter extends GenericFilterBean {
```
benrudolph: how does this work?

src/main/java/application/SessionController.java
```diff
@@ -166,11 +200,101 @@ public CaseFilterResponseBean filterCasesHQ(@RequestBody CaseFilterRequestBean f
...
+    }
+
+    @RequestMapping(Constants.URL_MENU_SELECT)
+    public SessionBean selectMenu(@RequestBody MenuSelectBean menuSelectBean) throws Exception {
+        MenuSession menuSession = new MenuSession(menuRepo.find(menuSelectBean.getSessionId()), restoreService, installService);
+        boolean redrawing = menuSession.handleInput(menuSelectBean.getSelection());
+        menuRepo.save(menuSession.serialize());
+        Screen nextScreen;
+
+        if(!redrawing){
```
benrudolph: nit: weird spacing

src/main/java/application/SessionController.java
```diff
@@ -166,11 +200,101 @@ public CaseFilterResponseBean filterCasesHQ(@RequestBody CaseFilterRequestBean f
...
+                String[] rows = entityListSubscreen.getRows();
+                HashMap<Integer, String> optionsStrings = new HashMap<Integer, String>();
+                for(int i=0; i <rows.length; i++){
+                    optionsStrings.put(i, rows[i]);
+                }
+                MenuResponseBean menuResponseBean = new MenuResponseBean();
+                menuResponseBean.setMenuType(Constants.MENU_ENTITY);
+                menuResponseBean.setOptions(optionsStrings);
+                menuResponseBean.setSessionId(menuSession.getSessionId());
+                return menuResponseBean;
```
benrudolph: the above 8 lines seem like they can be abstracted into something like `getMenuResponse(entityListRows, sessionId, menuType)`

src/main/java/application/SessionController.java
```diff
@@ -166,11 +200,101 @@ public CaseFilterResponseBean filterCasesHQ(@RequestBody CaseFilterRequestBean f
...
+    @ExceptionHandler(Exception.class)
+    public String handleError(HttpServletRequest req, Exception exception) {
+        log.error("Request: " + req.getRequestURL() + " raised " + exception);
+        exception.printStackTrace();
+
+        JSONObject errorReturn = new JSONObject();
+        errorReturn.put("exception", exception);
+        errorReturn.put("url", req.getRequestURL());
+        errorReturn.put("status", "error");
+        return errorReturn.toString();
```
benrudolph: we should also look into email handling so we get errors sent to us: http://docs.spring.io/spring-integration/reference/html/mail.html

src/main/java/application/WebAppContext.java
```diff
@@ -89,15 +92,25 @@ public SessionRepo sessionRepo(){
     }
 
     @Bean
+    public MenuRepo menuRepo(){
+        MenuImpl impl = new MenuImpl();
+        return impl;
```
benrudolph: `return new MenuImpl()`

benrudolph: maybe you can't do that in Java...

src/main/java/beans/AnswerQuestionRequestBean.java
```diff
@@ -31,21 +28,13 @@ public void setFormIndex(String formIndex) {
         this.formIndex = formIndex;
     }
     @JsonGetter(value = "answer")
-    public String getAnswer() {
+    public Object getAnswer() {
```
benrudolph: this is either an array or a string right? may want to make a comment or a class specifying

src/main/java/beans/EvaluateXPathResponseBean.java
```diff
@@ -22,7 +20,7 @@ public EvaluateXPathResponseBean(){}
 
     public EvaluateXPathResponseBean(FormEntrySession formEntrySession, String xpath) throws IOException, XPathSyntaxException {
         // TODO: don't always return success
-        status = "success";
+        status = "accepted";
```
benrudolph: what does status actually do in this function? i don't see it referenced

src/main/java/beans/NewSessionResponse.java
```diff
@@ -1,30 +1,32 @@
...
-import java.util.ArrayList;
+import java.io.IOException;
+import java.util.Arrays;
 
 /**
  * Created by willpride on 1/12/16.
  */
-public class NewSessionResponse {
+public class NewSessionResponse extends SessionBean{
     // TODO: This should be a QuestionBean array
```
benrudolph: remove TODO

src/main/java/beans/QuestionBean.java
```diff
@@ -108,11 +113,14 @@ public void setDatatype(String datatype) {
         this.datatype = datatype;
     }
 
-    public StyleBean getStyle() {
-        return style;
+    public HashMap<String, String> getStyle() {
+        if(style != null) {
+            return style;
+        }
+        return new HashMap<String, String>();
```
benrudolph: should `style` just be initialized as `new HashMap<String, String>()` so we can just `return style`

src/main/java/beans/SubmitResponseBean.java
```diff
@@ -53,6 +38,6 @@ public void setStatus(String status) {
 
     @Override
     public String toString(){
-        return "CurrentResponseBean: [sessionId=" + sessionId + "]";
+        return "SubmitResponseBean: [sessionId=" + sessionId + ",output= " + output + "]";
```
benrudolph: seems like these `toString` functions could be some sort of utility where you pass the class and args and then generate a toString()

src/main/java/beans/SyncDbResponseBean.java
```diff
@@ -9,7 +7,7 @@
     private String status;
 
     public String getStatus() {
-        return "success";
+        return "accepted";
```
benrudolph: these should be constants somewhere

src/main/java/install/FormplayerConfigEngine.java
```diff
@@ -0,0 +1,461 @@
...
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Hashtable;
+import java.util.Vector;
+import java.util.zip.ZipFile;
+
+
+/**
+ * @author ctsims
```
benrudolph: :eyes:

benrudolph: is this copied code?

src/main/java/install/FormplayerConfigEngine.java
```diff
@@ -0,0 +1,461 @@
...
+        StorageManager.registerStorage(Profile.STORAGE_KEY, Profile.class);
+        StorageManager.registerStorage(Suite.STORAGE_KEY, Suite.class);
+        StorageManager.registerStorage(FormDef.STORAGE_KEY,FormDef.class);
+        StorageManager.registerStorage(FormInstance.STORAGE_KEY, FormInstance.class);
+    }
+
+    private void setRoots() {
+        ReferenceManager._().addReferenceFactory(new JavaHttpRoot());
+
+        this.mArchiveRoot = new ArchiveFileRoot();
```
benrudolph: i do like the `m*` variable style to denote member variables. would be nice to adopt some sort of convention

src/main/java/install/FormplayerConfigEngine.java
benrudolph: seems like it might be, if it is, is there no way to use it as a library? this means we have to keep both up to date

src/main/java/requests/InstallRequest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public MenuResponseBean getResponse(){
+        MenuResponseBean menuResponseBean = new MenuResponseBean();
+        menuResponseBean.setMenuType(Constants.MENU_MODULE);
+        menuResponseBean.setOptions(parseMenuChoices(this.menuSession.getChoices()));
+        menuResponseBean.setSessionId(this.menuSession.getSessionId());
+        return menuResponseBean;
+    }
+
+    public Map<Integer, String> parseMenuChoices(String[] menuDisplayables){
```
benrudolph: seems like we could use this else where as well

src/main/java/session/FormEntrySession.java
```diff
@@ -1,33 +1,38 @@
...
-import java.io.StringReader;
+import java.io.*;
 import java.util.Map;
 import java.util.UUID;
+import org.apache.commons.codec.binary.Base64;
 
 /**
- * Created by willpride on 1/15/16.
+ *
+ * OK this (and MenuSession) is a total god object that basically amanges everything about the state of
```
benrudolph: amanges

src/main/java/session/FormEntrySession.java
```diff
@@ -1,33 +1,38 @@
...
+ * a form entry session. We turn this into a SerializableFormSession to persist it. Within that we also
+ * serialize the formDef to persist the session, in addition to a bunch of other information like the restoreXml.
+ * Confusingly we also have a SessionWrapper object within this session which tracks a bunch of other information. There
+ * is a lot of unification that needs to happen here.
```
benrudolph: yay for docs!

src/main/java/session/FormEntrySession.java
```diff
@@ -83,8 +90,28 @@ public FormEntrySession(String formXml, String restoreXml, String initLang, Stri
...
+        formEntryModel = new FormEntryModel(formDef, FormEntryModel.REPEAT_STRUCTURE_NON_LINEAR);
+        formEntryController = new FormEntryController(formEntryModel);
+        formEntryController.setLanguage(initLang);
+        title = formDef.getTitle();
+        langs = formEntryModel.getLanguages();
+        this.initLang = initLang;
+        uuid = UUID.randomUUID().toString();
         this.sequenceId = 0;
         initialize(true, sessionData);
+        getFormTree();
```
benrudolph: this seems like a misleading function name given it doesn't return anything

src/main/java/session/FormEntrySession.java
```diff
@@ -93,17 +120,14 @@ public void initialize(boolean newInstance, Map<String, String> sessionData) thr
...
-
     private FormDef parseFormDef(String formXml) throws IOException {
         XFormParser mParser = new XFormParser(new StringReader(formXml));
         return mParser.parse();
     }
 
     public String getInstanceXml() throws IOException {
-        return XmlUtil.getPrettyXml(new XFormSerializingVisitor().serializeInstance(formDef.getInstance()));
+        byte[] bytes = new XFormSerializingVisitor().serializeInstance(formDef.getInstance());
+        return new String(bytes, "US-ASCII");
```
benrudolph: will this work on unicode in the XML?

src/main/java/session/MenuSession.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.UUID;
+
+/**
+ * Created by willpride on 2/5/16.
+ */
+@Component
+public class MenuSession {
+    FormplayerConfigEngine engine;
```
benrudolph: i'm not running this through the linter, but assuming that the ones that are marked non private are meant to be non-private

src/main/java/session/MenuSession.java
```diff
@@ -0,0 +1,224 @@
...
+        boolean redrawing = screen.handleInputAndUpdateSession(sessionWrapper, input);
+        if(screen instanceof EntityScreen){
+            EntityScreen entityScreen = (EntityScreen) screen;
+            if(entityScreen.getCurrentScreen() instanceof EntityDetailSubscreen){
+                EntityDetailSubscreen entityDetailSubscreen = (EntityDetailSubscreen) entityScreen.getCurrentScreen();
+                int currentIndex = entityDetailSubscreen.getCurrentIndex();
+                currentSelection = input;
+            }
+        }
+        return redrawing;
```
benrudolph: does redrawing seems kind of tertiary to what this function is doing. i feel like it could be 2 functions. one that gets redrawing and one that handles input (could be wrong though, you have more context)

src/main/java/session/MenuSession.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+    public Screen getNextScreen() throws CommCareSessionException {
+        String next = sessionWrapper.getNeededData();
+
+        if (next == null) {
+            //XFORM TIME!
```
benrudolph: lol

src/main/java/session/MenuSession.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    private void computeDatum() {
+        //compute
+        SessionDatum datum = sessionWrapper.getNeededDatum();
+        XPathExpression form;
+        try {
+            form = XPathParseTool.parseXPath(datum.getValue());
+        } catch (XPathSyntaxException e) {
+            //TODO: What.
```
benrudolph: remove

benrudolph: or make more meaningful

src/main/java/session/MenuSession.java
```diff
@@ -0,0 +1,224 @@
...
+            e.printStackTrace();
+            throw new RuntimeException(e.getMessage());
+        }
+        EvaluationContext ec = sessionWrapper.getEvaluationContext();
+        if (datum.getType() == SessionDatum.DATUM_TYPE_FORM) {
+            sessionWrapper.setXmlns(XPathFuncExpr.toString(form.eval(ec)));
+            sessionWrapper.setDatum("", "awful");
```
benrudolph: awful?

benrudolph:
OK, gave it a go. This is monstrous. None of my comments I think are blocking and this PR is getting almost too big to review. I think it makes sense to merge now. Going forward it'd be great to have follow up the comments made. I also think it could use some more documentation in general. There aren't too many that are familiar with Java so it'll be especially helpful to have good docs for web devs to come in and contribute. Do you feel like after that we're at a good place for a design review and hash out the next steps that have to be made?

super legit :smoking: :fire:

wpride:
@benrudolph thanks for the review, agree this is a monster. Hold off on merge, I'll address tomorrow and run through the linter too. Moving forward hoping this is the last PR of this size, everything else should be more incremental now that the core functionality is in place

benrudolph:
:+1:

src/main/java/application/SessionController.java
wpride: lol I just refactored this and its so much better, great call

src/main/java/beans/AnswerQuestionRequestBean.java
wpride: Can also be an int or a Date, but agreed

src/main/java/beans/EvaluateXPathResponseBean.java
wpride: Everything with a getter in this class will be returned as a JSON property

src/main/java/beans/NewSessionResponse.java
wpride: love deleting TODOs

src/main/java/install/FormplayerConfigEngine.java
wpride: Yeah I have to merge a few classes together for sure. Unfortunately a lot of this code is paired with the CLI code pretty heavily atm

src/main/java/session/FormEntrySession.java
wpride: how to test?

benrudolph: add a unicode character to a form instance. so you could potentially answer a question with something here: http://unicode-table.com/en/

benrudolph:
:boom: nice.

src/main/java/session/FormEntrySession.java
wpride: returns the form tree actually haha, I just need to do this here also to make sure the repeats are expanded.

wpride:
@benrudolph think this is gtm

benrudolph:
:tada: :tada:

PR #20, opened 2016-03-07
manually deregister JDBC drivers

wpride:
@benrudolph this also should help with memory leaks

### DIFF
--- a/src/main/java/application/WebAppContext.java
+++ b/src/main/java/application/WebAppContext.java
@@ -1,5 +1,7 @@
 package application;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


benrudolph:
awesome :non-potable_water:

snopoke:
A bit confused by this - why do we need to de-register the drivers? Can't we just not include them in the first place?

wpride:
@snopoke we use the JDBC drivers to interact with the sqlite dbs so the class loader will register these classes, not sure how we'd get around that?

snopoke:
Ah right - I missed one key detail: `@PreDestroy`. Still seems a bit strange that we'd need to do this since we aren't running in a container. But I'll assume its right.

PR #21, opened 2016-03-09
form session locking

wpride:
@benrudolph think this is all we need

### DIFF
--- a/src/main/java/application/SessionController.java
+++ b/src/main/java/application/SessionController.java
@@ -8,19 +8,19 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.commcare.api.json.AnswerQuestionJson;
 import org.commcare.modern.process.FormRecordProcessorHelper;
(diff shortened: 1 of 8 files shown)


src/main/java/application/SessionController.java
```diff
@@ -274,6 +288,12 @@ public String handleError(HttpServletRequest req, Exception exception) {
         return errorReturn.toString();
     }
 
+    public Lock getLockAndLock(SessionBean sessionBean){
```
benrudolph: nice this makes sense to lock on session id. not sure if java has a decorator concept but this would be a perfect decorator. or maybe we make a response subclass that automatically does this

benrudolph:
test failure looks real:
```
tests.CaseTests > testCases STANDARD_ERROR
    java.lang.NullPointerException
    	at application.SessionController.getLockAndLock(SessionController.java:293)
    	at application.SessionController.answerQuestion(SessionController.java:75)
```

PR #24, opened 2016-04-06
Fix

wpride:


wpride:
@benrudolph would be good to merge this to get up to date with javarosa and commcare libs

benrudolph:
right on

benrudolph:
holding off on merge cause i'm not sure if that label implies not merging. feel free to self merge if that's not what it implies

PR #26, opened 2016-04-11
use jedis host

wpride:
See https://github.com/dimagi/commcarehq-ansible/pull/377

@benrudolph


config/application.properties.example
```diff
@@ -1,2 +1,3 @@
 # set to the URL of your CCHQ instance
-commcarehq.host=http://localhost:8000
\ No newline at end of file
+commcarehq.host=http://localhost:8000
+redis.hostname=http://localhost:8000
```
benrudolph: a better default would be `localhost`. in the code below you're just setting the hostname (and the default port for redis isn't 8000)

PR #27, opened 2016-04-12
Swagger

wpride:
Add docs

screen shot 2016-04-12 at 4 54 43 pm
screen shot 2016-04-12 at 4 54 36 pm

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,8 @@ dependencies {
     compile("com.fasterxml.jackson.core:jackson-databind:2.4.3")
     compile("org.springframework.data:spring-data-redis:1.5.0.RELEASE")
     compile("javax.servlet:jstl:1.2")
+    compile("io.springfox:springfox-swagger2:2.4.0")
+    compile("io.springfox:springfox-swagger-ui:2.4.0")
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -44,6 +44,8 @@ dependencies {
     compile("com.fasterxml.jackson.core:jackson-databind:2.4.3")
     compile("org.springframework.data:spring-data-redis:1.5.0.RELEASE")
     compile("javax.servlet:jstl:1.2")
+    compile("io.springfox:springfox-swagger2:2.4.0")
+    compile("io.springfox:springfox-swagger-ui:2.4.0")
```
benrudolph: is there a way we can only build this for dev environments so we don't do this on every deploy where we don't need the docs?

benrudolph: Ah nvm this'll probably break the api

benrudolph:
nice, can you add some docs in the readme either explaining how to render these docs or where to view them?

benrudolph:
also looking at this i think it'd be good to make some of the json parameters that we know are required/always will be there non-optional. but that can be for a later PR

wpride:
@benrudolph agreed

PR #42, opened 2016-05-25
Fix repeats

wpride:
Fix for http://manage.dimagi.com/default.asp?227919


src/main/java/beans/RepeatRequestBean.java
```diff
@@ -9,22 +9,31 @@
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class RepeatRequestBean extends SessionBean {
+    private String ix;
```
benrudolph: is using `ix` part of the fix here? i like `formIndex` better -- more understandable

benrudolph: ah i see. how about calling it `index` or something with a comment as to why we have to do this

benrudolph:
👍 tests are broken

PR #50, opened 2016-06-28
Postgres Sessions and Django Auth

wpride:
This PR contains two big pieces of functionality, both related to Postgres

cross request: https://github.com/dimagi/commcare-hq/pull/12265

**Incomplete forms** 
Added functionality for:

- Storing incomplete form sessions in Postgres
- Listing all incomplete form sessions belonging to a particular user
- Selecting an incomplete form session by session id, loading the instance XML into a new form def, and then returning the new session response

**Django Auth**
Added functionality for:

- Filtering each GET and POST request to verify its auth token
- Checking auth token against the Django postgres table
- DAO and DataSource representing this data

**TODO**

- ~~Add Hibernate Postgres migration for incomplete form session table~~


libs/javarosa
```diff
@@ -1 +0,0 @@
-Subproject commit a79412199978a2ab3a86b79877aeef411e8d9fdd
```
benrudolph: was this supposed to be deleted?

src/main/java/repo/FormplayerPostgresConfig.java
```diff
@@ -0,0 +1,29 @@
...
+public class FormplayerPostgresConfig extends TomcatDataSourceConfiguration {
+
+    @Bean(name = "dsFormplayer")
+    public DataSource dataSource() {
+        return super.dataSource();
+    }
+
+    @Bean(name = "jdbcFormplayer")
+    public JdbcTemplate jdbcTemplate(DataSource dsFormplayer) {
+        System.out.println("Returning Formplayer JDBC template: " + this.getUrl());
```
benrudolph: print

src/main/java/repo/impl/PostgresSessionRepo.java
```diff
@@ -0,0 +1,129 @@
...
+
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        ObjectOutputStream oos;
+        try {
+            oos = new ObjectOutputStream(baos);
+            oos.writeObject(session.getSessionData());
+            byte[] sessionDataBytes = baos.toByteArray();
+
+            int sessionCount = this.jdbcTemplate.queryForObject(
+                    replaceTableName("select count(*) from %s where id = ?"), Integer.class, session.getId());
```
benrudolph: what does `Integer.class` do here?

benrudolph: ah the return type

src/main/java/repo/impl/PostgresSessionRepo.java
```diff
@@ -0,0 +1,129 @@
...
+                    "(id, instanceXml, formXml, " +
+                    "restoreXml, username, initLang, sequenceId, " +
+                    "domain, postUrl, sessionData) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)");
+            this.jdbcTemplate.update(query,  new Object[] {session.getId(), session.getInstanceXml(), session.getFormXml(),
+                    session.getRestoreXml(), session.getUsername(), session.getInitLang(), session.getSequenceId(),
+                    session.getDomain(), session.getPostUrl(), sessionDataBytes}, new int[] {
+                    Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.VARCHAR,
+                    Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.BINARY});
+        } catch (IOException e) {
+            e.printStackTrace();
```
benrudolph: we should probably handle this error

benrudolph: or at least fail hard instead of swallowing and continuing as if we successfully saved

src/main/java/repo/impl/PostgresSessionRepo.java
```diff
@@ -0,0 +1,129 @@
...
+            byte[] st = (byte[]) rs.getObject("sessionData");
+
+            if(st != null) {
+                ByteArrayInputStream byteInputStream = new ByteArrayInputStream(st);
+                ObjectInputStream objectInputStream;
+                try {
+                    objectInputStream = new ObjectInputStream(byteInputStream);
+                    Map<String, String> sessionData = (HashMap) objectInputStream.readObject();
+                    session.setSessionData(sessionData);
+                } catch (IOException e) {
```
benrudolph: can you leave a comment as to when we'd see these and why we don't care about them?

src/main/java/session/FormSession.java
```diff
@@ -130,7 +140,7 @@ public FormSession(UserSandbox sandbox, FormDef formDef, String username, String
     private void initialize(boolean newInstance, Map<String, String> sessionData) {
         CommCarePlatform platform = new CommCarePlatform(2, 27);
         FormplayerSessionWrapper sessionWrapper = new FormplayerSessionWrapper(platform, this.sandbox, sessionData);
-        FormplayerConfigEngine.setupStorageManager(username, "dbs");
+        FormplayerConfigEngine.setupStorageManager("test", "dbs/" + domain + "/" + username);
```
benrudolph: is `"test"` correct?

benrudolph: if so it's unintuitive to have test code being run outside of tests, can you leave a comment?

src/main/java/repo/impl/PostgresTokenRepo.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private final Log log = LogFactory.getLog(PostgresTokenRepo.class);
+
+    @Autowired
+    @Qualifier("jdbcHq")
+    private JdbcTemplate jdbcTemplate;
+
+    @Override
+    public boolean isAuthorized(String tokenId) {
+        String sql = String.format("SELECT * FROM %s WHERE session_key = ?",
```
benrudolph: nit, but you could just include the the date check in the where clause instead of checking it after

libs/javarosa
wpride: yes @phillipm !

src/main/java/application/FormController.java
```diff
@@ -102,6 +111,7 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
         FormRecordProcessorHelper.processXML(formEntrySession.getSandbox(), formEntrySession.submitGetXml());
         SubmitResponseBean submitResponseBean = new SubmitResponseBean(formEntrySession);
         log.info("Submit response bean: " + submitResponseBean);
+        sessionRepo.delete(submitRequestBean.getSessionId());
```
snopoke: Why would you delete the session here?

src/main/java/repo/impl/PostgresSessionRepo.java
```diff
@@ -0,0 +1,125 @@
...
+
+        String query = replaceTableName("INSERT into %s " +
+                "(id, instanceXml, formXml, " +
+                "restoreXml, username, initLang, sequenceId, " +
+                "domain, postUrl, sessionData) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)");
+        this.jdbcTemplate.update(query,  new Object[] {session.getId(), session.getInstanceXml(), session.getFormXml(),
+                session.getRestoreXml(), session.getUsername(), session.getInitLang(), session.getSequenceId(),
+                session.getDomain(), session.getPostUrl(), sessionDataBytes}, new int[] {
+                Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.VARCHAR,
+                Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.BINARY});
```
snopoke: Yikes, I'd forgotten how horrible raw SQL in Java is.

snopoke: how does this table get created and maintained?

wpride: 😢 any protips for the raw SQL?

And the table manipulation is the TODO above, going to implement a Hibernate ORM to manage that

src/main/java/repo/impl/PostgresSessionRepo.java
wpride: going to bubble these up - should fail hard on this since loading the form later without session data will probably crash

src/main/java/application/FormController.java
wpride: This is a form entry session, so once the form's submitted we don't need this information anymore. We could keep it around and expire it somehow, but I can't think of a situation when we'd to restore a finished session. Can you @benrudolph

benrudolph: i can't either. though may want to just run a task that cleans up stale sessions. the only thing i could think that could wrong is if we think we've submitted it, but it has actually silently errored and we then delete the session

wpride: @benrudolph @snopoke  Actually now that I think about it fresh, we definitely shouldn't delete here - as currently configured, formplayer (like the old server) returns the submit XML to the browser which itself submits to HQ. There's potential for loss here so we'll definitely want to keep around on formplayer at this point.

Is there any reason that Formplayer shouldn't itself submit the XML to HQ? And then we can safely delete the session once that's confirmed. 

I have a TODO for this here - https://github.com/dimagi/commcare-hq/blob/def10b6b089b81048c3046cf7121f58897787091/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js#L82-L82

wpride: @benrudolph And I spoke about this, going to move submission to happen in Formplayer in a future PR

wpride:
@benrudolph this and his buddy are good to review. ended up removing all Redis cruft in this PR too

PR #51, opened 2016-06-29
Formplayer Case Tiles

wpride:
image

Remake of https://github.com/dimagi/formplayer/pull/47

cross https://github.com/dimagi/commcare-hq/pull/12281


build.gradle
```diff
@@ -81,6 +81,12 @@ dependencies {
     testCompile("com.jayway.jsonpath:json-path-assert:0.8.1")
 }
 
+test {
+    testLogging {
+        exceptionFormat = 'full'
```
wpride: @benrudolph so we'll have the stack trace on Travis

benrudolph: 🙌

src/main/java/beans/menus/Style.java
```diff
@@ -18,7 +18,7 @@ public Style(DetailField detail){
             try {
                 setFontSize(Integer.parseInt(detail.getFontSize()));
             } catch (NumberFormatException nfe) {
-                nfe.printStackTrace();
+                //nfe.printStackTrace();
```
benrudolph: something should be handled here or logging of the stack trace. maybe set a default fontsize?

src/main/java/beans/menus/Tile.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public int getGridHeight() {
+        return gridHeight;
+    }
+
+    public void setGridHeight(int gridHeight) {
+        this.gridHeight = gridHeight;
+    }
+
+    public String getCssId() {
```
benrudolph: it's kind of strange that the java class knows about css

src/test/java/tests/BaseMenuTestClass.java
```diff
@@ -0,0 +1,205 @@
...
+                            engine.initFromLocalFileResource(absolutePath);
+                        } else if (absolutePath.endsWith(".ccz")) {
+                            engine.initFromArchive(absolutePath);
+                        } else {
+                            throw new RuntimeException("Can't install with reference: " + absolutePath);
+                        }
+                        engine.initEnvironment();
+                        return engine;
+                    } catch (Exception e) {
+                        e.printStackTrace();
```
benrudolph: doesn't hte stack trace get printed and halt execution anyways? seems like it'd have the same behavior if we removed this try block

src/main/java/beans/menus/Tile.java
wpride: Its legacy from mWellCare unfortunately, seemed like a good idea at the time :(

benrudolph:
👍 on ✔️

PR #54, opened 2016-07-11
Wsp/edit form fix

wpride:
Fix for bug mentioned here https://github.com/dimagi/commcare-hq/pull/12341. We were re-firing the initialization code even when we had an instance to load. This was causing us to load the form with a new instanceID which HQ accepted as a new form submission rather than an edit. 

@benrudolph

### DIFF
--- a/src/main/java/session/FormSession.java
+++ b/src/main/java/session/FormSession.java
@@ -98,6 +98,9 @@ public class FormSession {
 
         if(instanceContent != null){
             loadInstanceXml(formDef, instanceContent);
+            initialize(false, sessionData);
+        } else {
+            initialize(true, sessionData);
         }
 
         formEntryModel = new FormEntryModel(formDef, FormEntryModel.REPEAT_STRUCTURE_NON_LINEAR);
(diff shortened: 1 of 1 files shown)


src/main/java/session/FormSession.java
```diff
@@ -98,6 +98,9 @@ public FormSession(String formXml, String restoreXml, String locale, String user
 
         if(instanceContent != null){
             loadInstanceXml(formDef, instanceContent);
+            initialize(false, sessionData);
+        } else {
+            initialize(true, sessionData);
         }
```
benrudolph: nit: `initialize(instanceContent == null, sessionData)` so it can be put out of the if

benrudolph:
👍

src/main/java/session/FormSession.java
```diff
@@ -108,7 +111,6 @@ public FormSession(String formXml, String restoreXml, String locale, String user
         setLocale(locale, langs);
         uuid = UUID.randomUUID().toString();
```
benrudolph: i might be getting my PRs crossed, but don't we need to change this?

benrudolph: ok cool, makes sense

PR #60, opened 2016-07-20
sub update

wpride:
@benrudolph


wpride:
(self merging since its a subrepo update)

PR #61, opened 2016-07-20
Wsp/email exceptions

wpride:
@benrudolph this will allow Formplayer to send emails when it receives an exception. We'll need to configure the properties added to the example config file.

Also includes a refactor of some shared logic


src/main/java/application/UtilController.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Controller class (API endpoint) containing all all logic that isn't associated with
+ * a particular form session or menu navigation. Includes:
+ *      Get Cases
+ *      Filter Cases
+ *      Sync User DB
+ *      Get Sessions (Incomplete Forms)
+ */
+@Api(value = "Util Controller", description = "Operations that aren't associated with form or menu navigation")
```
wpride: everything in this class is just moved from FormController

config/application.properties.example
```diff
@@ -22,4 +22,9 @@ flyway.password=flyway
...
 flyway.location=classpath:db/migration
 
-spring.jpa.hibernate.ddl-auto
\ No newline at end of file
+spring.jpa.hibernate.ddl-auto
+
+smtp.host=127.0.0.1
+smtp.port=8888
+smtp.from.address=mail.formplayer.commcarehq.org
+smtp.to.address=exceptions@commcarehq.org
```
benrudolph: we're going to need to be able to authentication (provide a username and password) in order for us to actually use our mail service. when we add it, it should still work for users that don't need to email

benrudolph:
tests are failing

PR #64, opened 2016-07-21
add locking to session repo

wpride:
Add locking to Postgres session repository


benrudolph:
wonder if there's a way to test this at all. this is gonna be one of those things where the bug will be hard to repro and only appear in weird situations

benrudolph:
what level of locking does this happen at? just the row level i'm assuming? i.e. is this a global lock on the table?

wpride:
@benrudolph yes, at the row level

PR #69, opened 2016-07-27
Better defaults for example file

benrudolph:
@wpride i think HQ devs will find these defaults better


config/application.properties.example
```diff
@@ -2,23 +2,23 @@
 commcarehq.host=http://localhost:8000
 user.suffix=commcarehq.org
 logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=DEBUG
-server.port=8090
+server.port=8080
```
benrudolph: matches what's in HQ's settings

PR #71, opened 2016-07-28
Submission and Constraint Fixes

wpride:
Fix a few issues:
First, some of the labels were wrong ("constraint" vs "restraint"...) which was causing some issues in the JS
Second, FP wasn't respecting the "prevalidated" field so sometimes when the browser wanted to indicate there was an invalid answer (but not one visible to the server) we weren't noticing that
Third, the processXML method wasn't some form constraints
Added tests to cover these issues

cross https://github.com/dimagi/commcare-hq/pull/12648
cross https://github.com/dimagi/touchforms/pull/244


src/main/java/application/FormController.java
```diff
@@ -117,20 +125,60 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
...
+
+            ResponseEntity<String> submitResponse =
+                    submitService.submitForm(formEntrySession.getInstanceXml(),
+                            formEntrySession.getPostUrl(),
+                            new DjangoAuth(authToken));
+
+            if (submitResponse.getStatusCode().is2xxSuccessful()) {
+                sessionRepo.delete(submitRequestBean.getSessionId());
+            } else {
+                submitResponseBean.setStatus("error");
```
benrudolph: i think if you don't send back a 200 response, the frontend will handle it as an error

src/main/java/beans/menus/ErrorBean.java
```diff
@@ -0,0 +1,25 @@
+package beans.menus;
```
benrudolph: just curious why you put this in the menu directory, seems like it belons in a formentry package or something

benrudolph:
👍

src/main/java/beans/menus/ErrorBean.java
wpride: Looks like I accidentally but in the "menus" directory instead of its parent ("beans") where I'm putting all the response-formatting related stuff.

PR #74, opened 2016-08-04
Formplayer End of Form Navigation

wpride:
Implement end of form navigation for formplayer. Needed a few things:

We add the ability to store MenuSessions in much the same way we store FormSessions. Add a repository and mappings to do this.
When we start form entry, we save the MenuSession and pass its ID to the form session (new column in that table).
When form entry completes (in the submit controller) we re-instantiate the MenuSession from its ID. Then, we execute the form completion against that MenuSession and check if we have another command on the frame to perform.
If so, we get this command or entity list and return it along with the submit response.
We then save this menu session and add another route that allows the front end to perform new steps against this saved session.
Also add test coverage for this. 

cross https://github.com/dimagi/commcare-hq/pull/12718/files


src/main/java/application/AbstractBaseController.java
```diff
@@ -37,6 +56,70 @@
...
+            return generateFormEntryScreen(menuSession);
+        } else {
+            MenuBean menuResponseBean;
+
+            // We're looking at a module or form menu
+            if (nextScreen instanceof MenuScreen) {
+                menuResponseBean = generateMenuScreen((MenuScreen) nextScreen, menuSession.getSessionWrapper(),
+                        menuSession.getId());
+            }
+            // We're looking at a case list or detail screen (probably)
```
benrudolph: probably feels kind of scary in a comment. do you know the other cases?

src/main/java/application/FormController.java
```diff
@@ -104,13 +107,15 @@ public FormEntryResponseBean getCurrent(@RequestBody CurrentRequestBean currentR
     @ApiOperation(value = "Submit the current form")
     @RequestMapping(value = Constants.URL_SUBMIT_FORM, method = RequestMethod.POST)
     @ResponseBody
-    public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitRequestBean,
+    public Object submitForm(@RequestBody SubmitRequestBean submitRequestBean,
```
benrudolph: why `Object` now?

src/main/java/db/migration/V2__init_menu_session.java
```diff
@@ -0,0 +1,26 @@
...
+                "    id text NOT NULL,\n" +
+                "    username text,\n" +
+                "    domain text,\n" +
+                "    appid text,\n" +
+                "    installreference text,\n" +
+                "    locale text,\n" +
+                "    commcaresession bytea,\n" +
+                "    CONSTRAINT menu_sessions_pkey PRIMARY KEY (id)\n" +
+                ")");
+        jdbcTemplate.execute("ALTER TABLE formplayer_sessions ADD menu_session_id text");
```
benrudolph: we may want to add a constraint to this to ensure that there is a matching menu_session_id (i.e. it really is a foreign key)

src/main/java/db/migration/V2__init_menu_session.java
```diff
@@ -0,0 +1,26 @@
...
+ * @author wspride
+ */
+public class V2__init_menu_session implements SpringJdbcMigration {
+
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        jdbcTemplate.execute("CREATE TABLE menu_sessions (\n" +
+                "    id text NOT NULL,\n" +
+                "    username text,\n" +
+                "    domain text,\n" +
```
benrudolph: it's good practice to label these as NOT NULL if they are required. that way we never accidentally commit bad data

src/main/java/repo/impl/PostgresMenuSessionRepo.java
```diff
@@ -0,0 +1,145 @@
...
+
+        if(sessionCount > 0){
+            String query = replaceTableName("UPDATE %s SET commcaresession = ? WHERE id = ?");
+            this.jdbcTemplate.update(query,  new Object[] {session.getCommcareSession(), session.getId()},
+                    new int[] {Types.BINARY, Types.VARCHAR});
+            return session;
+
+        }
+
+        String query = replaceTableName("INSERT into %s " +
```
benrudolph: shouldn't the insert only happen if sessionCount == 0

src/main/java/requests/NewFormRequest.java
```diff
@@ -77,15 +77,17 @@ public NewFormRequest(NewSessionRequestBean bean, SessionRepo sessionRepo,
...
     }
 
     public NewFormSessionResponse getResponse() throws IOException {
         return new NewFormSessionResponse(formEntrySession);
     }
 
     private String getRestoreXml(){
-        return restoreService.getRestoreXml(domain, auth);
+        String xml =  restoreService.getRestoreXml(domain, auth);
+        System.out.println("Restoring XML: " + xml);
```
benrudolph: print

src/test/java/tests/CaseTests.java
```diff
@@ -45,6 +45,8 @@ public void testCases() throws Exception {
 
         caseFilterResponseBean = filterCases("requests/filter/filter_cases_5.json");
 
+        System.out.println("Got case filte response: " + caseFilterResponseBean);
```
benrudolph: print

src/test/java/tests/DoubleManagementTest.java
```diff
@@ -63,6 +64,21 @@ public void testDoubleForm() throws Exception {
...
+        CommandListResponseBean commandListResponseBean = mapper.readValue(mapper.writeValueAsString(submitResponseBean.getNextScreen()),
+                CommandListResponseBean.class);
+        assert commandListResponseBean.getCommands().length == 2;
+        assert commandListResponseBean.getCommands()[0].getDisplayText().equals("Update Parent");
+        JSONObject endOfFormNavResponse = sessionNavigateWithId(new String[] {"0"}, "derp");
+        NewFormSessionResponse followupFormResponse =
+                mapper.readValue(endOfFormNavResponse.toString(), NewFormSessionResponse.class);
+        assert followupFormResponse.getTree().length == 2;
+        assert followupFormResponse.getTree()[0].getAnswer().equals("David Ortiz");
+        assert followupFormResponse.getTree()[1].getAnswer().equals(40);
```
benrudolph: 🙌

src/test/java/tests/DoubleManagementTest.java
```diff
@@ -120,8 +136,9 @@ public void testAllPermutations() throws Exception {
         JSONObject parentResponseObject2 = sessionNavigate(new String[] {"2"}, "doublemgmt");
         EntityListResponse entityListResponse =
                 mapper.readValue(parentResponseObject2.toString(), EntityListResponse.class);
+        System.out.println("Entity List Response: " + entityListResponse);
```
benrudolph: print

benrudolph:
nice! 📝 🌏 

one thing i noticed is that menusession and form session seem very similar. is there any way we can formalize that with a interface or some shared utils? not quite sure, some of the logic goes over my head a bit so quite possibly not possible

src/main/java/application/AbstractBaseController.java
wpride: Haha it actually can't be anything else at the moment, removing ambiguous comment.

src/main/java/application/FormController.java
wpride: Good call, changed this during testing and needs to go back.

src/main/java/repo/impl/PostgresMenuSessionRepo.java
wpride: Yes, we return on line 51 otherwise.

wpride:
@benrudolph agreed, these share a lot of code now. I think I'd rather address that in a later PR - I could imagine going a few different ways with it now (interface, abstract class, or merging the two completely)

src/main/java/repo/impl/PostgresMenuSessionRepo.java
benrudolph: 👍 missed that

benrudolph:
ok sounds good

PR #75, opened 2016-08-05
fix email sending

benrudolph:
@wpride aight fixed this email thing. pandora's box has been opened 📖  best read commit by commit

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,6 +53,8 @@ dependencies {
     compile 'org.springframework.boot:spring-boot-starter-data-jpa'
     // https://mvnrepository.com/artifact/org.apache.commons/commons-lang3
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.4'
+    // https://mvnrepository.com/artifact/org.apache.commons/commons-email
+    compile group: 'org.apache.commons', name: 'commons-email', version: '1.2'
     compile group: 'org.springframework', name: 'spring-jdbc', version: '3.1.0.RELEASE'
     compile group: 'org.postgresql', name: 'postgresql', version: '9.3-1100-jdbc4'
     compile group: 'org.springframework.boot', name: 'spring-boot-starter-jdbc', version: '1.3.5.RELEASE'
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -2,11 +2,12 @@ package application;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.commons.mail.EmailException;
+import org.apache.commons.mail.HtmlEmail;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.json.JSONObject;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.mail.MailSendException;
-import org.springframework.mail.SimpleMailMessage;
-import org.springframework.mail.javamail.JavaMailSenderImpl;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import repo.SessionRepo;
 import services.RestoreService;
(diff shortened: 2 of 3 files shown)


src/main/java/application/WebAppContext.java
```diff
@@ -83,10 +83,10 @@
     @Value("${smtp.port}")
     private int smtpPort;
 
-    @Value("${smtp.username")
+    @Value("${smtp.username}")
```
benrudolph: ha ha ha

build.gradle
```diff
@@ -53,6 +53,8 @@ dependencies {
     compile 'org.springframework.boot:spring-boot-starter-data-jpa'
     // https://mvnrepository.com/artifact/org.apache.commons/commons-lang3
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.4'
+    // https://mvnrepository.com/artifact/org.apache.commons/commons-email
+    compile group: 'org.apache.commons', name: 'commons-email', version: '1.2'
```
benrudolph: i've been finding apache's libraries have a cleaner api and are easier to use than some of the other ones. for example i switched email libs because the other one didn't seem to have a way to send an html message (probably just couldn't find out how)

src/main/java/application/AbstractBaseController.java
```diff
@@ -64,8 +66,17 @@ private void sendExceptionEmail(Exception exception) {
...
+            StringUtils.join(stackTrace, "<br />"), "\t", "&nbsp;&nbsp;&nbsp;"
+        );
+        return "<html>" +
+                "<h3>Message</h3>" +
+                "<p>" + exception.getMessage() + "</p>" +
+                "<h3>Time</h3>" +
+                "<p>" + formattedTime + "<p>" +
+                "<h3>Trace</h3>" +
+                "<p>" + stackTraceHTML + "</p>" +
+                "</html>";
```
benrudolph: java not having multi line strings is a joke

src/main/java/application/AbstractBaseController.java
```diff
@@ -64,8 +66,17 @@ private void sendExceptionEmail(Exception exception) {
     private String getExceptionEmailBody(Exception exception){
         DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
         String formattedTime = dateFormat.format(new Date());
-        return "Message: " + exception.getMessage() + " \n " +
-                "Time : " + formattedTime + " \n " +
-                "Trace: " + Arrays.toString(exception.getStackTrace());
+        String[] stackTrace = ExceptionUtils.getStackTrace(exception).split("\n");
+        String stackTraceHTML = StringUtils.replace(
+            StringUtils.join(stackTrace, "<br />"), "\t", "&nbsp;&nbsp;&nbsp;"
```
benrudolph: also java not having string.join without a 3rd party lib is pretty crazy

benrudolph:
screen shot 2016-08-05 at 1 54 42 am

src/main/java/application/WebAppContext.java
wpride: ugh... sorry bruv 😢

src/main/java/application/AbstractBaseController.java
wpride: Yeah IOUtils and StringUtils both add some pretty basic stuff

src/main/java/application/AbstractBaseController.java
wpride: ha honestly I've never known another way of life.

wpride:
@benrudolph What email address will those go to? Can it notify the CloudCare virtual user somehow?

benrudolph:
that's a good question, i'm sure we could send it to some email address that would auto register it as a fogbugz ticket. we currently don't do that for our HQ 500 emails cause it's way too noisy. One 500 could happen hundreds of times which would mean lots of tickets. were you thinking of another form of notification besides creating a case?

wpride:
Yeah I guess we should probably wait and see how noisy this ends up being, but ideally we'd only be emailing for really bad code errors and then I could see opening a case being useful.

PR #76, opened 2016-08-05
build post URL manually for edit submissions

wpride:
~~If we are editing a form then we need to manually build to postUrl (since we don't install the application and thus can't query for the profile property)~~

Others options would be:
Performing full installation (correctest, but most overhead)
Sending the PostURL from the browser

EDIT: Rewrote to do option 2 (since this seems to strike the best balance of correctness and efficiency)


benrudolph:
👍 conflicts

src/main/java/application/MenuController.java
```diff
@@ -91,7 +91,7 @@ private MenuSession performInstall(InstallRequestBean bean, String authToken) th
             if (password == null || "".equals(password.trim())) {
                 throw new RuntimeException("Either authToken or password must be non-null");
             }
-            auth = new BasicAuth(bean.getUsername(), bean.getDomain(), "commcarehq.org", password);
+            auth = new BasicAuth(bean.getUsername(), bean.getDomain(), host, password);
```
benrudolph: 👍

PR #78, opened 2016-08-08
add git status route

wpride:
screen shot 2016-08-08 at 7 07 42 pm

@benrudolph this is pretty nifty but I don't think it'll work without the repository being on the machine running the jar 😢  will probably have to fold this into the gradle job


benrudolph:
that is a bummer! hmm, do you think it's still worth it then? i don't think i'd ever use it for my local env since i'd just use regular git. maybe i'm missing something?

wpride:
@benrudolph yeah agreed. the plan I was trying to outline above was actually to use this library when gradle is run (IE when the jar is built) to add a value to the jar that's the commit hash - then we can just access this static variable in the application (and it'll always be based on the repo of the builder IE jenkins)

benrudolph:
gotcha 👍

wpride:
(gonna just close for now and reopen when I have a chance to fix)

PR #79, opened 2016-08-10
Tweak emails to give more information

benrudolph:
@wpride tweaking them emails to give more info. can't seem to get post body which is a shame

screen shot 2016-08-10 at 2 00 26 am

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -16,6 +16,7 @@ import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.json.JSONObject;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.web.bind.annotation.ExceptionHandler;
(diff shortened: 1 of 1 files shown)


src/main/java/application/AbstractBaseController.java
```diff
@@ -157,6 +161,10 @@ private String getExceptionEmailBody(Exception exception){
         return "<html>" +
                 "<h3>Message</h3>" +
                 "<p>" + exception.getMessage() + "</p>" +
+                "<h3>Request URI</h3>" +
+                "<p>" + req.getRequestURI() + "<p>" +
```
snopoke: @benrudolph  Did you try `getQueryString()` or `getParameterMap()` to get more details?

benrudolph: yeah i thought i had. i'll retry. need to revisit this anyways since gmail isn't parsing the html correctly

benrudolph: ok so double checked and those two methods don't return anything. i think this is because we send the information in the post body. when i try and read that information using `getInputStream` it throws an error saying that that stream has already been closed. i believe this is due to the jackson api library reading the post data when the request is first made

PR #82, opened 2016-08-12
Set logged in user after restore

wpride:
So the NPE Priya was seeing was due to getLoggedInUser of the SQLSandbox returning null. When I examined the table there were no rows for the User which was really weird. Hopefully this will obviate the need to read that table and will also fail early if we do a restore and don't get a user entry.


benrudolph:
test failure

PR #83, opened 2016-08-16
Fix up html

benrudolph:
@wpride hoping this'll fix up some of the html email formatting. hard to tell since it already works locally for me

cc: @czue

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -158,17 +158,15 @@ public abstract class AbstractBaseController {
         String stackTraceHTML = StringUtils.replace(
             StringUtils.join(stackTrace, "<br />"), "\t", "&nbsp;&nbsp;&nbsp;"
         );
-        return "<html>" +
-                "<h3>Message</h3>" +
+        return "<h3>Message</h3>" +
                 "<p>" + exception.getMessage() + "</p>" +
                 "<h3>Request URI</h3>" +
-                "<p>" + req.getRequestURI() + "<p>" +
+                "<p>" + req.getRequestURI() + "</p>" +
                 "<h3>Host</h3>" +
-                "<p>" + hqHost + "<p>" +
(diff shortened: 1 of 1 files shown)


src/main/java/application/AbstractBaseController.java
```diff
@@ -158,17 +158,15 @@ private String getExceptionEmailBody(HttpServletRequest req, Exception exception
         String stackTraceHTML = StringUtils.replace(
             StringUtils.join(stackTrace, "<br />"), "\t", "&nbsp;&nbsp;&nbsp;"
         );
-        return "<html>" +
```
benrudolph: gmail doesn't allow `<html>` tags

src/main/java/application/AbstractBaseController.java
```diff
@@ -158,17 +158,15 @@ private String getExceptionEmailBody(HttpServletRequest req, Exception exception
...
                 "<h3>Host</h3>" +
-                "<p>" + hqHost + "<p>" +
+                "<p>" + hqHost + "</p>" +
                 "<h3>Time</h3>" +
-                "<p>" + formattedTime + "<p>" +
+                "<p>" + formattedTime + "</p>" +
                 "<h3>Trace</h3>" +
-                "<p>" + stackTraceHTML + "</p>" +
-                "</html>";
+                "<p>" + stackTraceHTML + "</p>";
```
czue: dear lord does java not have string templating?

wpride: This would probably be the closest out of the box you'd get:

https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html

benrudolph: i played with this for a second: http://www.stringtemplate.org/ but then couldn't get it working and decided it was too much work for just this

snopoke: You can do `String.format('<p>%s</p><p>%s</p>', var1, var2)`

PR #85, opened 2016-08-16
Set form title on answer and repeat responses

wpride:
Fix for http://manage.dimagi.com/default.asp?pre=preMultiSearch&pg=pgList&pgBack=pgSearch&search=2&searchFor=235105


benrudolph:
haha how did we never notice this?

benrudolph:
👍 on ✔️

benrudolph:
(it also does seem like a thing that's very easily tested)

PR #87, opened 2016-08-17
throw more explicit error when offset overflows caselist

wpride:


benrudolph:
hooray for error handling!

PR #88, opened 2016-08-17
Lock API routes by user

wpride:
This should clean up most of the issues that have been coming up recently - basically, only allow a single user to be using any route at a given time.

Definitely want to clean up the FormController routes in particular - lots of shared code now. 

Also removed redundant mocking in test context

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -32,6 +32,7 @@ import util.Constants;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.locks.Lock;
 
 /**
  * Controller class (API endpoint) containing all form entry logic. This includes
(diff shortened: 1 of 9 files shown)


wpride:
@benrudolph we should get this merged and deployed ASAP

src/main/java/application/FormController.java
```diff
@@ -226,31 +262,41 @@ public EvaluateXPathResponseBean evaluateXpath(@RequestBody EvaluateXPathRequest
...
+            JSONObject response = JsonActionUtils.descendRepeatToJson(formEntrySession.getFormEntryController(),
+                    formEntrySession.getFormEntryModel(),
+                    newRepeatRequestBean.getRepeatIndex());
+            updateSession(formEntrySession, serializableFormSession);
+            FormEntryResponseBean repeatResponseBean = mapper.readValue(response.toString(), FormEntryResponseBean.class);
+            repeatResponseBean.setTitle(formEntrySession.getTitle());
+            log.info("New response: " + repeatResponseBean);
+            return repeatResponseBean;
+        } finally {
+            lock.unlock();
```
benrudolph: so brutal we have to put this in every method 😞

PR #94, opened 2016-08-22
Give defaults to old incomplete form sessions, sort on server side

wpride:
Fixes https://github.com/dimagi/commcare-hq/pull/12913

Cross https://github.com/dimagi/commcare-hq/pull/12913

@benrudolph I'm not too worried about the default values since these should only ever show up for people on the test versions

### DIFF
--- a/src/main/java/db/migration/V4__incomplete_forms_default.java
+++ b/src/main/java/db/migration/V4__incomplete_forms_default.java
@@ -1,0 +1,19 @@
+package db.migration;
+
+        import org.flywaydb.core.api.migration.spring.SpringJdbcMigration;
+        import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Add default values to new incomplete forms columns
+ * @author wspride
+ */
+public class V4__incomplete_forms_default implements SpringJdbcMigration {
+
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        jdbcTemplate.execute("ALTER TABLE formplayer_sessions " +
+                "ALTER title SET DEFAULT 'CommCare Form'");
+        jdbcTemplate.execute("ALTER TABLE formplayer_sessions " +
+                "ALTER dateOpened SET DEFAULT 'Tue Jul 3 14:19:24 EDT 1990'");
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/db/migration/V4__incomplete_forms_default.java
```diff
@@ -0,0 +1,19 @@
...
+ * @author wspride
+ */
+public class V4__incomplete_forms_default implements SpringJdbcMigration {
+
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        jdbcTemplate.execute("ALTER TABLE formplayer_sessions " +
+                "ALTER title SET DEFAULT 'CommCare Form'");
+        jdbcTemplate.execute("ALTER TABLE formplayer_sessions " +
+                "ALTER dateOpened SET DEFAULT 'Tue Jul 3 14:19:24 EDT 1990'");
```
benrudolph: this is kind of a confusing default, can we set it to either `CURRENT_DATE` or Jan 1 1970 (beginning of epoch)?

src/main/java/db/migration/V5__incomplete_forms_set.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Give old incomplete forms a default value
+ */
+public class V5__incomplete_forms_set implements SpringJdbcMigration {
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        jdbcTemplate.execute("UPDATE formplayer_sessions " +
+                "SET title='CommCare Form' WHERE title IS NULL");
+        jdbcTemplate.execute("UPDATE formplayer_sessions " +
+                "SET dateOpened='Tue Jul 3 14:19:24 EDT 1990' WHERE dateOpened IS NULL");
```
benrudolph: same as above

src/main/java/db/migration/V4__incomplete_forms_default.java
wpride: Ben what is confusing about this its my birthday

benrudolph:
not too worried about the default values either, but still nice to have something logical so that it's fairly obvious that the default value was used when we inspect the data

PR #96, opened 2016-08-22
Don't re-save menuSession during stepping

wpride:
Also fixes http://manage.dimagi.com/default.asp?235449

Basically we only want to save menu state when we start or finish form entry - otherwise, remain RESTful


benrudolph:
👍

PR #99, opened 2016-08-23
Route to delete application databases

benrudolph:
@wpride this provides the ability to delete an application's databases. Needed for "refresh" in app preview.

cc: @dannyroberts


src/main/java/util/ApplicationUtils.java
```diff
@@ -0,0 +1,22 @@
+package util;
+
+import org.commcare.api.persistence.SqlSandboxUtils;
+
+/**
+ * Utility methods for dealing with Applications
+ */
+public class ApplicationUtils {
```
benrudolph: Wasn't sure if this was the best place for this, but couldn't find anything else

src/main/java/application/UtilController.java
```diff
@@ -57,6 +57,15 @@ public SyncDbResponseBean syncUserDb(@RequestBody SyncDbRequestBean syncRequest,
         return new SyncDbResponseBean();
     }
 
+    @ApiOperation(value = "Wipe the applications databases")
+    @RequestMapping(value = Constants.URL_DELETE_APPLICATION_DBS, method = RequestMethod.POST)
+    public StatusResponseBean deleteApplicationDbs(
+            @RequestBody DeleteApplicationDbsRequestBean deleteRequest) {
+
+        Boolean success = deleteRequest.clear();
+        return new StatusResponseBean(success);
```
wpride: our pal @phillipm would prefer 

`return new StatusResponseBean(deleteRequest.clear())`

src/main/java/beans/DeleteApplicationDbsRequestBean.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * A request that delete's an application's databases
+ */
+public class DeleteApplicationDbsRequestBean {
+    private String appId;
+
+    public DeleteApplicationDbsRequestBean(){}
+
+    public Boolean clear() {
```
wpride: any reason for not using the primitive type (`boolean`) here?

src/main/java/util/ApplicationUtils.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public static Boolean deleteApplicationDbs(String appId) {
+        String dbPath = "dbs/" + appId;
```
wpride: Hm, this exposes a potential problem. Until now all these dbs had been shared across users since the assumption was they'd basically remain unchanged, but if we start living in a world where they can be wiped out we'll need to introduce some locking around them or have namespaced per-user installs such that we can piggy back on the per-user locking (leaning towards the latter)

src/main/java/beans/StatusResponseBean.java
```diff
@@ -0,0 +1,41 @@
+package beans;
+
+/**
+ * A generic status response that returns either "ok" or "error" with a message
+ * if it has errored.
+ */
+public class StatusResponseBean {
```
wpride: This class seems to have a lot in common with NotificationMessageBean - they should probably be merged

src/main/java/beans/StatusResponseBean.java
```diff
@@ -0,0 +1,41 @@
...
+    public static final String OK = "ok";
+    public static final String ERROR = "error";
+    public static final String GENERIC_ERROR = "There was an error processing the request";
+
+    private String status;
+    private String message;
+
+    public StatusResponseBean() {}
+
+    public StatusResponseBean(Boolean success) {
```
wpride: using `Boolean` instead of `boolean` is usually unnecessary (only really need the wrapper class when you have operations that require an object)

src/main/java/beans/DeleteApplicationDbsRequestBean.java
benrudolph: 👍 only reason is ignorance

src/main/java/beans/StatusResponseBean.java
benrudolph: 👍

src/main/java/util/ApplicationUtils.java
benrudolph: yeah might be worth it to offline. my gut is to namespace the app installs as well too

src/main/java/util/ApplicationUtils.java
benrudolph: punted on moving this for now since FileUtils is only defined in the `test` directory

src/main/java/application/UtilController.java
benrudolph: https://github.com/dimagi/formplayer/pull/99/commits/e7a79162f07def2344bb864c7797e118c53de9ae

src/main/java/beans/StatusResponseBean.java
benrudolph: https://github.com/dimagi/formplayer/pull/99/commits/8915a5102754207fc7129489dcecbcf870bece08

benrudolph:
Ok updated pressing comments, I think I'll take the namespaced application dbs in a different PR. I don't think it's critical for this one since app preview is the only place where it's exposed and that's not live yet

benrudolph:
ended up not being that hard to namespace the app installs: https://github.com/dimagi/formplayer/pull/99/commits/41872b779c3ecf64c9c30e76cbe49794b6f39762

src/test/java/tests/DeleteApplicationDbsTests.java
```diff
@@ -22,16 +24,16 @@
     @Test
     public void testDeleteApplicationDbsView() throws Exception {
         // Create application db by making an install request
-        String appId = "casetestappid";
+        String dbPath = getApplicationDBPath("casetestdomain", "casetestuser", "casetestappid");
```
wpride: NIT but usually we would do `ApplicationUtils.getApplicationDBPath()` since this is more readable http://stackoverflow.com/a/421127

src/main/java/util/ApplicationUtils.java
```diff
@@ -7,16 +7,19 @@
...
+            SqlSandboxUtils.deleteDatabaseFolder(getApplicationDBPath(domain, username, appId));
         } catch (Exception e) {
             e.printStackTrace();
             success = false;
         }
         return success;
     }
+
+    public static String getApplicationDBPath(String domain, String username, String appId) {
+        return "dbs/" + domain + "/" + username + "/" + appId;
```
wpride: nit: `dbs` -> `UserSqlSandbox.DEFAULT_DATBASE_PATH`

wpride:
Looking good! Just a couple nits

src/test/java/tests/DeleteApplicationDbsTests.java
benrudolph: good to know

benrudolph: https://github.com/dimagi/formplayer/pull/99/commits/5fc10a193401fcc879a072a24cc57b7eb25ae991

src/main/java/util/ApplicationUtils.java
benrudolph: https://github.com/dimagi/formplayer/pull/99/commits/842b475373add9e76e3cc5a9f34eb70baa3c7a11

benrudolph:
✔️

PR #100, opened 2016-08-27
Fix form linking

benrudolph:
@wpride fixes form linking. previously it would try to get the title of the screen even though the next screen was null causing an error. Also wouldn't this function be more aptly named `getNextScreen` rather than `getNextMenu`? maybe we can pair on a way to test this. couldn't think of any way that wasn't super heavy like creating and installing an app that has form linking

cc: @dannyroberts


benrudolph:
cross: https://github.com/dimagi/commcare-hq/pull/13015

benrudolph:
woot PR 100!! 💯

PR #102, opened 2016-08-30
Handle app config errors

benrudolph:
@wpride @snopoke this'll handle handle app config errors and display the message. i don't think this'll yet handle xpath errors, but all we need to do is catch those and re wrap as an ApplicationConfigError and that'll stop the emails

screen shot 2016-08-29 at 10 00 53 pm

I know it's ugly at the moment, going to work on fixing later

cross: https://github.com/dimagi/commcare-hq/pull/13038

### DIFF
--- a/src/main/java/install/FormplayerConfigEngine.java
+++ b/src/main/java/install/FormplayerConfigEngine.java
@@ -159,7 +159,7 @@ public class FormplayerConfigEngine {
         try {
             IOUtils.copy(errorStream, writer, "utf-8");
         } catch (IOException e) {
-            e.printStackTrace();
+            log.error("Unable to read error stream", e);
         }
         String errorMessage = writer.toString();
         JSONObject errorJson = new JSONObject(errorMessage);
--- a/src/main/java/install/FormplayerConfigEngine.java
+++ b/src/main/java/install/FormplayerConfigEngine.java
@@ -3,8 +3,12 @@
  */
 package install;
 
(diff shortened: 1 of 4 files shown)


src/main/java/install/FormplayerConfigEngine.java
```diff
@@ -148,6 +154,18 @@ public void initFromLocalFileResource(String resource) throws InstallCancelledEx
         init(reference);
     }
 
+    private void handleInstallError(InputStream errorStream) {
+        StringWriter writer = new StringWriter();
+        try {
+            IOUtils.copy(errorStream, writer, "utf-8");
+        } catch (IOException e) {
+            e.printStackTrace();
```
snopoke: This should rather log the error.

snopoke:
:+1:

src/main/java/install/FormplayerConfigEngine.java
benrudolph: ah good to know, thanks

benrudolph: https://github.com/dimagi/formplayer/pull/102/commits/793a4107629b237db12fc09f05dacbbb8faab730

wpride:
@benrudolph probably XPathException (and its sub classes) belong in the new handler too

wpride:
but 👍 on ✅

benrudolph:
right on, updated to catch base class

PR #103, opened 2016-08-30
Remove sys.exit

benrudolph:
@wpride i see a bunch of these in the commcare lib, but a couple in formplayer. isn't it really dangerous to have sys.exit? doesn't that mean it just stops the web server for everyone. not sure how to handle the ones in the commcare lib yet


wpride:
@benrudolph good point - those are from the CLI which does need to exit. I'll need to do a refactor to separate them out.

PR #104, opened 2016-08-30
safely get the app title if no selections

wpride:
Don't rely on there being a nextScreen when determining app title (since sometime there isn't)


benrudolph:
nice 👍

PR #106, opened 2016-08-31
Better error message with invalid preview command

wpride:
This will give a better error message when we end up with an invalid preview command as in http://manage.dimagi.com/default.asp?236315#1211260

screen shot 2016-08-31 at 9 39 55 am


benrudolph:
the stack trace in the ticket is different than the one you're catching i think (can offline during sync)

wpride:
@benrudolph Yeah, this is a different issue than the original but still a "real" one

benrudolph:
ok do we want to have this send an email? i think right now it will (it's unclear to me if this is something actionable on our end)

wpride:
@benrudolph yeah still interested in having it send an error for now - I want to make sure we're not giving a bad preview link somewhere

PR #111, opened 2016-09-01
Error gracefully if trying to access a form that doesn't exist

wpride:
Should prevent those "Expected 1 but found 0" exceptions that are cropping up in the 500s

Also updates our compatibility with master


benrudolph:
offlined, but i think the plan is to handle the error with a different status code so that formplayer knows to restart itself

benrudolph:
(offlined going to addressing redirect in other PR)

PR #112, opened 2016-09-02
don't email on session exception

wpride:
@benrudolph yet another one we shouldn't email on (happens when something in menu nav is misconfigured)


benrudolph:
👍 conflicts

PR #113, opened 2016-09-02
configurable sqlite dir

benrudolph:
@wpride this allows for a configurable sqlite data dir and allow us to encrypt the data directory. test don't pass yet because i have to figureout how to inject the proper dataDir when running tests. also can't merge this until settings are updated. 

cc: @dannyroberts


benrudolph:
✔️

PR #114, opened 2016-09-02
Allow local install, Enable UATBC Calendar View

wpride:
PR was initially created for the Calendar Case view but ended up being a lot of random items

1. Enabling local install IE you can pass in a URL path /local/path/to/app.ccz and then this will start a session playing this local file
2. Return appId with all menu responses
3. Renaming install_reference to installReference
4. And finally, returning how many entities we want to display per row and if we want > 1, return all cases at once

cross-request https://github.com/dimagi/commcare-hq/pull/13096


build.gradle
```diff
@@ -94,4 +94,10 @@ test {
 
 task wrapper(type: Wrapper) {
     gradleVersion = '2.3'
-}
\ No newline at end of file
+}
+
+task copyTestResources(type: Copy) {
```
wpride: @benrudolph I needed this gradle task in order to make my tests work from IntelliJ http://stackoverflow.com/a/26033752

benrudolph: oh awesome 🙌

src/main/java/beans/menus/EntityListResponse.java
```diff
@@ -29,12 +29,13 @@
...
     private Tile[] tiles;
     private int[] widthHints;
+    private int numEntitiesPerRow;
 
     private int pageCount;
     private int currentPage;
     private final String type = "entities";
 
-    public static final int CASE_LENGTH_LIMIT = 10;
+    public static int CASE_LENGTH_LIMIT = 10;
```
benrudolph: any reason to remove `final`?

src/main/java/beans/menus/EntityListResponse.java
```diff
@@ -50,12 +51,12 @@ public EntityListResponse(EntityScreen nextScreen, int offset, String searchText
 
         Vector<TreeReference> references = ec.expandReference(((EntityDatum) session.getNeededDatum()).getNodeset());
         processTitle(session);
+        processCaseTiles(shortDetail);
```
benrudolph: order is dependent here?

benrudolph:
test failure, but looking good :tight:

benrudolph:
:tight: like a 🐯

PR #115, opened 2016-09-05
Use case_id instead of index when selecting case

wpride:
@benrudolph here's the backend for this change, hold off on merge until frontend is fixed. 

cross https://github.com/dimagi/commcare-hq/pull/13106


benrudolph:
legit

PR #118, opened 2016-09-07
subby plus escape commit

wpride:
@benrudolph this will fix the plus escaping issue clark reported


wpride:
adds this commit: https://github.com/dimagi/commcare-core/commit/ae6de0880978edb11d2df763ecfb656f5507e33c

PR #119, opened 2016-09-07
Full user check

benrudolph:
@wpride this implements auth with HQ to ensure that the domain and user match up with the domain and user passed in by the session cookie. quite few things going on here, may be best to review synchronously. tests are coming (which i'm sure will be bear 🐻 to get working ). did my best to break it down commit by commit, but there's still a decent amount of churn

cc: @gcapalbo


src/main/java/application/FormplayerAuthFilter.java
```diff
@@ -38,26 +55,78 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
         if (isAuthorizationRequired(request)) {
-            if (!authorizeRequest(request)) {
+            // These are order dependent
+            if (getSessionId(request) == null) {
                 setResponseUnauthorized((HttpServletResponse) res);
                 return;
             }
+            setToken(request);
+            setUser(request);
+            setDomain(request);
```
benrudolph: this should allow us to access the user and domain in error emails as well

src/main/java/util/RequestUtils.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.InputStreamReader;
+
+/**
+ * Utility function to deal with request objects.
+ */
+public class RequestUtils {
+
+    // Logic taken from here:
+    // http://stackoverflow.com/a/14885950/835696
+    public static String getBody(HttpServletRequest request) throws IOException {
```
benrudolph: unreal this is what it takes to read the body of a request... java

wpride:
This looks awesome @benrudolph ! You savage!

src/main/java/application/FormplayerAuthFilter.java
wpride: nice! seems like these three methods could maybe be combined somehow since they share a lot of variables - maybe in the constructor. Let's sync tomorrow

src/main/java/application/FormplayerAuthFilter.java
```diff
@@ -38,26 +55,78 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
 
-    private boolean authorizeRequest(HttpServletRequest request){
+    /**
+     * Takes a request object and queries postgres and couch to get the corresponding
+     * session user's.
+     *
+     * @param request
+     */
+    private void setUser(FormplayerHttpRequest request) {
+        JSONObject data = getPostData(request);
```
wpride: data is unused

src/main/java/application/AbstractBaseController.java
```diff
@@ -194,17 +205,33 @@ private void sendExceptionEmail(HttpServletRequest req, Exception exception) {
...
+        String params = "";
+        try {
+            params = RequestUtils.getBody(req);
+        } catch (IOException e) {
+            params = "";
+        }
         return "<h3>Message</h3>" +
                 "<p>" + exception.getMessage() + "</p>" +
+                "<h3>Domain</h3>" +
+                "<p>" + req.getDomain() + "</p>" +
```
wpride: 😂

src/main/java/application/AbstractBaseController.java
```diff
@@ -194,17 +205,33 @@ private void sendExceptionEmail(HttpServletRequest req, Exception exception) {
...
         );
+        String username = "Unknown";
+        if (req.getCouchUser() != null) {
+            username = req.getCouchUser().getUsername();
+        }
+        String params = "";
+        try {
+            params = RequestUtils.getBody(req);
+        } catch (IOException e) {
+            params = "";
```
wpride: NIT but this would still be "" from 219

src/main/java/application/FormplayerAuthFilter.java
```diff
@@ -38,26 +55,66 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
-        HttpServletRequest request = (HttpServletRequest) req;
+        FormplayerHttpRequest request = new FormplayerHttpRequest((HttpServletRequest) req);
         if (isAuthorizationRequired(request)) {
-            if (!authorizeRequest(request)) {
+            // These are order dependent
+            if (getSessionId(request) == null) {
+                setResponseUnauthorized((HttpServletResponse) res);
+                return;
+            }
+            setToken(request);
```
wpride: I think it might be cleaner to pass in the repositories to the FormplayerHttpRequest class and then do all this instantiation work there

src/main/java/beans/AuthenticatedRequestBean.java
```diff
@@ -0,0 +1,32 @@
+package beans;
+
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonSetter;
+
+/**
+ * Created by benrudolph on 9/8/16.
+ */
+public class AuthenticatedRequestBean {
```
wpride: Nice, really like this session vs. not breakdown. Might be good to add a comment clarifying when to use one vs the other

src/main/java/application/FormplayerAuthFilter.java
benrudolph: i tried doing that but i need a full instantiated request class in order to do these functions (i need to be able to get the POST). we'd have to make another method that would be initialize request or something which felt a little weird, but i'm not opposed to it

src/main/java/application/AbstractBaseController.java
benrudolph: https://github.com/dimagi/formplayer/pull/119/commits/de07543d8f65ffb8340b735229d0352aa3502215

src/main/java/beans/AuthenticatedRequestBean.java
benrudolph: https://github.com/dimagi/formplayer/pull/119/commits/7dc60baa60d441f149cff1ad236adcb4a7a9831f

PR #129, opened 2016-09-12
use database name

benrudolph:
@wpride need this for it to work on staging (should still work on prod)


benrudolph:
https://github.com/dimagi/commcarehq-ansible/pull/531

PR #132, opened 2016-09-13
Formplayer allow one question per screen

wpride:
@benrudolph

### DIFF
--- a/src/main/java/session/FormSession.java
+++ b/src/main/java/session/FormSession.java
@@ -1,8 +1,6 @@
 package session;
 
-import application.SQLiteProperties;
 import hq.CaseAPIs;
-import install.FormplayerConfigEngine;
 import objects.SerializableFormSession;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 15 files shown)


src/test/java/tests/FormEntryTest.java
```diff
@@ -133,4 +132,24 @@ public void testFormEntry2() throws Exception {
         assert green.getAnswer().equals(2);
 
     }
+
+    //Integration test of form entry functions
+    @Test
+    public void testOQPS() throws Exception {
```
wpride: Here is the actual new test

src/main/java/session/FormSession.java
```diff
@@ -194,7 +201,12 @@ public String getTitle(){
     }
 
     public JSONArray getFormTree() {
-        return JsonActionUtils.walkToJSON(getFormEntryModel(), getFormEntryController());
+        if (oneQuestionPerScreen) {
+            return JsonActionUtils.getOneQuestionPerScreenJSON(getFormEntryModel(),
+                    getFormEntryController(),
+                    JsonActionUtils.indexFromString("" + currentIndex, formDef));
```
benrudolph: nit: `(String) currentIndex` I think would be more idiomatic

benrudolph:
:tight:!!!

PR #133, opened 2016-09-14
add previous and next commands

wpride:
@benrudolph


src/main/java/application/FormController.java
```diff
@@ -321,6 +299,75 @@ public FormEntryResponseBean deleteRepeat(@RequestBody RepeatRequestBean deleteR
...
+    @ApiOperation(value = "Get the questios for the previous index in OQPS mode")
+    @RequestMapping(value = Constants.URL_PREV_INDEX, method = RequestMethod.POST)
+    @ResponseBody
+    public FormEntryResponseBean getPrevious(@RequestBody SessionRequestBean requestBean) throws Exception {
+        log.info("Get previous questions for: " + requestBean);
+        SerializableFormSession serializableFormSession = formSessionRepo.findOneWrapped(requestBean.getSessionId());
+        Lock lock = getLockAndBlock(serializableFormSession.getUsername());
+        try {
+            FormSession formSession = new FormSession(serializableFormSession);
+            formSession.setCurrentIndex(formSession.getCurrentIndex() - 1);
```
benrudolph: does this always work? don't we want to call `getAdjacentIndex`? if the index is of a question that shouldn't be displayed won't this evaluate incorrectly? highly plausible i don't understand the mobile code well enough

src/main/java/application/FormController.java
```diff
@@ -321,6 +299,75 @@ public FormEntryResponseBean deleteRepeat(@RequestBody RepeatRequestBean deleteR
...
+            updateSession(formSession, serializableFormSession);
+            FormEntryResponseBean responseBean = mapper.readValue(resp.toString(), FormEntryResponseBean.class);
+            log.info("Next index response: " + responseBean);
+            return responseBean;
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @ApiOperation(value = "Get the questios for the previous index in OQPS mode")
```
benrudolph: don't love OQPS acronym, reminiscent of fem, fec, ix, q

benrudolph:
nice nice

PR #135, opened 2016-09-15
Wsp/test refactor 2

wpride:
@benrudolph big test infrastructure refactor. Biggest changes

- Removing stupid function that mapped between an appid and a ccz, instead just pass in install reference directly
- Unify a bunch of repeated mock request code. To do so, have to pass in the return type as a class so that Jackson can map properly.


src/test/java/tests/BaseTestClass.java
```diff
@@ -164,57 +162,6 @@ public Condition newCondition() {
         }).when(userLockRegistry).obtain(any());
     }
 
-    private String resolveAppId(String ref){
```
benrudolph: 😂 😂 😂

src/test/java/tests/BaseTestClass.java
```diff
@@ -549,10 +476,11 @@ CommandListResponseBean doInstall(String requestPath) throws Exception {
         FORM, MENU, UTIL
     }
 
-    private String generateMockQuery(ControllerType controllerType,
+    private <T> T generateMockQuery(ControllerType controllerType,
```
benrudolph: nice this is great, noticed this too when working in the test code

benrudolph:
`+141 −319`

benrudolph:
conflicts

PR #136, opened 2016-09-15
Refactor

wpride:
Refactor the NewFormSessionRequest class to be more Spring-y, fix up some other random todos and ugliness


benrudolph:
nice looking good

PR #140, opened 2016-09-21
Restore refactor, restoreAs feature

wpride:
@benrudolph @czue 

This should be everything that we need on the backend for restoring as a different user than the logged in one. Started as a refactor, but ended up introducing some annoying complexity w/r/t keeping track of where the DBs actually are for a given session. I think we need another factory bean wrapping the entire SqlSandbox lifecycle. However, we can now set the username, domain, and auth immediately instead of passing them around until the restore actually occurs.

Hold off on merge until we can add a front end component, and also because I want to think about this with clear eyes


src/main/java/services/RestoreFactory.java
```diff
@@ -0,0 +1,88 @@
...
+        return response.getBody();
+    }
+
+    public static String getRestoreUrl(String host, String domain){
+        return host + "/a/" + domain + "/phone/restore/?version=2.0";
+    }
+
+    public String getRestoreUrl(String host, String domain, String username) {
+        return host + "/hq/admin/phone/restore/?as=" + username + "@" +
+                domain + ".commcarehq.org&version=2.0&raw=true";
```
benrudolph: minor thing that we probably don't have to worry about right now but this'll fail if you try to restore as another web user. this could happen if you're a web user and want to edit a form that was submitted by a web user. probably pretty rare

benrudolph: also nit, could use the URLBuilder

src/main/java/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -185,6 +186,7 @@ public SerializableFormSession mapRow(ResultSet rs, int rowNum) throws SQLExcept
             session.setDateOpened(rs.getString("dateOpened"));
             session.setOneQuestionPerScreen(rs.getBoolean("oneQuestionPerScreen"));
             session.setCurrentIndex(rs.getInt("currentIndex"));
+            session.setAsUser(rs.getString("asUser"));
```
benrudolph: so the frontend would have to set this parameter right

benrudolph:
nice cleanup

czue:
this is awesome. was not expecting you to just crank this out. :slot_machine:

src/main/java/services/RestoreFactory.java
wpride: @benrudolph looks like URIBuilder will escape the '@' character and break the restoreUrl, we'd probably need to implement our own subclass  of it

benrudolph: damn, i thought that was the whole point of the uri builder. seems fine to leave out

benrudolph:
@wpride is this close to being ready for merge?

wpride:
@benrudolph this should be gtg, removing don't merge label. Would be good to take a close look at the async restore changes - unfortunately one of my refactors conflicted closely with this (though I think the merge as easy enough).

src/main/java/services/RestoreFactory.java
```diff
@@ -110,8 +111,61 @@ private void handleAsyncRestoreResponse(String xml, HttpHeaders headers) {
...
+        }
+        return response.getBody();
+    }
+
+    public static String getRestoreUrl(String host, String domain){
         return host + "/a/" + domain + "/phone/restore/?version=2.0";
     }
+
+    public String getRestoreUrl(String host, String domain, String username) {
+        return host + "/hq/admin/phone/restore/?as=" + username + "@" +
```
benrudolph: we should use the `/a/<domain>/` route. i don't think this'll work for anyone who isn't a super usr

src/main/java/hq/CaseAPIs.java
```diff
@@ -30,7 +32,10 @@
  */
 public class CaseAPIs {
 
-    public static UserSqlSandbox forceRestore(String username, String domain, String xml) throws Exception {
+    public static UserSqlSandbox forceRestore(RestoreFactory restoreFactory) throws Exception {
+        String domain = restoreFactory.getDomain();
+        String username = TableBuilder.scrubName(restoreFactory.getUsername());
```
benrudolph: I think for sanity/debugging it'll be good to have this table has the asUsername as well so we can tell if the db is a restore as db

PR #141, opened 2016-09-22
subrepo update

wpride:


wpride:
@benrudolph tests pass locally, good to get into deploy today

PR #142, opened 2016-09-22
Respect OQPS from menu selection

wpride:
@benrudolph @biyeun 

This will allow us to set the oneQuestionPerScreen value here:

https://github.com/dimagi/commcare-hq/blob/245868750c34aa0fb301136e0f96e5dd05fa189f/corehq/apps/cloudcare/static/cloudcare/js/formplayer/entities/menu.js#L91-L91

And have that be respected. I didn't write the migration (see the TODO) because it'll clash with https://github.com/dimagi/formplayer/pull/140/files but will once that's merged


biyeun:
yes!!!!

PR #147, opened 2016-09-26
handle one question per screen on question answer

biyeun:
fyi for @wpride 

along with https://github.com/dimagi/commcare-core/pull/415

attempts to fix the issue where a question is answered in oneQuestionPerScreen mode and the whole question list is sent back rather than the current question

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -102,6 +102,7 @@ public class FormController extends AbstractBaseController{
             FormSession formEntrySession = new FormSession(session);
             JSONObject resp = formEntrySession.answerQuestionToJSON(answerQuestionBean.getAnswer(),
                     answerQuestionBean.getFormIndex());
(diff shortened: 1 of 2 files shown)


wpride:
@biyeun Good stuff! I moved the logic into FormSession and switched the commcare-core branch to our spring one but otherwise looks great. Longer term might want the one question per screen stuff in a different package or class, but that's more of a mobile architecture question.

cc @benrudolph

wpride:
(will let you decide on removing don't merge)

benrudolph:
👍 from me, not merging due to label

biyeun:
@wpride is this one no longer relevant then? https://github.com/dimagi/commcare-core/pull/415

wpride:
@biyeun yep

biyeun:
@benrudolph @wpride wanna merge :-D

PR #148, opened 2016-09-27
Error html

benrudolph:
@wpride this'll add the ability to handle html error returned from commcarehq


benrudolph:
cc: @proteusvacuum

PR #150, opened 2016-10-04
One Question Per Screen Updates

biyeun:
- handle submit in one question per screen mode
- know if index is end of form

related: https://github.com/dimagi/commcare-hq/pull/13441

@wpride @benrudolph 

cc @orangejenny

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -349,7 +349,9 @@ public class FormController extends AbstractBaseController{
                     formSession.getFormEntryModel(),
                     formSession.getCurrentIndex());
             updateSession(formSession, serializableFormSession);
-            FormEntryResponseBean responseBean = mapper.readValue(resp.toString(), FormEntryResponseBean.class);
+            FormEntryNavigationResponseBean responseBean = mapper.readValue(resp.toString(), FormEntryNavigationResponseBean.class);
+            responseBean.setIsAtLastIndex(formSession.getIsAtLastIndex());
+            responseBean.setCurrentIndex(formSession.getCurrentIndex());
             log.info("Next index response: " + responseBean);
             return responseBean;
         } finally {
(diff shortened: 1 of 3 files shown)


src/main/java/application/FormController.java
```diff
@@ -363,8 +373,9 @@ public FormEntryResponseBean getPrevious(@RequestBody SessionRequestBean request
                     formSession.getFormEntryModel(),
                     formSession.getCurrentIndex());
             updateSession(formSession, serializableFormSession);
-            FormEntryResponseBean responseBean = mapper.readValue(resp.toString(), FormEntryResponseBean.class);
+            FormEntryNavigationResponseBean responseBean = mapper.readValue(resp.toString(), FormEntryNavigationResponseBean.class);
```
wpride: Method signature for this and getNext can be changed to FormEntryNavigationResponseBean

src/main/java/application/FormController.java
```diff
@@ -341,7 +349,9 @@ public FormEntryResponseBean getNext(@RequestBody SessionRequestBean requestBean
...
+            responseBean.setIsAtLastIndex(formSession.getIsAtLastIndex());
```
wpride: I'd move this logic into the getCurrentJson method which should allow you to remove some of the logic out of FormSession as well

biyeun: Do you want me to change the `getCurrentJson` signature to accept additional params just for this use case? seems like more of a hassle than using it only where it's needed

wpride: Right but getCurrentJson does instantiate the object through the indexFromString method. Is also gettable through the getFormIndex() method of FormEntryModel. Writing that though I'm realizing we should definitely unify the way we're passing around the formIndex, though that's maybe out of scope for this PR.

wpride:
@biyeun maybe for now add a getCurrentJson method to FormSession that makes the call to the util method, then adds the currentIndex and isLast variables itself. This helps keep the logic out of the controller (which I try to minimize since those classes tend towards hugeness)

wpride:
(Accidental close)

biyeun:
@wpride made those changes to `FormSession` as you suggested

PR #152, opened 2016-10-12
Route to delete session

benrudolph:
@wpride this'll delete a session

cc: @orangejenny


src/main/java/application/FormController.java
```diff
@@ -92,6 +93,21 @@ public NewFormResponse openIncompleteForm(@RequestBody IncompleteSessionRequestB
         }
     }
 
+    @ApiOperation(value = "Open an incomplete form session")
```
orangejenny: "Open" => "Delete"

benrudolph: https://github.com/dimagi/formplayer/pull/152/commits/6465f10a5fe7dfe67b7f2b739cafc42c3a798d3c

PR #153, opened 2016-10-14
Async restores

benrudolph:
@wpride this implements the backend code needed to handle async restores. frontend handling is still a work in progress. We can go ahead and merge this is as though since async restores itself is behind a feature flag

cc: @orangejenny

### DIFF
--- a/src/main/java/services/impl/RestoreServiceImpl.java
+++ b/src/main/java/services/impl/RestoreServiceImpl.java
@@ -42,12 +42,14 @@ public class RestoreServiceImpl implements RestoreService {
         RestTemplate restTemplate = new RestTemplate();
         log.info("Restoring at domain: " + domain + " with auth: " + auth);
         HttpHeaders headers = auth.getAuthHeaders();
-        headers.add("X_OPENROSA_VERSION", "2.0");
+        headers.add("x-openrosa-version",  "2.0");
 
-        ResponseEntity<String> response =
-                restTemplate.exchange(getRestoreUrl(domain),
-                        HttpMethod.GET,
-                        new HttpEntity<String>(auth.getAuthHeaders()), String.class);
+        ResponseEntity<String> response = restTemplate.exchange(
+                getRestoreUrl(domain),
+                HttpMethod.GET,
+                new HttpEntity<String>(headers),
+                String.class
+        );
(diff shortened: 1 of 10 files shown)


benrudolph:
best read commit-by-commit

src/main/java/services/impl/RestoreServiceImpl.java
```diff
@@ -42,12 +42,14 @@ public String getRestoreXml(String domain, HqAuth auth) {
         RestTemplate restTemplate = new RestTemplate();
         log.info("Restoring at domain: " + domain + " with auth: " + auth);
         HttpHeaders headers = auth.getAuthHeaders();
-        headers.add("X_OPENROSA_VERSION", "2.0");
+        headers.add("x-openrosa-version",  "2.0");
```
benrudolph: thanks for your help @proteusvacuum 😖

src/main/java/services/impl/RestoreServiceImpl.java
```diff
@@ -27,13 +41,74 @@
...
+
+        String message = messageNodes.item(0).getTextContent();
+        Node progressNode = progressNodes.item(0);
+        NamedNodeMap attributes = progressNode.getAttributes();
+
+        throw new AsyncRetryException(
+                message,
+                Integer.parseInt(attributes.getNamedItem("done").getTextContent()),
+                Integer.parseInt(attributes.getNamedItem("total").getTextContent()),
+                Integer.parseInt(attributes.getNamedItem("retry-after").getTextContent())
```
proteusvacuum: I might have forgotten to mention this, can you use the Rety-After header to get this value instead of reading it from the XML? The mobile does it that way, so would be good keep consistent.

wpride:
Test fail looks real

src/main/java/services/impl/RestoreServiceImpl.java
benrudolph: https://github.com/dimagi/formplayer/pull/153/commits/aa332353b3bef331a577e168961cc3fef7f28cfe

proteusvacuum: 👍

benrudolph:
this is good to go. the frontend for this has already been merged, would be great to get this in too

PR #154, opened 2016-10-17
Case name in form session

benrudolph:
@wpride this'll include the case name in the session list. one of jwax's feedback items. i'm not sure what the performance implications are of taking all the serializable form sessions and converting them to regular form sessions are, but it didn't seem too bad

cc: @orangejenny

### DIFF
--- a/src/main/java/session/FormSession.java
+++ b/src/main/java/session/FormSession.java
@@ -1,5 +1,6 @@
 package session;
 
+import beans.CaseBean;
 import hq.CaseAPIs;
 import objects.SerializableFormSession;
 import org.apache.commons.codec.binary.Base64;
@@ -8,6 +9,8 @@ import org.apache.commons.io.output.ByteArrayOutputStream;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.commcare.api.json.JsonActionUtils;
+import org.commcare.api.persistence.SqliteIndexedStorageUtility;
+import org.commcare.cases.model.Case;
 import org.commcare.core.interfaces.UserSandbox;
 import org.commcare.modern.database.TableBuilder;
 import org.commcare.util.CommCarePlatform;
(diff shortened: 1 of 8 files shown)


src/main/java/session/FormSession.java
```diff
@@ -365,4 +370,21 @@ public JSONObject getNextJson() {
...
+    public void setDateOpened(String dateOpened) {
+        this.dateOpened = dateOpened;
+    }
+
+    public String getCaseName() {
+        String caseId = this.getSessionData().get("case_id");
+        if (caseId == null) {
+            return null;
+        }
+        CaseBean caseBean = CaseAPIs.getFullCase(caseId, (SqliteIndexedStorageUtility<Case>) this.getSandbox().getCaseStorage());
```
benrudolph: this seems not great. let me know if you have a better way of doing this

src/main/java/util/FormplayerDateUtils.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Date;
+import java.util.Locale;
+import java.util.TimeZone;
+
+/**
+ * Date utility functions for dealing with dates in formplayer
+ */
+public class FormplayerDateUtils {
+    public static String convertJavaDateStringToISO(String date) {
+        DateFormat dfFrom = new SimpleDateFormat("EEE MMM dd kk:mm:ss z yyyy", Locale.ENGLISH);
```
benrudolph: probably should have caught this earlier, but eventually we should be storing our dates in ISO format. it's going to leave us headaches down the line. currently we just `date.toString()` which java outputs as some non-iso standard date. that means when we send any dates anywhere else it's hard to parse. in this case i convert it to iso string so that the browser can more easily handle it

src/main/java/util/FormplayerDateUtils.java
```diff
@@ -0,0 +1,28 @@
...
+        Date result;
+        try {
+            result =  dfFrom.parse(date);
+        } catch (ParseException e) {
+            // Could not parse date
+            return null;
+        }
+        TimeZone tz = TimeZone.getTimeZone("UTC");
```
benrudolph: I believe that should be the timezone we're converting to, see test for confirmation: https://github.com/dimagi/formplayer/pull/154/files/039fc5ec595295f9e1f223f2144df1fc75f06bb1#diff-d98abeb04d3db7c31c3f68b399f19511R21

PR #155, opened 2016-10-19
AspectJ

benrudolph:
@wpride did some digging into how we can more cleaning define things that will execute before and after a function. This would be clutch for the locking code. anyways, i think i got it mostly setup, i think i'm just missing some sort of wiring. here's the documentation i'm working off of: http://docs.spring.io/spring/docs/current/spring-framework-reference/html/aop.html

curious if you have any thoughts/have done this before

### DIFF
--- a/src/main/java/aspects/LockOnUsername.java
+++ b/src/main/java/aspects/LockOnUsername.java
@@ -1,0 +1,27 @@
+package aspects;
+
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.annotation.Pointcut;
+import org.codehaus.jackson.map.RuntimeJsonMappingException;
+import org.springframework.stereotype.Component;
+
+/**
+ * Created by benrudolph on 10/19/16.
+ */
+@Component
+@Aspect
+public class LockOnUsername {
+
+    @Around("@annotation(annotations.LockAndBlock)")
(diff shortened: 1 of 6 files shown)


src/main/java/aspects/LockOnUsername.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Created by benrudolph on 10/19/16.
+ */
+@Component
+@Aspect
+public class LockOnUsername {
+
+    @Around("@annotation(annotations.LockAndBlock)")
```
benrudolph: this should run around any method that has the `@LockAndBlock` annotation, but it's not currently working

src/main/java/aspects/LockOnUsername.java
```diff
@@ -0,0 +1,27 @@
...
+    public Object aroundLockAndBlockOnUsername(ProceedingJoinPoint pjp) throws Throwable {
+        // start stopwatch
+        if (1== 1) {
+            throw new RuntimeException("BEN");
```
benrudolph: (just for testing purposes)

snopoke:
I've used AOP before and it's very useful for this kind of thing. Don't think I'll be much help with setting it up though.

benrudolph:
Ah that's good to hear. It does look incredibly useful once the setup is done. will offline with will about it

wpride:
Awesome find, wasn't aware of this feature of Spring at all. Would be interested on pairing on this if you're looking for an assist

snopoke:
This isn't specific to spring. You can use it in any Java runtime.

PR #160, opened 2016-10-27
Case List Sorting in Nimbus

wpride:
WIP, implements case list sorting

~~TODO~~

Could be merged without those if we want this though

screen shot 2016-10-27 at 3 22 20 pm

### DIFF
--- a/src/main/java/beans/menus/EntityListResponse.java
+++ b/src/main/java/beans/menus/EntityListResponse.java
@@ -1,6 +1,10 @@
 package beans.menus;
 
 import io.swagger.annotations.ApiModel;
+import org.commcare.cases.entity.Entity;
+import org.commcare.cases.entity.EntitySortNotificationInterface;
+import org.commcare.cases.entity.EntitySorter;
+import org.commcare.cases.entity.NodeEntityFactory;
 import org.commcare.modern.session.SessionWrapper;
 import org.commcare.modern.util.Pair;
 import org.commcare.suite.model.Action;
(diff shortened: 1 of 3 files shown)


wpride:
Wiped out TODO, think all of that should be resourced/handled separately and merge this core feature as is

src/main/java/beans/menus/EntityListResponse.java
```diff
@@ -156,27 +161,59 @@ private boolean matchEntity(Entity entity, String searchText) {
...
+    private List<Entity<TreeReference>> buildEntityList(Detail shortDetail, EvaluationContext context, Vector<TreeReference> references) {
+        NodeEntityFactory nodeEntityFactory = new NodeEntityFactory(shortDetail, context);
+        List<org.commcare.cases.entity.Entity<TreeReference>> full = new ArrayList<>();
+        for (TreeReference reference: references) {
+            full.add(nodeEntityFactory.getEntity(reference));
+        }
+        return full;
+    }
+
+    private void sort (List<Entity<TreeReference>> entityList, Detail shortDetail) {
```
benrudolph: i don't really get how this works. i would have expected some sort of sort that would take place in the object fetch, not after the objects have been retrieved from the database. possible i'm completely misinterpreting what this code is doing. if so might be good to add a comment!

benrudolph:
seems fine with me to break it up into smaller PRs

benrudolph:
test failure

src/main/java/beans/menus/EntityListResponse.java
wpride: The short answer is no, we can't retrieve in sorted order because this isn't part of the XPath spec. IE we can't give an XPath selector path with a sort included. IE we have a path like:

instance('casedb')/cases/case/[case_type = "our_type]

And we expand this to our set of references to our cases like

instance('casedb')/cases/case[401]
instance('casedb')/cases/case[404]
...

At this point all we know is we have these references into our context

Then we wrap these in the Entity() class so that we can treat them like cases with certain properties. Only then can we start indexing their properties and start sorting them.

benrudolph: oh gotcha, so there's an order property attached

PR #161, opened 2016-10-31
hack fix for model iteration expansion

wpride:
This fixes http://manage.dimagi.com/default.asp?239335

There's something we're not expanding during the usual getInstanceXml() call that is expanded when generating the form tree (only noticed this because the debugger had the correct XML while the submission didn't)

The right way to fix this is to to unify the android and nimbus submission code, which I'll do next

Can merge this for now as a hack


benrudolph:
this doesn't actually seem that time sensitive, worth trying to get to the bottom of it? does this process the xml twice?

PR #162, opened 2016-11-01
Add hq environment to emails

benrudolph:
@wpride this'll add the environment to emails so staging and production emails will thread separately. hold off on merge, needs ansible component


benrudolph:
good to merge

PR #163, opened 2016-11-01
Incomplete Session migration Old CloudCare -> Nimbus

wpride:
This will allow Formplayer to display incomplete sessions from CloudCare as if they were her own children. "Migration" is done piecemeal. When listing all of the sessions, we just load the migrated and regular forms up next to each other. Once a migrated form is opened, then Formplayer will save its own version of this session in its own db that will take precedence in all future interactions.


wpride:
@benrudolph realize we're probably going to need to give Formplayer write access to the HQ DB in order for this to work.

wpride:
@benrudolph this is ready to be looked at. I held off on adding the appId stuff since that'll require a migration and didn't end up being needed for this PR. Soon to come, however.

wpride:
@benrudolph Re-adding tag - occurred to me that when we submit the form and delete the Formplayer session, we'll be back in the same state as at the outst - IE the old session will start re-appearing in the session list.

src/main/java/application/IncompleteSessionController.java
```diff
@@ -0,0 +1,115 @@
...
+import repo.FormSessionRepo;
+import session.FormSession;
+import util.Constants;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.locks.Lock;
+
+/**
+ * Controller class (API endpoint) containing all incomplete session management commands
```
benrudolph: 👍 for moving this to its own controller

src/main/java/application/IncompleteSessionController.java
```diff
@@ -0,0 +1,115 @@
...
+        SerializableFormSession session;
+        restoreFactory.configure(incompleteSessionRequestBean, new DjangoAuth(authToken));
+        try {
+            session = formSessionRepo.findOneWrapped(incompleteSessionRequestBean.getSessionId());
+        } catch(FormNotFoundException e) {
+            session = migratedFormSessionRepo.findOneWrapped(incompleteSessionRequestBean.getSessionId());
+            // Move over to formplayer db
+            formSessionRepo.save(session);
+        }
+        Lock lock = getLockAndBlock(session.getUsername());
```
benrudolph: this'll null pointer if it can't find it in either dbs

src/main/java/application/IncompleteSessionController.java
```diff
@@ -0,0 +1,115 @@
...
+
+        restoreFactory.configure(getSessionRequest, new DjangoAuth(authToken));
+
+        List<SerializableFormSession> migratedSessions = migratedFormSessionRepo.findUserSessions(
+                getSessionRequest.getUsername());
+
+        List<SerializableFormSession> formplayerSessions = formSessionRepo.findUserSessions(username);
+
+        ArrayList<FormSession> formSessions = new ArrayList<>();
+        restoreFactory.configure(getSessionRequest, new DjangoAuth(authToken));
```
benrudolph: it may be nice to abstract this out so we can 1. remove the migration code easily and 2. run it in a separate command. i imagine that once we switch the toggle on for everyone and are happy with it, we'll want to just migrate all form sessions and then delete any migration code

src/main/java/beans/GetSessionsBean.java
```diff
@@ -15,6 +15,11 @@ public void setDomain(String domain) {
         this.domain = domain;
     }
 
+    @Override
+    public String getAsUser() {
+        return null;
```
benrudolph: why null?

src/main/java/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -73,10 +77,12 @@
 
         if(sessionCount > 0){
             String query = replaceTableName("UPDATE %s SET instanceXml = ?, sessionData = ?, " +
-                    "sequenceId = ?, currentIndex = ? WHERE id = ?");
+                    "sequenceId = ?, currentIndex = ?, postUrl = ?, restoreXml = ? WHERE id = ?");
```
benrudolph: sweet. just flagging now we could potentially have 3 different types of restores in a db. one from a restoreAs, a regular restore, and a old session one. the old session one is probably the one that would need to be namespaced the most if possible. i can imagine someone opening an old session in one screen and then in another filling out a form

src/test/java/tests/LoadOldIncompleteTest.java
```diff
@@ -0,0 +1,28 @@
...
+import util.SessionUtils;
+import utils.FileUtils;
+import utils.TestContext;
+
+/**
+ * Test loading the JSON from an old cloudcare sessions
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = TestContext.class)
+public class LoadOldIncompleteTest {
```
benrudolph: 💯 for testing this. might want to add a few more scenarios

src/main/java/repo/impl/PostgresMigratedFormSessionRepo.java
```diff
@@ -0,0 +1,416 @@
...
+        try {
+            XFormParser mParser = new XFormParser(new StringReader(formXml));
+            return mParser.parse();
+        } catch(Exception e) {
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+    private static String convertIsoToJavaDate(String date) {
```
benrudolph: it's counter part lives here: https://github.com/dimagi/formplayer/blob/master/src/main/java/util/FormplayerDateUtils.java#L14. may make sense to put them in the same spot

src/main/java/application/IncompleteSessionController.java
wpride: if the second one can't find it the FormNotFoundException will be bubbled to the user (as before)

src/main/java/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -73,10 +77,12 @@
```
snopoke: Not related to this PR but rather than doing a count you should do this:
```
select exists(select 1 from %s where id = ?)
```

src/main/java/repo/impl/PostgresMigratedFormSessionRepo.java
```diff
@@ -0,0 +1,418 @@
...
+            sessions.add(buildSerializedSession(entrySession, instanceSessions.get(0)));
+        }
+        return sessions;
+    }
+
+    @Override
+    @Lock(LockModeType.OPTIMISTIC)
+    public <S extends SerializableFormSession> Iterable<S> save(Iterable<S> entities) {
+        for(SerializableFormSession session: entities){
+            save(session);
```
snopoke: if `save` is doing nothing then why bother iterating? Also would it not be better to raise an exception here since presumably this should never get called.

src/main/java/repo/impl/PostgresMigratedFormSessionRepo.java
```diff
@@ -0,0 +1,418 @@
...
+            session.setRestoreXml(restoreFactory.getRestoreXml());
+        }
+
+        if(session.getPostUrl() == null) {
+            session.setPostUrl(
+                    StringUtils.getPostUrl(host,
+                            session.getDomain(),
+                            entrySession.getAppId()));
+        }
+        PrototypeUtils.setupPrototypes();
```
snopoke: What does this do? It's a bit weird having a static call that doesn't take any args and doesn't return anything.

Should this not be called as part of the server setup?

src/main/java/repo/impl/PostgresMigratedFormSessionRepo.java
```diff
@@ -0,0 +1,418 @@
...
+
+    @Override
+    public void deleteAll() {
+        Iterable<SerializableFormSession> allSessions = findAll();
+        for(SerializableFormSession session: allSessions){
+            delete(session.getId());
+        }
+    }
+
+    public String replaceTableName(String query, String tableName){
```
snopoke: This function doesn't seem to add any value.

src/main/java/repo/impl/PostgresMigratedFormSessionRepo.java
```diff
@@ -0,0 +1,418 @@
...
+
+    // Copied from FormSession
+    private static String serializeFormDef(FormDef formDef) {
+        try {
+            org.apache.commons.io.output.ByteArrayOutputStream baos = new org.apache.commons.io.output.ByteArrayOutputStream();
+            DataOutputStream serializedStream = new DataOutputStream(baos);
+            formDef.writeExternal(serializedStream);
+            return Base64.encodeBase64String(baos.toByteArray());
+        } catch(Exception e) {
+            e.printStackTrace();
```
snopoke: should log these exceptions

src/main/java/repo/impl/PostgresMigratedFormSessionRepo.java
snopoke: I think it still makes sense to have the two functions live together

src/main/java/util/FormplayerDateUtils.java
```diff
@@ -12,6 +12,7 @@
  */
 public class FormplayerDateUtils {
     public static String convertJavaDateStringToISO(String date) {
+        System.out.println("Parsing date " + date);
```
snopoke: sysout

src/main/java/repo/impl/PostgresMigratedFormSessionRepo.java
wpride: 👍

src/main/java/application/IncompleteSessionController.java
```diff
@@ -0,0 +1,132 @@
...
+                                              @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
+        log.info("Incomplete session request with bean: " + incompleteSessionRequestBean + " sessionId :" + authToken);
+        SerializableFormSession session;
+        restoreFactory.configure(incompleteSessionRequestBean, new DjangoAuth(authToken));
+        try {
+            session = formSessionRepo.findOneWrapped(incompleteSessionRequestBean.getSessionId());
+        } catch(FormNotFoundException e) {
+            session = migratedFormSessionRepo.findOneWrapped(incompleteSessionRequestBean.getSessionId());
+            // Move over to formplayer db
+            formSessionRepo.save(session);
```
wpride: 👍  good call. Would've worked this way too since we'd always find the formplayer version first, but agreed this is much cleaner.

src/main/java/application/IncompleteSessionController.java
```diff
@@ -0,0 +1,132 @@
...
+    public GetSessionsResponse getSessions(@RequestBody GetSessionsBean getSessionRequest,
+                                           @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
+        log.info("Get Session Request: " + getSessionRequest);
+
+        String username = TableBuilder.scrubName(getSessionRequest.getUsername());
+
+        restoreFactory.configure(getSessionRequest, new DjangoAuth(authToken));
+
+        List<SerializableFormSession> migratedSessions = migratedFormSessionRepo.findUserSessions(
+                getSessionRequest.getUsername());
```
wpride: Adding comment and renaming variable, but Old CloudCare doesn't use the escaped username

src/main/java/application/IncompleteSessionController.java
```diff
@@ -0,0 +1,132 @@
...
+        ArrayList<FormSession> formSessions = new ArrayList<>();
+
+        for (int i = 0; i < formplayerSessions.size(); i++) {
+            formSessions.add(new FormSession(formplayerSessions.get(i)));
+        }
+
+        if (migratedSessions.size() > 0) {
+
+            // First, get the id of every session we  got from the Formplayer repo so we don't process duplicates
+            ArrayList<String> formplayerSessionIds = new ArrayList<>();
```
wpride: Interesting, I used an ArrayList because I wanted the contains() function. Apparently using an array and Arrays.asList() would be more performant though: http://stackoverflow.com/questions/1128723/how-can-i-test-if-an-array-contains-a-certain-value

snopoke: Sets also have [contains](https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#contains(java.lang.Object)) and are much more performant lookups (O(1) for the set vs O(n) for the list)

snopoke:
merge conflict

PR #165, opened 2016-11-03
Handle 500 error for app builds

benrudolph:
@wpride couldn't get case tiles to work, but did realize we don't properly handle 500 errors from app manager
screen shot 2016-11-03 at 10 47 23 am
I think it's fine not to email, since that 500 email will be sent from HQ's side


benrudolph:
✅

PR #166, opened 2016-11-03
get the username from old cloudcare requests

wpride:
Old CloudCare puts the username in a different place than Nimbus


benrudolph:
debating if we should just not support this and pass the username in properly. at the very least we should probably comment this to say that it's ok to remove when touchforms is dead

PR #167, opened 2016-11-03
Debugger Route

benrudolph:
@czue @wpride

### DIFF
--- a/src/main/java/application/DebuggerController.java
+++ b/src/main/java/application/DebuggerController.java
@@ -1,0 +1,52 @@
+package application;
+
+import auth.DjangoAuth;
+import beans.SessionRequestBean;
+import beans.debugger.DebuggerFormattedQuestionsResponseBean;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import objects.SerializableFormSession;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.web.bind.annotation.*;
+import repo.SerializableMenuSession;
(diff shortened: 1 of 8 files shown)


benrudolph:
gonna make a documentation pass

src/main/java/application/DebuggerController.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @ApiOperation(value = "Get formatted questions and instance xml")
+    @RequestMapping(value = Constants.URL_DEBUGGER_FORMATTED_QUESTIONS, method = RequestMethod.POST)
+    public DebuggerFormattedQuestionsResponseBean getFormattedQuesitons(
+            @RequestBody SessionRequestBean debuggerRequest,
+            @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
+
+        SerializableFormSession serializableFormSession = formSessionRepo.findOneWrapped(debuggerRequest.getSessionId());
+        FormSession formSession = new FormSession(serializableFormSession);
+        SerializableMenuSession serializableMenuSession = menuSessionRepo.findOne(serializableFormSession.getMenuSessionId());
```
wpride: @benrudolph Want to flag that this isn't going to always exist, particularly for incomplete forms and edit forms. (Its also, in theory, possible to start a form without doing menu navigation though we don't do this at any point currently)

wpride: I think we should address this in a separate PR that also adds the app_id to the form session so that we don't need to do this silly chaining, anyhow

benrudolph: gotcha that makes sense. ok agree, the app id in the form session would be hugely helpful

PR #171, opened 2016-11-08
UserLock Aspect

wpride:
Remake of https://github.com/dimagi/formplayer/pull/155 but using Java configuration and some other changes

Weaves a function to handle locking and unlocking around method exception, allowing us to remove much cruft code

### DIFF
--- a/src/main/java/application/MenuController.java
+++ b/src/main/java/application/MenuController.java
@@ -1,5 +1,6 @@
 package application;
 
+import annotations.UserLock;
 import auth.BasicAuth;
 import auth.DjangoAuth;
 import auth.HqAuth;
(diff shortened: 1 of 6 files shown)


src/main/java/application/MenuController.java
```diff
@@ -67,90 +63,84 @@ public BaseResponseBean installRequest(@RequestBody InstallRequestBean installRe
...
-            } else {
-                return new BaseResponseBean(null, "Got null menu, redirecting to home screen.", false, true);
+            BaseResponseBean syncResponse = checkDoSync(nextScreen,
+                    menuSession,
+                    notificationMessageBean,
+                    auth);
+            if (syncResponse != null) {
+                return syncResponse;
             }
-        } catch (MenuNotFoundException e) {
```
benrudolph: this looks to actually be removed, how come we don't need it anymore?

benrudolph:
one question, but otherwise looks great!! so excited to see this. 😂 😂 😂

PR #173, opened 2016-11-09
Update spring gradle plugin to work with gradle 3

phillipm:


wpride:
@phillipm gonna hold off on this one until Ben is online - don't want to deploy and find that this requires updating some vars on ansible

benrudolph:
we install gradle here: https://github.com/dimagi/commcarehq-ansible/blob/7b68bc44bb7953c4ed98f2c58027581cd93e1e09/ansible/roles/common_installs/tasks/main.yml#L15. we haven't pinned it to a version, does the plugin require a certain version of gradle?

wpride:
@phillipm @benrudolph FYI this caused formplayer to fail on startup when deployed. Looks like the issue described here:

https://github.com/spring-projects/spring-boot/issues/6314

Was failing silently on 1.3 but started failing hard on 1.4. I've reverted this for the time being

wpride:
@benrudolph we should see if we can incorporate actually starting the server into the testing process

benrudolph:
yeah strange i would've thought travis would catch. agree we should try to boot the server up. there are quite a few advantages to knowing that it starts

PR #175, opened 2016-11-09
Return proceed result after logging

wpride:


wpride:
(self merging since formplayer is down)

PR #177, opened 2016-11-09
update commcare subrepo

wpride:


benrudolph:
diff is too big for me to grok. seems fine to me as long as tests pass

PR #178, opened 2016-11-15
Fix restoreAs

wpride:
Started off addressing the remaining restoreAs issues @czue pointed out, ended up refactoring a lot more

- Changed AsUserBean from an interface to a class (actually just part of AuthenticatedRequestBean) since the logic was repeated
- Change asUser to restoreAs (since this is what we call it on the front end and this mismatch was causing us to not actually use the restoreAs param in half the requests)
- Move the logic for user DB name spacing into the RestoreFactory
- fix all of the tests (many of the JSON files used an old format but weren't failing because of how we were mocking)

This enables the "submitted on behalf of" readout to work correctly

screen shot 2016-11-14 at 8 07 05 pm

Realized partway through that this is a half measure, though, as evidenced by the hacked in RestoreFactory for old sessions. I'm constantly running into the problem where the clean solution for session requests requires hackiness for other requests and vice versa. I think we need some FormSessionFactory to deal with this but I haven't thought out the design clearly yet. 

cc @benrudolph


czue:
hotness

src/main/java/beans/AuthenticatedRequestBean.java
```diff
@@ -11,24 +8,29 @@
...
     public String getDomain() {
         return domain;
     }
 
-    @JsonSetter(value = "domain")
     public void setDomain(String domain) {
         this.domain = domain;
     }
+
+    public String getRestoreAs() {
```
benrudolph: dig moving this here

src/main/java/session/FormSession.java
```diff
@@ -94,9 +94,7 @@ public FormSession(SerializableFormSession session) throws Exception{
         this.asUser = session.getAsUser();
         this.restoreXml = session.getRestoreXml();
         this.domain = session.getDomain();
-        this.sandbox = CaseAPIs.restoreIfNotExists(asUser != null ? asUser : username,
-                this.domain,
-                restoreXml);
+        this.sandbox = CaseAPIs.restoreIfNotExists(username, asUser, this.domain, restoreXml);
```
benrudolph: 👏

PR #184, opened 2016-11-17
WIP Add CommCare functions to autocomplete

wpride:
@benrudolph @phillipm 

screen shot 2016-11-16 at 7 45 59 pm

@benrudolph We'll want to figure out how to get the cursor to move between the parens after making a selection. Also we should think about incorporating argument count and insert the commas too. 

cross https://github.com/dimagi/commcare-hq/pull/13955


src/main/java/util/PrototypeUtils.java
```diff
@@ -37,38 +46,43 @@ public static void setupPrototypes(){
...
-                "org.javarosa.xpath.expr.XPathEqExpr",
-                "org.javarosa.xpath.expr.XPathFilterExpr",
-                "org.javarosa.xpath.expr.XPathFuncExpr",
-                "org.javarosa.xpath.expr.XPathNumericLiteral",
-                "org.javarosa.xpath.expr.XPathNumNegExpr",
-                "org.javarosa.xpath.expr.XPathPathExpr",
-                "org.javarosa.xpath.expr.XPathStringLiteral",
-                "org.javarosa.xpath.expr.XPathUnionExpr",
-                "org.javarosa.xpath.expr.XPathVariableReference"};
+                XPathPathExpr.class.getName(),
```
benrudolph: 👍

src/main/java/beans/debugger/QuestionResponseItem.java
```diff
@@ -11,6 +11,12 @@
     private String label;
     private String type;
 
+    public QuestionResponseItem(String value, String label, String type) {
```
benrudolph: this might make sense to refactor to implement XPathAutoCompleteItem or something so that it makes more semantic sense

emord:
> i think we'll just have to have an event listener for when it gets inserted and then move cursor in the browser. atjs might even have a callback to do stuff after insert

It does have that callback, but I think the harder thing will be that you need to do something funky with the matcher to make sure that the dropdown doesn't pop up when inside the parens. After that is solved you guys should port this back to vellum :)

benrudolph:
ah true i forgot about that that would happen. hmmm... we may be shoehorning atwho.js

src/main/java/beans/debugger/AutoCompletableItem.java
```diff
@@ -6,18 +6,18 @@
...
 
-    public QuestionResponseItem(String value, String label, String type) {
+    public AutoCompletableItem(String value, String label, String type) {
         this.value = value;
         this.label = label;
         this.type = type;
     }
 
-    public QuestionResponseItem(JSONObject questionJSON) {
+    public AutoCompletableItem(JSONObject questionJSON) {
```
benrudolph: nit: `questionJSON` should probably be renamed

benrudolph:
eh? why close?

wpride:
was contained in https://github.com/dimagi/formplayer/pull/185

wpride:
The commcare-core situation got really messy and I ended up just merging the two

benrudolph:
ah i see missed that

PR #185, opened 2016-11-17
App updates for Formplayer

wpride:
This also folds in fairly big refactor of FormplayerConfigEngine (to stop existing) by using the unfortunate factory pattern already in CommCareConfigEngine
Also enables WAL for SQLite so that we can perform updates
We should probably fold some "shouldUpdate" parameter into our other requests but that can wait.
This also includes the function autocomplete code
Onwards and upwards!


benrudolph:
❌ and conflicts

src/main/java/install/FormplayerConfigEngine.java
```diff
@@ -1,239 +0,0 @@
...
-    public CommCarePlatform getPlatform() {
-        return platform;
-    }
-
-    public FormDef loadFormByXmlns(String xmlns) {
-        IStorageUtilityIndexed<FormDef> formStorage =
-                (IStorageUtilityIndexed) StorageManager.getStorage(FormDef.STORAGE_KEY);
-        return formStorage.getRecordForValue("XMLNS", xmlns);
-    }
-}
```
benrudolph: :clap:

src/main/java/application/MenuController.java
```diff
@@ -249,4 +260,10 @@ private MenuSession performInstall(InstallRequestBean bean, String authToken) th
                 bean.getInstallReference(), bean.getLocale(), installService, restoreFactory, auth, host,
                         bean.getOneQuestionPerScreen(), bean.getRestoreAs());
     }
+
+    private MenuSession performUpdate(UpdateRequestBean updateRequestBean, String authToken) throws Exception {
+        MenuSession currentSession = performInstall(updateRequestBean, authToken);
+        currentSession.updateApp(updateRequestBean.getUpdateMode());
```
benrudolph: how does this exactly work? given an update mode it will query HQ at a certain URL to ensure that it gets the latest release/build/save for that app id?

src/main/java/beans/debugger/AutoCompletableItem.java
```diff
@@ -17,10 +17,10 @@ public AutoCompletableItem(String value, String label, String type) {
         this.type = type;
     }
 
-    public AutoCompletableItem(JSONObject questionJSON) {
-        this.value = questionJSON.getString("value");
-        this.label = questionJSON.getString("label");
-        this.type = questionJSON.getString("type");
+    public AutoCompletableItem(JSONObject itemnJSON) {
+        this.value = itemnJSON.getString("value");
```
benrudolph: haha this says `itemn`

src/main/java/application/MenuController.java
wpride: Yep - basically just gets passed directly here https://github.com/dimagi/commcare-core/blob/a24d48d1f57dd7bcbde34354a827fa92afc1d49f/util/src/org/commcare/util/engine/CommCareConfigEngine.java#L419-L419 where its used as a URL param

PR #186, opened 2016-11-17
Move oqps and currentIndex vars into formSession

wpride:
cross-request: https://github.com/dimagi/commcare-core/pull/459

Since we were already storing these variables in FormSession and they're not useful for Android (where its tracked separately) I refactored this a bit. Would be good to validate this @biyeun @benrudolph


biyeun:
✅

biyeun:
waiting on build

PR #188, opened 2016-11-18
Unlock even when exception is thrown

wpride:


benrudolph:
Sick glad this got tracked down

PR #189, opened 2016-11-18
don't crash hard on deserialization of incomplete forms

wpride:


wpride:
(self merging to enable this as a holdover)

PR #192, opened 2016-11-22
re-add formlayer config engine to return install errors

wpride:
@benrudolph this will allow us to return the pretty errors again


benrudolph:
sweet

PR #193, opened 2016-11-22
Route for app updates and reloading form session

wpride:
cross-request https://github.com/dimagi/commcare-hq/pull/14004


benrudolph:
also, i talked to clayton about this and he says that this won't actually improve anything since we aren't versioning the applications between ccz pulls. still think it's good to have so we're ready when HQ supports versioning when downloading the direct ccz

PR #197, opened 2016-11-28
add form validation endpoint

snopoke:
The idea here is to get rid of the current form translate jar and rather have HQ just call this endpoint to do form validation. 

What do you think?

@benrudolph @wpride


benrudolph:
oh nice this would be awesome. looks good to me, nice auth cleanup

wpride:
this is great @snopoke

PR #198, opened 2016-11-29
Add actuator

benrudolph:
@wpride this provides some incredibly useful endpoints on the port specified. this will not be open to the world. read more here: http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-endpoints. once this is deployed i think it makes sense to replace our servup check with `:8081/health`


snopoke:
with `address=127.0.0.1` will this be accessible from off the machine?

benrudolph:
hmm not positive. the ansible portion (not yet PR'd) will be the one that sets these values. i'll use 0.0.0.0 just in case

PR #199, opened 2016-11-29
use proper delete. delete DBs when install fails

wpride:


src/main/java/services/impl/InstallServiceImpl.java
```diff
@@ -23,23 +25,28 @@
...
+            });
+
+            CommCareConfigEngine engine = new FormplayerConfigEngine(PrototypeManager.getDefault());
+            engine.initFromArchive(reference);
+            engine.initEnvironment();
+            return engine;
+        } catch (Exception e) {
+            log.error("Got exception " + e + " while installing reference " + reference + " at path " + dbPath);
+            SqlSandboxUtils.deleteDatabaseFolder(dbPath);
+            throw e;
```
benrudolph: do we need to convert this to a runtime exception so that it's handled correctly?

wpride: I don't think so - my guess is that this was handling the initial error (the one we're throw here) correctly already, then the subsequent ones were throwing that obfuscated NPE

benrudolph: 👍

PR #201, opened 2016-11-30
don't get restore XML for each session

wpride:
We were looking up the restoreXml each time we looked up a session, even if this user already had extant databases. fix that.


src/main/java/hq/CaseAPIs.java
```diff
@@ -29,15 +29,17 @@
...
         return restoreIfNotExists(restoreFactory);
     }
 
     public static UserSqlSandbox restoreIfNotExists(RestoreFactory restoreFactory) throws Exception{
         File db = new File(restoreFactory.getDbFile());
         if(db.exists()){
+            System.out.println("db exists for path " + db);
             return restoreFactory.getSqlSandbox();
         } else{
+            System.out.println("NOT db exists for path " + db);
```
benrudolph: print

PR #206, opened 2016-12-01
Revert "Revert "Reuse application DBs""

benrudolph:
Reverts dimagi/formplayer#205


benrudolph:
@wpride

wpride:
@benrudolph stupid mistake wasn't resetting the storage factory on update so if you used an app -> used a different app -> went back to first and reused DBs, you would be using the old storageManager and the old DBs.

PR #207, opened 2016-12-01
Storage layer refactor

wpride:
Fairly big refactor to get us out of static singleton storage hell. Hopefully this should clear up a good deal of the "Storage not registered" exceptions.

Primarily this overrides the InstallerFactory and *ResourceInstaller classes to access Formplayer's user-specific SQL storage factories instead of the static StorageManager's version. This code is also moved into factory beans.

At the CommCare level there are a few refactors to allow this, the biggest being the removal of the (as far as I can tell) useless CommCareInstance interface which allows us to access the storage layer through the platform when necessary (instead of the static helper). 

@benrudolph


benrudolph:
(let me know when this is ready for review)

src/main/java/application/IncompleteSessionController.java
```diff
@@ -20,8 +20,10 @@
 import session.FormSession;
 import util.Constants;
 
-import java.util.*;
-import java.util.concurrent.locks.Lock;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
```
benrudolph: nit: not sure why this file's imports changed

src/main/java/db/migration/V10__form_session_add_app_id.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Add appId to form session
+ */
+public class V10__form_session_add_app_id implements SpringJdbcMigration {
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        jdbcTemplate.execute("ALTER TABLE formplayer_sessions " +
+                "ADD appid VARCHAR DEFAULT '1'");
```
benrudolph: should the default not by NULL so we don't get confused if there was an actual app associated with the form?

src/main/java/installers/FormplayerXFormInstaller.java
```diff
@@ -0,0 +1,52 @@
...
+        storageFactory = new FormplayerStorageFactory();
+        storageFactory.configure(databasePath, trimmedUsername);
+
+    }
+
+    @Override
+    public void writeExternal(DataOutputStream out) throws IOException {
+        super.writeExternal(out);
+        ExtUtil.writeString(out, ExtUtil.emptyIfNull(storageFactory.getDatabasePath()));
+        ExtUtil.writeString(out, ExtUtil.emptyIfNull(storageFactory.getTrimmedUsername()));
```
benrudolph: all these writeExternal and readExternal seem to be very similar, could make it slightly DRYer

src/main/java/objects/SerializableFormSession.java
```diff
@@ -29,6 +27,7 @@
     private boolean oneQuestionPerScreen;
     private String asUser;
     private String currentIndex = "0";
+    private String appId;
```
benrudolph: 😂 😂 😂

src/main/java/application/IncompleteSessionController.java
wpride: I just ran optimize imports across the source set

src/main/java/installers/FormplayerXFormInstaller.java
wpride: Agreed - unfortunately the extant inheritance pattern and lack of multiple inheritance makes that slightly hard

PR #210, opened 2016-12-05
Update subrepo to use ThreadLocal<> wrapped Localizer

wpride:
@benrudolph I think this will resolve our Localization issues until we can get a more complete refactor in place. https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html


benrudolph:
awesome! 💀 to `The Localizer could not find a definition for ID: modules.m12 in the 'fra' locale.`

PR #216, opened 2016-12-08
OQPS: Skip first index if it is a Group

biyeun:
@benrudolph / @wpride 
fixes http://manage.dimagi.com/default.asp?239709


wpride:
Nice catch! Kicking build, fail was an I/O issue

PR #221, opened 2016-12-12
Reuse DBs

wpride:
@benrudolph changes needed to not all user data when Formplayer is booted up. 
Moved test DBs into the testdbs/ folder so that we can clear those

At the moment re-using application DBs between runtimes will fail because the ReferenceManager() contains static mappings for URI's to the user archives. Currently this will fail elegantly and just clear the old DBs/reinstall. Fixing this will require refactoring the ReferenceManager() class or backing its mappings up to somewhere persistent. I'm eying that class as a source of the concurrency issues we're seeing too.

User DBs can be reused just fine

### DIFF
--- a/src/main/java/application/Application.java
+++ b/src/main/java/application/Application.java
@@ -40,8 +40,6 @@ public class Application {
     public static void main(String[] args) {
         ConfigurableApplicationContext context = SpringApplication.run(Application.class, args);
         migrate();
-        // wipe out all user data for now. We need this until we can fix up some static mapping objects.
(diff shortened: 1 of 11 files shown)


src/main/java/application/Application.java
```diff
@@ -40,8 +40,6 @@
     public static void main(String[] args) {
         ConfigurableApplicationContext context = SpringApplication.run(Application.class, args);
         migrate();
-        // wipe out all user data for now. We need this until we can fix up some static mapping objects.
```
benrudolph: 🙌

PR #224, opened 2016-12-13
Better logging and error handling around menu nav

wpride:
This adds some logging and better error messaging around the exception Jessica reported:

java.lang.ArrayIndexOutOfBoundsException: 2
   at org.commcare.util.screen.MenuScreen.handleInputAndUpdateSession(MenuScreen.java:158)
   at session.MenuSession.handleInput(MenuSession.java:167)

the logs around this are very strange:

Request to navigate_menu with bean SessionNavigationBean [id= 3d45b792-cf0e-4d61-9ca5-60e5e74e7deb, selections=[2, 41df827f-bf39-41a5-804d-d0e17836b455, 74a8153d-2bde-4069-b3a9-48ab6a5243b8] parent=InstallRequestBean: [installReference=null, username=asi-test@asi-selever.commcarehq.org, domain=asi-selever, appId=cb33358bb543442b0fc14990358f7118, oneQuestionPerScreen: false]]
2016-12-13 11:01:59.936  INFO 30929 --- [nio-8181-exec-7] services.impl.InstallServiceImpl         : Configuring application with reference https://www.commcarehq.org/a/asi-selever/apps/api/download_ccz/?app_id=cb33358bb543442b0fc14990358f7118&latest=save and dbPath: /opt/data/formplayer/asi-selever/asi_test@asi_selever_commcarehq_org/null.

In the database path the appId is null (presumably what's causing this issue) but in the request the appId is present. Further, in the database path the username and domain are set fine so this doesn't look like a concurrency issue. Stranger things

### DIFF
--- a/src/main/java/services/FormplayerStorageFactory.java
+++ b/src/main/java/services/FormplayerStorageFactory.java
@@ -2,8 +2,6 @@ package services;
 
 import beans.InstallRequestBean;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.commcare.api.persistence.SqliteIndexedStorageUtility;
 import org.javarosa.core.services.storage.IStorageIndexedFactory;
 import org.javarosa.core.services.storage.IStorageUtilityIndexed;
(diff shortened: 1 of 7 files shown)


src/main/java/services/FormplayerStorageFactory.java
```diff
@@ -29,6 +29,10 @@ public void configure(InstallRequestBean authenticatedRequestBean) {
     }
 
     public void configure(String username, String domain, String appId) {
+        if(username == null || domain == null || appId == null) {
+            throw new RuntimeException(String.format("Cannot configure FormplayerStorageFactory with null arguments. " +
```
benrudolph: nice like that we throw this early. might be a good use case for contracts

PR #226, opened 2016-12-13
Fix another concurrency issue

wpride:


wpride:
adding don't merge, seeing some weird behavior

PR #231, opened 2016-12-28
Increase lock expiration time, don't crash when it timesout

wpride:
Partial fix for http://manage.dimagi.com/default.asp?244524#

Before we were expiring the lock after 60 seconds of ownership (the default) so if the request took longer than this the request would fail (even though the result was returned correctly). This bumps that expiry time to five minutes. 

I've also made it so that we won't crash and discard the result if we've lost the lock. I'm not sure if this is the right thing to do since there are potential concurrency issues (IE another thread could have gotten the lock and changed the DBs). However, this seems exceedingly unlikely to me since 1) the expiry period is now five minutes and 2) now that app and user DBs are used data churn seems much less likely.

In fact I was considering whether we can disregard the user locking entirely now that we're reusing DBs and handling concurrency more cleanly, but that's a separate discussion.

CC @czue @benrudolph

### DIFF
--- a/src/main/java/aspects/LockAspect.java
+++ b/src/main/java/aspects/LockAspect.java
@@ -1,8 +1,6 @@
 package aspects;
 
 import beans.AuthenticatedRequestBean;
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.annotation.Around;
 import org.aspectj.lang.annotation.Aspect;
@@ -19,8 +17,6 @@ import java.util.concurrent.locks.Lock;
  */
 @Aspect
 public class LockAspect {
-
-    private final Log log = LogFactory.getLog(LockAspect.class);
 
     @Autowired
     protected LockRegistry userLockRegistry;
@@ -43,7 +39,9 @@ public class LockAspect {
         try {
             return joinPoint.proceed();
         } finally {
-            lock.unlock();
+            if (lock != null) {
+                lock.unlock();
+            }
         }
     }
 
--- a/src/main/java/aspects/LockAspect.java
+++ b/src/main/java/aspects/LockAspect.java
@@ -1,6 +1,8 @@
 package aspects;
 
 import beans.AuthenticatedRequestBean;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


benrudolph:
Nice, this is great. I could've sworn we increased the lock timeout at some point, though not to as high as 5 mins. I'm thinking 5 minutes should be our max (and even that is pretty high). We've gotten to point in HQ where we accept 15 minute requests, which is not an ideal place to be in.

> I've also made it so that we won't crash and discard the result if we've lost the lock.

I'm not positive this is the best route to go. I think it's more clear to just fail hard if we hit the 5 min mark and say "Request took too long." At some point the fix isn't to increase the lock timeout but optimize the code. If some app some day takes 10 minutes to load, then we'll be seeing harder to track down errors than the "Lock expired" one, which is annoying, but easily decipherable.

I think I'd have to think more/discuss with you about discarding concurrency totally. As far as I'm aware you can still have two browser tabs open at the same time and be manipulating the form session at the same time

src/main/java/aspects/LockAspect.java
```diff
@@ -40,7 +44,12 @@ public Object beforeLock(ProceedingJoinPoint joinPoint) throws Throwable {
             return joinPoint.proceed();
         } finally {
             if (lock != null) {
-                lock.unlock();
+                try {
+                    lock.unlock();
+                } catch (IllegalStateException e) {
+                    // Don't crash here, still return result
```
benrudolph: this is up for debate, but i think we should crash here. @czue thoughts?

wpride:
@benrudolph we increased the timeout for acquiring the lock but not for possessing it as far as I know

benrudolph:
aha makes sense

src/main/java/aspects/LockAspect.java
czue: not sure i have enough context to weigh in. but if this is basically a 5 minute timeout then I agree crashing and saying it timed out seems fine for now - especially if doing whatever is being done here results in some inconsistent state or undefined behavior.

src/main/java/aspects/LockAspect.java
```diff
@@ -39,9 +43,7 @@ public Object beforeLock(ProceedingJoinPoint joinPoint) throws Throwable {
         try {
             return joinPoint.proceed();
         } finally {
-            if (lock != null) {
```
benrudolph: do we still need this check?

PR #232, opened 2017-01-02
Refactor formplayer breadcrumbs

benrudolph:


src/main/java/beans/exceptions/BaseExceptionResponseBean.java
```diff
@@ -10,6 +10,8 @@
     protected String status;
     protected String url;
     protected String type;
+    // For backwards compatibility with old CloudCare, we need to add the message property
```
wpride: Do we care about this now that we disabled the "Use formplayer server" flag?

benrudolph: unfortunately (or fortunately) both old and new cloudcare utilize the same form entry javascript code and that javascript code expects `message`. we could just change the JS to look for `.exception` as well

benrudolph: actually changing the javascript code is easier

benrudolph:
updated, can be merged before the other PR

PR #233, opened 2017-01-03
module relevancy condition fix, test

wpride:
Fix for http://manage.dimagi.com/default.asp?244868

We were evaluating relevancy conditions even when the menu wasn't going to be displayed because its root wasn't the current screen.


benrudolph:
test failure looks real, but otherwise looks good on ✔️ . nice find

wpride:
@benrudolph This evolved a little bit. I implemented a real fix yesterday on commcare-core that centralized the menu navigation logic. The error from the test was real so I resolved by merging in the fix I made last night. That fix also included some error handling that uses platform-wide localized strings, so I added those the Formplayer as well. Then I added the nepali and ethiopian month lists to those as well since Jeremy had that bug where they were missing

The TLDR is we can add strings to the formplayer_translatable_strings.txt file which will be localizable across Formplayer now

src/main/resources/formplayer_translatable_strings.txt
```diff
@@ -0,0 +1,4 @@
+app.menu.display.cond.bad.xpath=Invalid form display condition (${0}): ${1}
+app.menu.display.cond.xpath.err=Error evaluating form display condition (${0}): ${1}
+ethiopian_months=Mäskäräm,T’ïk’ïmt,Hïdar,Tahsas,T’ïr,Yäkatit,Mägabit,Miyaziya,Gïnbot,Säne,Hämle,Nähäse,P’agume
+nepali_months=Baishakh,Jestha,Ashadh,Shrawan,Bhadra,Ashwin,Kartik,Mangsir,Poush,Magh,Falgun,Chaitra
```
benrudolph: oo interesting

PR #236, opened 2017-01-04
split case list and case details into separate calls

wpride:
@benrudolph I think this is all we need on the backend. The model returned should be identical. Hold off on merge until front end is merged because this will blank all details on its own.

fixes http://manage.dimagi.com/default.asp?244524

cc @czue


src/main/java/application/MenuController.java
```diff
@@ -84,6 +88,46 @@ public BaseResponseBean updateRequest(@RequestBody UpdateRequestBean updateReque
...
+        Screen currentScreen = menuSession.getNextScreen();
+        if (!(currentScreen instanceof EntityScreen)) {
+            throw new RuntimeException("Tried to get details while not on a case list.");
+        }
+        EntityScreen entityScreen = (EntityScreen) currentScreen;
+        TreeReference reference = entityScreen.resolveTreeReference(detailSelection);
+
+        if (reference == null) {
+            throw new RuntimeException("Could not find case with ID " + detailSelection);
+        }
```
benrudolph: 👍 for errors

benrudolph:
conflix

PR #237, opened 2017-01-05
Use statsdclient for metrics

benrudolph:
@wpride @snopoke  gave up trying to get datadog's jmxfetch utility to work with spring actuator and just wrote the metrics myself. there's a write up here: https://github.com/DataDog/jmxfetch/issues/118 of the issue i was facing. think this is better in the end anyways because it will allow us to capture custom metrics if we want. if the statsd client isn't running, it just silently fails which is what we want

cc: @millerdev @gcapalbo

### DIFF
--- a/src/main/java/application/WebAppContext.java
+++ b/src/main/java/application/WebAppContext.java
@@ -2,6 +2,9 @@ package application;
 
 import aspects.LockAspect;
 import aspects.LoggingAspect;
+import aspects.MetricsAspect;
+import com.timgroup.statsd.NonBlockingStatsDClient;
+import com.timgroup.statsd.StatsDClient;
 import installers.FormplayerInstallerFactory;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -169,6 +172,16 @@ public class WebAppContext extends WebMvcConfigurerAdapter {
     @Bean
     public static PropertySourcesPlaceholderConfigurer propertiesResolver() {
         return new PropertySourcesPlaceholderConfigurer();
+    }
+
+    @Bean
+    public StatsDClient datadogStatsDClient() {
+        return new NonBlockingStatsDClient(
+                "formplayer.metrics",
+                "localhost",
+                8125,
+                new String[] {"tag:formplayer"}
+        );
     }
 
     @Bean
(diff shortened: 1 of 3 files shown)


src/main/java/aspects/MetricsAspect.java
```diff
@@ -0,0 +1,48 @@
...
+        if (args[0] instanceof AuthenticatedRequestBean) {
+            AuthenticatedRequestBean bean = (AuthenticatedRequestBean) args[0];
+            domain = bean.getDomain();
+            user = bean.getUsernameDetail();
+        }
+
+        datadogStatsDClient.increment("requests", domain, user, requestPath);
+        long startTime = System.nanoTime();
+        Object result = joinPoint.proceed();
+        datadogStatsDClient.gauge("timings", (System.nanoTime() - startTime) / 1000000, domain, user, requestPath);
```
benrudolph: the `/ 100000` is just converting nanosecond to milliseconds

gcapalbo:
looks ok to me, will give @wspride a chance to take a pass before merging

wpride:
@benrudolph stud! 💰

src/main/java/application/WebAppContext.java
```diff
@@ -172,6 +175,16 @@ public static PropertySourcesPlaceholderConfigurer propertiesResolver() {
     }
 
     @Bean
+    public StatsDClient datadogStatsDClient() {
+        return new NonBlockingStatsDClient(
+                "formplayer.metrics",
+                "localhost",
+                8125,
+                new String[] {"tag:formplayer"}
```
snopoke: `tag` here is meant to be the name of the tag. Since the metric is already prefixed with `formplayer` this tag doesn't seem useful.

src/main/java/aspects/MetricsAspect.java
snopoke: these tags should be formatted with the tag name as follows `domain:{domain}` etc.

snopoke:
nice, just need to fix the tag formatting.

benrudolph:
ok cool, i realized the error in my tagging ways: https://github.com/dimagi/formplayer/pull/239 and updated here. good point on the general tag, will update on that PR. this PR unfortunately doesn't work, still trying to figure out why

benrudolph:
hot damn! it is actually working!

benrudolph:
but we have a tag for every domain... lol

PR #239, opened 2017-01-06
better tags and use recordExecution time

benrudolph:
@wpride this didn't work which was disappointing. trying to figure out why not, but in the mean time i realized the tags could be better


snopoke:
what do you mean it didn't work? I can see the stats in datadog

benrudolph:
i didn't see the metrics at first, but i probably just didn't wait long enough

PR #240, opened 2017-01-06
Actually order by date

benrudolph:
@wpride realized we weren't actually ordering by date. since dateOpened is stored as text as a string like `"Fri Jan 06 15:58:03 SAST 2017`. thus when we order by dateopened it orders by the day of the week. i think at some point we should move dateOpened to actual date time column stored as an epoch or in UTC format. tried writing a test for this, but had trouble mocking the time, but verified it works locally.


benrudolph:
we also need to paginate this at some point. (not going to do that in this PR)

benrudolph:
g2m

PR #242, opened 2017-01-09
Metrics for errors

benrudolph:
Adding some more metrics. I'm envisioning a crash report dashboard @wpride @snopoke

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -5,6 +5,7 @@ import beans.exceptions.ExceptionResponseBean;
 import beans.exceptions.HTMLExceptionResponseBean;
 import beans.exceptions.RetryExceptionResponseBean;
 import beans.menus.*;
+import com.timgroup.statsd.StatsDClient;
 import exceptions.ApplicationConfigException;
 import exceptions.AsyncRetryException;
 import exceptions.FormNotFoundException;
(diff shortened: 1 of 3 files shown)


src/main/java/application/AbstractBaseController.java
```diff
@@ -289,6 +298,15 @@ public ExceptionResponseBean handleError(FormplayerHttpRequest req, Exception ex
         return new ExceptionResponseBean(exception.getMessage(), req.getRequestURL().toString());
     }
 
+    private void incrementDatadogCounter(String metric, FormplayerHttpRequest req) {
+        datadogStatsDClient.increment(
+                metric,
+                "domain:" + req.getDomain(),
+                "user:" + req.getCouchUser().getUsername(),
```
benrudolph: assumption here is that if we ever have a `FormplayerHttpRequest` then we have a user. let me know if you think that's a bad assumption

snopoke: I don't think form validation requests will have a user

wpride: What about serverup?

benrudolph: presumably serverup will never go through error handling, but maybe we should guard against that?

benrudolph: gonna do it for due diligence

PR #244, opened 2017-01-11
Make force restore overwrite cache

benrudolph:
@czue @wpride http://manage.dimagi.com/default.asp?245357#1272630 I _think_ this is the root cause. trying to validate. next step is figuring out if it's safe to deploy formplayer. we may need to make a PR off of the commit deployed to icds


czue:
nice, makes sense to me. build is failing, not sure if related

benrudolph:
✅ can anyone who merges this use squash commit (or just approve and i'll squash)

benrudolph:
actually hold off for a second

wpride:
Nice tests refactor

PR #247, opened 2017-01-12
Fix for recursive filter calls

wpride:
Have our filters extend a class that guarantees to only fire once, hopefully will resolve these recursive filtering overflows we saw. Also makes the code tidier.


src/main/java/application/FormplayerAuthFilter.java
```diff
@@ -47,31 +48,23 @@
...
             setDomain(request);
             JSONObject data = RequestUtils.getPostData(request);
             if (!authorizeRequest(request, data.getString("domain"), getUsername(data))) {
-                setResponseUnauthorized((HttpServletResponse) res);
+                setResponseUnauthorized(response);
                 return;
             }
         }
-
-        chain.doFilter(request, res);
```
benrudolph: did you mean to remove this?

benrudolph:
sweet big fan

PR #250, opened 2017-01-16
Don't crash on here() function

wpride:
This was actually already being overridden with a dummy function in the CLI (thanks Phillip) just needed to properly get the EvaluationContext from the screen

### DIFF
--- a/src/test/java/tests/GeoTests.java
+++ b/src/test/java/tests/GeoTests.java
@@ -1,0 +1,33 @@
+package tests;
+
+import beans.menus.EntityListResponse;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.test.context.ContextConfiguration;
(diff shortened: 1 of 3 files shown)


src/test/java/tests/GeoTests.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    protected String getMockRestoreFileName() {
+        return "restores/geo.xml";
+    }
+
+    // validate that we don't crash on here() function
+    @Test
+    public void testHereOverride() throws Exception {
+        sessionNavigate(new String[]{"5", "1"}, "basic", EntityListResponse.class);
```
benrudolph: where is the here function used?

wpride: The case list that shows up uses the here() function

PR #251, opened 2017-01-16
Don't revalidate answers on submit

wpride:
Discussed with @benrudolph and we don't think that we need this double check anymore


wpride:
@benrudolph That script never submitted forms just traversed menus and opened forms. I think we'd want to test this on the JS side to convince ourselves that the code will never submitted un-validated answers

benrudolph:
yes agree about the JS bit, gotta make sure we do indeed wait for all answers to complete before submitting

PR #255, opened 2017-01-17
Allow profile reference in installReference

wpride:
This still won't work for remote profile installs (but won't regress anything)


wpride:
I think so because I didn't see the utility at the time

PR #258, opened 2017-01-18
Fix for missing appIds

wpride:
Some migrated incomplete forms were missing appId's which caused errors as in http://manage.dimagi.com/default.asp?245942

This migration addresses that by using either the linked menu session or the postUrl depending on what is available. This migration is very janky and I am not proud of that. However its also very resourceful and I am proud of that 😑

### DIFF
--- a/src/main/java/db/migration/V11__form_session_add_missing_appid.java
+++ b/src/main/java/db/migration/V11__form_session_add_missing_appid.java
@@ -8,8 +8,8 @@ import org.springframework.jdbc.core.ResultSetExtractor;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
-import java.util.function.Consumer;
 
 /**
  * Attempt to fix null appId in form session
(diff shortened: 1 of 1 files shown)


src/main/java/db/migration/V11__form_session_add_missing_appid.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Attempt to fix null appId in form session
+ */
+public class V11__form_session_add_missing_appid implements SpringJdbcMigration {
+    @Override
+    public void migrate(final JdbcTemplate jdbcTemplate) throws Exception {
+        String sql = "SELECT * FROM formplayer_sessions WHERE COALESCE(appid, '') = ''";
```
benrudolph: `appid is NULL` is the idiomatic way to get null values

wpride: That wasn't working for me actually

wpride: http://stackoverflow.com/questions/14172978/postgresql-select-empty-fields

benrudolph: well i'll be damned

PR #259, opened 2017-01-20
add session list item tostring

wpride:
This has been bothering me in the logs


wpride:
Also removing the form tree in answer response - just cluttering and I've never used it

PR #260, opened 2017-01-20
Fix AppFixture install location 

wpride:
Fixes http://manage.dimagi.com/default.asp?245966

This feels hacky/dirty but I think does the right thing. Basically, instead of the app fixtures (IE from the suite) being installed in the user's db folder, we need them to be installed in the app's db folder, so we use the StorageManager to accomplish this. The usual dirtiness involved with these static managers is in full play.


benrudolph:
nice agree, but seems necessary

PR #261, opened 2017-01-20
Change StorageManager static to StorageManager thread safe singleton

wpride:
Don't know how I missed this one on my previous static object pass. Should fix the GLOBAL_RESOURCES_TABLE exception that we've been seeing. Hopefully the last of these.


benrudolph:
sweet was wondering about those

PR #262, opened 2017-01-20
Respect sync freq

benrudolph:
@wpride this makes formplayer respect HQ's sync frequency http://manage.dimagi.com/default.asp?245997

cc: @emord


wpride:
Shibby!

benrudolph:
what a horrible horrible experience. i think the tests are passing at last, though i could never get the mock to work properly in the tests where the restoreFactory was instantiated anew

PR #264, opened 2017-01-23
Catch and rethrow NPE

wpride:
@benrudolph this should cause us to display a proper error message and nav backwards when the 'null' ends up in the url (somehow)


benrudolph:
👍 sounds good. i think we should still try and find out how this thing is being set to `null`

PR #265, opened 2017-01-23
Need to set storageFactory now for form controller apis

wpride:
We need to configure the storageFactory in the FormController routes now due to https://github.com/dimagi/formplayer/pull/260


benrudolph:
hot fix, merging

benrudolph:
this actually seems ripe for error, should we configure the storage factory in the FormSession?

PR #266, opened 2017-01-23
update all spring libs

wpride:


benrudolph:
gave it a spin, seems fine to me

PR #267, opened 2017-01-23
Fix appId in menu session

wpride:
Use appId from the request instead of from the profile. This was not correct and was resulting in re-installations happening on some submissions (also explaining why submit times were so long for some requests).


wpride:
:shame:

PR #270, opened 2017-01-24
Don't instantiate formsession object for each session

wpride:
Should give us performance and robustness wins

### DIFF
--- a/src/main/java/beans/SessionListItem.java
+++ b/src/main/java/beans/SessionListItem.java
@@ -1,7 +1,14 @@
 package beans;
 
+import hq.CaseAPIs;
+import objects.SerializableFormSession;
+import org.commcare.api.persistence.SqliteIndexedStorageUtility;
+import org.commcare.cases.model.Case;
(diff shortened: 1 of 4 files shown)


src/main/java/beans/SessionListItem.java
```diff
@@ -13,11 +20,26 @@
...
+        String caseId = session.getSessionData().get("case_id");
+        if (caseId == null) {
+            return null;
+        }
+        try {
+            CaseBean caseBean = CaseAPIs.getFullCase(caseId, caseStorage);
+            return (String) caseBean.getProperties().get("case_name");
+        } catch (NoSuchElementException e) {
+            // This handles the case where the case is no longer open in the database.
+            // The form will crash on open, but I don't know if there's a more elegant but not-opaque way to handle
```
benrudolph: Hmm we should think about the best way to do this

wpride: I think there's no best way - Incomplete Forms are just terrible!

PR #277, opened 2017-01-25
Case list optimization

wpride:
Previously we were evaluating every case list field for every case in user's DB before showing the case list even when this list might be filtered and would certainly be paginated. This changes that pushing the row loading back in the EntityListScreen class until they're needed and by making the filter and paginate operations work on the list of pre-rendered Entity<TreeReference> objects instead of the fully actualized rows. Required moving some code down from -android into -core.

This should be a massive performance win for enikshay

@benrudolph cc @czue

### DIFF
--- a/src/main/java/beans/menus/EntityListResponse.java
+++ b/src/main/java/beans/menus/EntityListResponse.java
@@ -95,31 +95,7 @@ public class EntityListResponse extends MenuBean {
                                  int offset,
                                  String searchText) {
         EntityBean[] allEntities = generateEntities(screen, references, ec, offset, searchText);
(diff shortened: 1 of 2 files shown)


benrudolph:
sweet this is awesome. diff's a bit hard to read but looks good. assuming everything in the submodule commits are just moved code

czue:
nice, sounds like a huge speedup boost! build is failing fyi

wpride:
Tests are broken on master, fixed here https://github.com/dimagi/formplayer/pull/278

wpride:
@benrudolph gtm

benrudolph:
⚡️ ⚡️ ⚡️

PR #278, opened 2017-01-25
fix for tests broken when TF rotues removed

wpride:
Broken here, not sure how this build passed https://github.com/dimagi/formplayer/pull/272


benrudolph:
👍

PR #280, opened 2017-01-30
Make evaluate instance work better

wpride:
This aligns our code path with the one used by the CLI, in particular it allows us to return multiple nodes in the case that the expression matches multiple.


benrudolph:
😂 😂

PR #281, opened 2017-01-30
Return current frame stack with each menu nav

wpride:
Returns for example:

"sessionSteps":["COMMAND: m2","COMMAND: m2-f1"]

First step for having the debugger show menu nav.

### DIFF
--- a/src/main/java/session/MenuSession.java
+++ b/src/main/java/session/MenuSession.java
@@ -12,6 +12,7 @@ import org.commcare.session.CommCareSession;
 import org.commcare.session.SessionFrame;
 import org.commcare.suite.model.FormIdDatum;
 import org.commcare.suite.model.SessionDatum;
+import org.commcare.suite.model.StackFrameStep;
 import org.commcare.util.CommCarePlatform;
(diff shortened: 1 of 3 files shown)


src/main/java/session/MenuSession.java
```diff
@@ -334,4 +335,19 @@ public boolean isOneQuestionPerScreen() {
     public void setOneQuestionPerScreen(boolean oneQuestionPerScreen) {
         this.oneQuestionPerScreen = oneQuestionPerScreen;
     }
+
+    public String[] getSessionStack() {
+        ArrayList<String> steps = new ArrayList<>();
+        SessionFrame frame = sessionWrapper.getFrame();
+        for (StackFrameStep step : frame.getSteps()) {
+            if (step.getType().equals(SessionFrame.STATE_COMMAND_ID)) {
+                steps.add("COMMAND: " + step.getId());
```
benrudolph: to make this nicer for the frontend to parser it might make sense to return an array of objects, maybe something like
```
{
   stepType: 'command',
   value: 'm1-f0',
}
```

PR #283, opened 2017-01-31
Graphing in Nimbus

wpride:
Returns graph HTML/JS. WIP

cross-request: https://github.com/dimagi/commcare-hq/pull/14870


src/main/java/beans/menus/EntityListResponse.java
```diff
@@ -187,7 +190,15 @@ private EntityBean processEntity(TreeReference entity, EntityScreen screen, Eval
         for (DetailField field : fields) {
             Object o;
             o = field.getTemplate().evaluate(context);
-            data[i] = o;
+            if(o instanceof GraphData) {
+                try {
+                    data[i] = GraphUtil.getHTML((GraphData) o, "");
+                } catch (GraphException e) {
+                    data[i] = "Error loading graph " + e;
```
benrudolph: don't love this pattern, it makes it hard to tell which datums have failed graphs

benrudolph:
don't think this is actually as hard as you think. guessing all we need to do is call `.html()` on that blob and ensure that c3.js is imported. the one sticky part is the `<script>` tags. it'd be great if there was an option to remove those

benrudolph:
also not positive what graphing.min.js is

orangejenny:
hooray

src/main/java/beans/menus/EntityListResponse.java
benrudolph: if we're going to stick with this, this should at the least return valid html
`<html><body>Error loading...</body></html>`

PR #290, opened 2017-02-15
Ignore this goddamn exception

wpride:
Really this is the wrong exception to be thrown but the fix is deep in the engine and I don't think this will ever be a "real" 500 so

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -23,6 +23,7 @@ import org.apache.commons.mail.HtmlEmail;
 import org.commcare.core.process.CommCareInstanceInitializer;
 import org.commcare.modern.models.RecordTooLargeException;
 import org.commcare.modern.session.SessionWrapper;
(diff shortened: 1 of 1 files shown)


czue:
haha this has gotta be related to case claim

wpride:
@czue haha nope this is ~entirely related to app config errors, usually misconfigured parens or quotes in the suite that form builder misses. Occasionally we get bombed with them

orangejenny:
A+ PR naming, I can feel your pain all the way from Boston.

snopoke:
Can you give an example? Might be able to fix it on HQ as well.

wpride:
lol take your pick: https://groups.google.com/a/dimagi.com/forum/#!searchin/commcarehq-ops/UnresolvedResourceException%7Csort:date

Derek seems to have had about forty tonight

snopoke:
Hmm, I think it would be expensive for HQ to validate all of those unless we build a bulk xpath validation endpoint into Nimbus.

PR #293, opened 2017-02-16
use the correct response format to not error weirdly and wildly

wpride:
@benrudolph


wpride:
Correct. Maybe I'll keep adding really dumb code so I can remove it and get kudos

benrudolph:
Just ran into this, would be great to get this in. Tests are failing

PR #294, opened 2017-02-16
Write response when we get an authorized request

benrudolph:
@wpride it's super confusing when a blank response is returned when the request is invalid. this gives the response body a message.

Before:
```
HTTP/1.1 401
Content-Length: 0
Date: Thu, 16 Feb 2017 14:09:07 GMT

```
After:
```
HTTP/1.1 401
Content-Type: application/json;charset=ISO-8859-1
Transfer-Encoding: chunked
Date: Thu, 16 Feb 2017 14:11:30 GMT

{"error":"Invalid session id"}
```


wpride:
look at us error handling like mature citizens

PR #301, opened 2017-02-23
Fix indexed fixtures

wpride:
This tests and fixes the bug here http://manage.dimagi.com/default.asp?248158

Our SQLite database objects are agnostic as to whether the database backing them has been created or not - this allows us to reference tables by username and appId and get access to the tables if they exist and create them otherwise. The tables backing indexed fixtures thus created a problem

1. When you go to access the table you don't have access to the exact model you're instantiating because this model is dynamic (dependent on the table's indices).
2. To get around this, Phillip implemented a map between the table name and the table handle so that we wouldn't have to re-instantiate. However, when this sandbox was re-created on second access this map was, of course, null

This PR adds a new SQLiteIndexedStorageUtility path that allows you to access the DB without a prototype of the exact model and uses this to align the indexed fixture storage with our extant SQL concept.

Also includes some cleanup and a regression test for the ticket failure (required a second access of the indexed fixture DBs)


src/test/java/tests/IndexedFixtureTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    protected String getMockRestoreFileName() {
+        return "restores/indexed.xml";
+    }
+
+    @Test
+    public void testBadCaseSelection() throws Throwable {
+        sessionNavigate(new String[]{"0", "action 0"}, "indexed", null, NewFormResponse.class, "test");
+        sessionNavigate(new String[]{"1", "d83cf96f-6f7d-40e1-b3eb-ee8b2b8fed0f", "0"}, "indexed", null, NewFormResponse.class, "test");
```
benrudolph: what does this test? just that it doesn't error?

wpride: yes

PR #303, opened 2017-02-24
Force restore when loading a new form

benrudolph:
@wpride this ensure we do a restore on editing a form


wpride:
I don't think we want this, this will cause a restore to be triggered every time someone menu navs from a form list to a form

wpride:
Wait NVM I think you're right, totally separate route

benrudolph:
Yeah that's the conclusion I came to

benrudolph:
(i think our tests are broken)

wpride:
@benrudolph I think this change is actually causing that

benrudolph:
haha shit, will take a look. test didn't seem related

wpride:
Just guessing based on that test using this endpoint (new-form). Actually
not sure if test was always wrong and was testing the wrong thing... very
possible

PR #305, opened 2017-03-01
Fix for model iteration

wpride:
Love gnarly bugs that end up being a one line fix. This addresses https://manage.dimagi.com/default.asp?248557#1296003

This form uses model iteration, so the model is meant to be expanded when we first walk it based on the case db state. We were expanding this; however, we were saving the form state before the walk and so losing the expanded model. This then crashed when you tried to answer a question that had previously existed, but now does not.


benrudolph:
wow nice

PR #306, opened 2017-03-01
Skip any detail screens with nodesets

wpride:
Handling detail screens with a nodeset defined (IE a table) will require some work. For now we can skip these and show the other details.


wpride:
@benrudolph This was causing the mobile UCR screens to error. Specifically, mobile UCR tables have a special display where you iterate over a subset of the selected entity. This will be harder to implement. This PR will just skip those tables for now so we can still show any other tabs IE graphs

PR #310, opened 2017-03-07
Fix OQPS sub-group indexing issue

wpride:
fixes https://manage.dimagi.com/default.asp?247996

In OQPS when you answered a question in a sub-group, the currentIndex would focus on that specific answer (as if it had not been in a sub-group at all) so the other group members would not be displayed.

~This PR addresses that by using the top-level (shallowest) index (rather than the deepest one) to generate the current set of relevant questions.~

Actually, re-wrote to use currentIndex from the FormSession rather than hacky string chopping

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -202,7 +202,8 @@ public class FormController extends AbstractBaseController{
                             formEntryModel,
                             answer,
                             key,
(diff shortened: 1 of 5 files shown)


biyeun:
@wpride THANK YOU !!!! :fire:

PR #312, opened 2017-03-08
For binary questions return answer text

wpride:
Fixes https://manage.dimagi.com/default.asp?249073#1299220

Before, for binary type questions, Formplayer would return null as the answer. However, this makes editing forms with required binary-type questions impossible as they will be treated as unanswered, even when they have been. This returns the string path of the binary file (IE "1488964254571.jpg")

@ctsims @benrudolph curious if you know of any possible side effects of this


ctsims:
Interesting.

In the XForms spec when you submit a multipart MIME envelope you give the "attachment" part an ID. We match the internal string of the binary questions to match that ID, which happens to also match the filename. I think this makes sense as a similar strategy. Are you also using the blob filename as the multipart ID?

wpride:
@ctsims Currently this only is relevant for editing forms - specifically, preserving the ID as was originally entered. We don't have any plans for supporting multimedia entry at the moment. As I think about this again, I think this only changes the front end behavior (the ID was probably already present from Formplayer's perspective) so I think this is safe to merge.

PR #315, opened 2017-03-10
actually save restore

benrudolph:
Taking a look at refactoring restores to leverage streams, but first doing some smaller stuff. I don't think formplayer is actually saving restores when we open a form. At least that what I've gathered looking locally and inspecting the code. @wpride 

Going to write a test for this

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -248,7 +248,7 @@ public abstract class AbstractBaseController {
     }
 
     private NewFormResponse generateFormEntryScreen(MenuSession menuSession) throws Exception {
-        FormSession formEntrySession = menuSession.getFormEntrySession();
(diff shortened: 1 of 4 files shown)


benrudolph:
Actually been thinking about this more. I don't think we should actually save the restore in the form session. we currently aren't and i think that makes the most sense. else we'll have to do a restore every time someone opens a form. presumably we only need that restore to do better incomplete forms and i don't that is worth it. if we can drop this column, then we would only need to load the restore into memory every time we do a real restore and that would be trivially to change to a stream

wpride:
@benrudolph I'm fine. I think I added restoreXml originally to better support incomplete, but agree that the complication seems not worth it.

benrudolph:
that's awesome, will clean this up and then change us to using streams! 💧 💧 💧 💧

PR #317, opened 2017-03-10
Stream Response

benrudolph:
@wpride @snopoke this is some work towards only dealing with restores via streams. not completely finished but wanted to get opinions on the approach. so far this correctly saves all the restore xml into postgres and you can open a form. i think there's an error occurring when answering a question where it tries to read the full stream twice. i don't think loading the entire restore into memory is tenable for enikshay if we plan on having 100s of concurrent users

### DIFF
--- a/src/main/java/hq/CaseAPIs.java
+++ b/src/main/java/hq/CaseAPIs.java
@@ -21,6 +21,7 @@ import services.RestoreFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 
 /**
  * Created by willpride on 1/7/16.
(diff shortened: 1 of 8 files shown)


src/main/java/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -65,23 +65,85 @@
...
+        public void setValues(PreparedStatement ps) throws SQLException {
+            byte[] sessionDataBytes = writeToBytes(mSession.getSessionData());
+
+            ps.setString(1, mSession.getInstanceXml());
+            ps.setString(2, mSession.getId());
+            ps.setString(3, mSession.getFormXml());
+            try {
+                ps.setCharacterStream(4, new InputStreamReader(mSession.getRestoreXml(), "UTF-8"), mSession.getRestoreXml().available());
+            } catch (UnsupportedEncodingException e) {
+                e.printStackTrace();
```
benrudolph: will fix all these in another pass

src/main/java/hq/CaseAPIs.java
```diff
@@ -41,12 +42,12 @@ public static UserSqlSandbox restoreIfNotExists(RestoreFactory restoreFactory, b
             return restoreFactory.getSqlSandbox();
         } else{
             db.getParentFile().mkdirs();
-            String xml = restoreFactory.getRestoreXml(overwriteCache);
-            return restoreUser(restoreFactory.getWrappedUsername(), restoreFactory.getDbPath(), xml);
+            InputStream restoreStream = restoreFactory.getRestoreXml(overwriteCache);
```
benrudolph: 😂

benrudolph:
going to also have to figure out a strategy for when we want to get a new stream handler, i'm thinking if the stream is exhausted we fetch a new stream from the one saved in postgres

src/main/java/services/RestoreFactory.java
```diff
@@ -64,7 +67,7 @@
 
     private final Log log = LogFactory.getLog(RestoreFactory.class);
 
-    private String cachedRestore = null;
+    private InputStream cachedRestore = null;
```
benrudolph: we probably won't need this anymore

snopoke:
Don't have that much context on this. What's the reason for the change?

benrudolph:
We're currently thinking about improving restores on web apps for many users, one low hanging fruit i thought was changing the code to use streams instead of loading the entire payload into memory. their smallest load test user has a 17mb restore payload. i had actually imagined that their restore payload was larger, so maybe this change isn't worth it as much

PR #318, opened 2017-03-13
Stream response 2

benrudolph:
Ok @wpride I'm feeling better about this. Can be read commit by commit. This completely removes restoreXml and removes some complexity. The conversion to using streams is in the second commit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/
+testdbs/*
 .gradle
 config/application.properties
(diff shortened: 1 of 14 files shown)


wpride:
Conflicts, but loving this

benrudolph:
sweet sounds good, heh the conflict is because the submodule branch has diverged 😖

benrudolph:
O_o ✅

PR #321, opened 2017-03-15
persist function context, parse datetime correctly

wpride:
Fixes https://manage.dimagi.com/default.asp?249311

Two main issues here:

the function handlers were not being serialized along with the formDef so we need to persist them separately
The now() function wasn't being parsed properly because we were parsing a Date instead of a Datetime

This addresses both issues


benrudolph:
weird we need to clean travis, shouldn't have any prior build artifacts

wpride:
@benrudolph yeah I'm not clear what's going on here... looks like the test resource isn't being updated correctly but cleaning didn't fix it

wpride:
@benrudolph time zones are the worst

PR #322, opened 2017-03-15
Bulk Restore

wpride:
This adds two key optimizations, a Formplayer implementation of @ctsims bulk restoring code and an optimization that allows us to skip an extra update on SQLite writes by loading up the ID during read time (rather than after writing). Matrix showing performance of restoring Enikshay's loadtest4 user (~8000 cases):

Bylk YES ID Optimization YES 5.838 seconds
Bulk NO, ID Optimization YES  32.651 seconds
Bulk YES, ID Optimization NO 68.413 seconds
Bulk NO ID Optimization NO 70.471 seconds

The bulk reading is set via an application property with key `cc-enable-bulk-performance` so this will need to be added as a custom application property on enikshay. Clayton notes that this will likely break ledger processing at the moment, hence the need for the flag.

The ID optimization will be released for all so we should see significant gains from that immediately.

@snopoke Addressing your Slack comment I'm not too worried about merging a PR of this size before going offline since most of the changed code will be under said property flag.

cc @benrudolph


wpride:
@ctsims Assuming join() doesn't guarantee consistent order? Getting failures:

tests.CaseDbModelQueryTests > testDbModelQueryLookups FAILED
    junit.framework.ComparisonFailure: XPath: join(',',instance('casedb')/casedb/case[@case_type='unit_test_child_child'][@status='open'][true() and instance('casedb')/casedb/case[@case_id = instance('casedb')/casedb/case[@case_id=current()/index/parent]/index/parent]/test = 'true']/@case_id) expected:<child_[ptwo_one_one,child]_one_one> but was:<child_[one_one,child_ptwo]_one_one>


tests.CaseDbOptimizationsTest > testDbOptimizations FAILED
    junit.framework.ComparisonFailure: XPath: join(',',instance('casedb')/casedb/case[index/parent = 'test_case_parent']/@case_id) expected:<child_one,child_t[wo,child_three]> but was:<child_one,child_t[hree,child_two]>

(dunno why its doing that weird [] things but ignore that)

ctsims:
@wspride Weird, join actually _Should_ have a consistent order. Is this post-optimization implementation on your side? Did you use sets or lists for the index optimizations? If you used sets the set ordering wouldn't be consistent which might be resulting in what you are seeing.

src/main/java/sandbox/SqliteIndexedStorageUtility.java
```diff
@@ -193,8 +192,8 @@ public T getRecordForValue(String fieldName, Object value)
             if (!resultSet.next()) {
                 throw new NoSuchElementException();
             }
-            byte[] mBytes = resultSet.getBytes(org.commcare.modern.database.DatabaseHelper.DATA_COL);
-            return readFromBytes(mBytes);
+            byte[] mBytes = resultSet.getBytes(DatabaseHelper.DATA_COL);
```
benrudolph: i don't have a sense at when in the process this gets called, but if `mBytes` can be big, may want to use a stream

ctsims: That's a good catch in general.

For our purposes we treat these records as always being quite small (I think < 1mb) intentionally so I think we can get away without streaming these as-is.

benrudolph: makes sense thanks

src/main/java/sandbox/SqliteIndexedStorageUtility.java
```diff
@@ -105,7 +105,7 @@ public void write(Persistable p) {
         connection = getConnection();
         int id = SqlHelper.insertToTable(connection, tableName, p);
         p.setID(id);
-        SqlHelper.updateId(connection, tableName, p);
+        //SqlHelper.updateId(connection, tableName, p);
```
benrudolph: comment

src/main/java/sandbox/SqliteIndexedStorageUtility.java
```diff
@@ -516,4 +514,33 @@ public void bulkRead(LinkedHashSet<Integer> body, HashMap<Integer, T> recordMap)
...
+     * Retrieves a set of the models in storage based on a list of values matching one if the
+     * indexes of this storage
+     */
+    public List<T> getBulkRecordsForIndex(String indexName, Collection<String> matchingValues) {
+        List<T> returnSet = new ArrayList<>();
+        String fieldName = TableBuilder.scrubName(indexName);
+        List<Pair<String, String[]>> whereParamList = TableBuilder.sqlList(matchingValues, "?");
+        try {
+            for (Pair<String, String[]> querySet : whereParamList) {
+                //Cursor c = helper.getHandle().query(tableName, new String[]{DatabaseHelper.ID_COL, DatabaseHelper.DATA_COL, fieldName}, fieldName + " IN " + querySet.first, querySet.second, null, null, null);
```
benrudolph: comment

src/main/java/sandbox/SqliteIndexedStorageUtility.java
```diff
@@ -516,4 +514,33 @@ public void bulkRead(LinkedHashSet<Integer> body, HashMap<Integer, T> recordMap)
...
+                ResultSet resultSet = selectStatement.executeQuery();
+                while (resultSet.next()) {
+                    int index = resultSet.getInt(DatabaseHelper.DATA_COL);
+                    byte[] data = resultSet.getBytes(index);
+                    returnSet.add(newObject(data, resultSet.getInt(DatabaseHelper.ID_COL)));
+                }
+            }
+            return returnSet;
+        } catch (SQLException e) {
+
```
benrudolph: error handling

wpride:
@snopoke Could you take a look at this PR if you have a chance before EOD? Ben is offline and trying to push in before I'm off

src/main/java/database/models/EntityStorageCache.java
```diff
@@ -137,6 +140,19 @@ public void invalidateCache(String recordId) {
...
 
+    /**
+     * Removes cache records associated with the provided IDs
+     */
+    public void invalidateCaches(Collection<Integer> recordIds) {
+        List<Pair<String, String[]>> whereParamList = TableBuilder.sqlList(recordIds);
+        for(Pair<String, String[]> querySet : whereParamList) {
+            SqlHelper.deleteFromTableWhere(handler.getConnection(),
+                    TABLE_NAME,
+                    COL_CACHE_NAME + " = '" + this.mCacheName + "' AND " + COL_ENTITY_KEY + " IN " + querySet.first,
```
snopoke: Seems like a good candidate for `MessageFormat`

src/main/java/database/models/EntityStorageCache.java
```diff
@@ -137,6 +140,19 @@ public void invalidateCache(String recordId) {
         SqlHelper.deleteFromTableWhere(handler.getConnection(), TABLE_NAME, wherePair.first, wherePair.second);
     }
 
+    /**
+     * Removes cache records associated with the provided IDs
+     */
+    public void invalidateCaches(Collection<Integer> recordIds) {
```
snopoke: Other methods short circuit if `recordIds.size() == 0`. Would that be useful here?

src/main/java/engine/FormplayerTransactionParserFactory.java
```diff
@@ -45,11 +46,20 @@ public TransactionParser getParser(KXmlParser parser) {
 
     @Override
     public void initCaseParser() {
-        caseParser = new TransactionParserFactory() {
-            CaseXmlParser created = null;
+        if (true) {
```
snopoke: seems pointless

src/main/java/parsers/FormplayerBulkCaseXmlParser.java
```diff
@@ -0,0 +1,62 @@
...
+import org.xmlpull.v1.XmlPullParserException;
+import sandbox.SqliteIndexedStorageUtility;
+import sandbox.UserSqlSandbox;
+
+import java.io.IOException;
+import java.util.*;
+
+/**
+ * A bulk processing parser for the android platform. Provides superior performance when
+ * processing high case loads during syncing and/or processing.
```
snopoke: Comment needs updating

src/main/java/parsers/FormplayerBulkCaseXmlParser.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    @Override
+    protected void performBulkWrite(LinkedHashMap<String, Case> writeLog) throws IOException {
+        ArrayList<Integer> recordIdsToWipe = new ArrayList<>();
+        for (String cid : writeLog.keySet()) {
+            Case c = writeLog.get(cid);
+            storage.write(c);
+            recordIdsToWipe.add(c.getID());
```
snopoke: presumably `c.getID() == cid`?

wpride: Actually (confusingly) no. `c.getID()` returns the record's SQL ID while cId is the actual case ID. Renamed the var/added comment to clarify.

benrudolph:
tests are failing, would be great if we always did bulk restore eventually. don't see why it's even useful to not batch. one other note is that it'd be good to have the tests run with with the flag turned on

benrudolph:
looks good on ✅

src/main/java/sandbox/SqlSandboxUtils.java
```diff
@@ -51,7 +51,7 @@ public static SQLiteConnectionPoolDataSource getDataSource(String databaseName,
             }
             Class.forName("org.sqlite.JDBC");
             SQLiteConnectionPoolDataSource dataSource = new SQLiteConnectionPoolDataSource();
-            dataSource.setUrl("jdbc:sqlite:" + databasePath + "/" + databaseName + ".db");
+            dataSource.setUrl("jdbc:sqlite:" + databasePath + "/" + databaseName + ".db?journal_mode=MEMORY");
```
wpride: Switching our journaling mode to `memory`. Per SQLite docs:

The MEMORY journaling mode stores the rollback journal in volatile RAM. This saves disk I/O but at the expense of database safety and integrity. If the application using SQLite crashes in the middle of a transaction when the MEMORY journaling mode is set, then the database file will very likely go corrupt.

Since we're getting killed on I/O but have memory to spare and don't care about database corruption in the event of a crash, I think this is the way to go

ctsims: nice. I agree with this logic a lot. 

Since there are plenty of caching related reasons why we'd rebuild a db entirely anyway, seems totally fine to optimize for memory as long as the hard fallback is consistent

wpride:
@benrudolph I think one big reason is that we're not sure how this works with ledgers yet

src/main/java/sandbox/SqlSandboxUtils.java
benrudolph: 👍 though this seems like something that would belong in separate PR. it could potentially be confounding if something goes wrong when we merge this. we should also be more careful about handling SQL errors if this increases the number of them. maybe having a custom sql exception so we can handle and send an appropriate message to the user like "Please refresh the application" instead of  "[SQLITE_ERROR] SQL error or missing database (no such table: GLOBAL_RESOURCE_TABLE) "

PR #324, opened 2017-03-16
Revert "Revert "Restore transactions""

wpride:
Reverts dimagi/formplayer#323


wpride:
@benrudolph Apparently the @PreDestroy notation is not respected for request-scoped beans as these are outside the Spring-managed lifecycle. So, those static singleton variables could and would linger between requests, causing the concurrency issues we saw. This change nulls out of the connection whenever the storage factories are configured which is before anytime that they are used.

PR #328, opened 2017-03-19
Fix delete DBs for as user

wpride:
fixes https://manage.dimagi.com/default.asp?249750#1304393

### DIFF
--- a/src/main/java/util/ApplicationUtils.java
+++ b/src/main/java/util/ApplicationUtils.java
@@ -9,11 +9,11 @@ import org.commcare.modern.database.TableBuilder;
  */
 public class ApplicationUtils {
 
-    public static boolean deleteApplicationDbs(String domain, String username, String appId) {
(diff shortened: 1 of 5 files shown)


src/main/java/util/ApplicationUtils.java
```diff
@@ -24,4 +24,11 @@ public static boolean deleteApplicationDbs(String domain, String username, Strin
     public static String getApplicationDBPath(String domain, String username, String appId) {
         return SQLiteProperties.getDataDir() + domain + "/" + TableBuilder.scrubName(username) + "/" + appId;
     }
+
+    public static String getUsernameDetail(String username, String asUsername) {
+        if (asUsername != null) {
+            return username + "_" + asUsername;
+        }
+        return username;
```
benrudolph: i think this logic is repeated elsewhere. should just use this one

PR #329, opened 2017-03-20
Fix for indexing operation that happens for large case list loads

wpride:
Fix for these `java.sql.SQLException: [SQLITE_ERROR] SQL error or missing database (near "SELECT": syntax error)`

### DIFF
--- a/src/main/java/database/models/FormplayerCaseIndexTable.java
+++ b/src/main/java/database/models/FormplayerCaseIndexTable.java
@@ -251,17 +251,21 @@ public class FormplayerCaseIndexTable implements CaseIndexTable {
                                 "AND " +
                                 "%s IN %s",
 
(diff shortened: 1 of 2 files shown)


src/main/java/database/models/FormplayerCaseIndexTable.java
```diff
@@ -251,17 +251,21 @@ public DualTableSingleMatchModelQuerySet bulkReadIndexToCaseIdMatch(String index
                                 "AND " +
                                 "%s IN %s",
 
-                        COL_CASE_RECORD_ID, Case.STORAGE_KEY + "." + DatabaseHelper.ID_COL,
+                        COL_CASE_RECORD_ID, UserSqlSandbox.FORMPLAYER_CASE + "." + DatabaseHelper.ID_COL,
                         TABLE_NAME,
-                        Case.STORAGE_KEY,
+                        "CCCase",
```
benrudolph: this should use `UserSqlSandbox.FORMPLAYER_CASE`?

PR #331, opened 2017-03-23
Revert "Fix for indexing operation that happens for large case list loads"

benrudolph:
Reverts dimagi/formplayer#329

PR #333, opened 2017-03-23
Revert "Revert "Fix for indexing operation that happens for large case list loads""

benrudolph:
Reverts dimagi/formplayer#331 @wpride


wpride:
@benrudolph I've refactored the code so that the Connections in RestoreFactory and FormplayerStorageFactory are class members rather than static ThreadLocals - that's my best guess for how people were seeing each other's applications. Not sure how to convince myself of that, short of spending ~a day writing some concurrency testing framework (though maybe that'd be worth the investment). Any thoughts?

benrudolph:
hmm tough situation. i wish we had toggles in formplayer. do you still have your script that just hammers web apps?

PR #334, opened 2017-03-31
use correct helper function

snopoke:
https://manage.dimagi.com/default.asp?250785

Depends on https://github.com/dimagi/commcare-core/pull/543

### DIFF
--- a/src/main/java/database/models/FormplayerCaseIndexTable.java
+++ b/src/main/java/database/models/FormplayerCaseIndexTable.java
@@ -327,7 +327,7 @@ public class FormplayerCaseIndexTable implements org.commcare.modern.engine.case
                         COL_INDEX_NAME, indexName,
                         COL_CASE_RECORD_ID, querySet.first);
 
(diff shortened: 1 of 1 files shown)


snopoke:
Note: this code isn't tested - wasn't clear to me how to go about that.

snopoke:
Also note that I wasn't sure how to update the version of `commcare-core`. Some docs on how the dev process works for formplayer would be useful.

benrudolph:
thanks simon, this is something i've been trying to work on. we do have a process, but right now isn't written up. i'll add that to the docs. there should be a branch on `commcare-core` called `nimbus` that acts as the `master` branch. all major commcare version upgrades and nimbus only changes get PR'd into that branch

this bug i believe is actually already fixed in here: https://github.com/dimagi/formplayer/pull/333 (but was causing other issues so needed i needed to revert it)

snopoke:
Closed in favour of https://github.com/dimagi/formplayer/pull/333

PR #335, opened 2017-03-31
Contributing process

benrudolph:
@wpride @snopoke some updates to the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,12 @@ FormPlayer
 FormPlayer is a RESTful XForm entry service written on the [Spring Framework](https://projects.spring.io/spring-framework/).
 Given a [user restore](https://confluence.dimagi.com/display/commcarepublic/OTA+Restore+API) and 
 an [XForm](http://dimagi.github.io/xform-spec/) FormPlayer enables form entry via JSON calls and responses (detailed below).
-These files will often be hosted by a [CommCareHQ](https://www.github.com/dimagi/commcare-hq) server instance. FormPlayer
-relies on the [CommCare](https://www.github.com/dimagi/commcare-hq) and [Javarosa](https://www.github.com/dimagi/commcare-hq) 
-libraries (included as subrepositories). FormPlayer is built via gradle (wrapper files included). 
+These files will often be hosted by a [CommCareHQ](https://www.github.com/dimagi/commcare-hq) server instance. Formplayer relies on the [CommCare](https://www.github.com/dimagi/commcare-core) libraries (included as subrepositories). Formplayer is built via gradle (wrapper files included).
 
 ### Dependencies
-+ FormPlayer caches session instances via Redis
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,58 +45,27 @@ When building on Linux it is sometimes necessary to run:
...
+### Contributing
+
+For PRs that just modify code in the Formplayer repo, submit a PR to Formplayer on a separate branch.
+
+
+#### Contributing changes to commcare
+
+Formplayer also has a dependency on the commcare-core repository. The commcare-core `master` branch is not
+stable and Formplayer uses a different branch. The submodule repo `libs/commcare` should always be pointing to
+the `formplayer` branch.
```
benrudolph: note @wpride i've changed this to `formplayer`. I think `nimbus` doesn't make sense to other devs and we should try and consolidate names

snopoke: I actually looked for a formplayer branch

README.md
```diff
@@ -47,58 +45,27 @@ When building on Linux it is sometimes necessary to run:
...
+
+#### Updating the CommCare version
+
+When updating the commcare version on formplayer. A PR should be opened from the `commcare_X.Y` branch into
+the `formplayer` branch. Once QA has been finished, merge the PR and update the Formplayer submodule.
```
snopoke: This could be bit clearer to indicate that you're updating to a new release version of CommCare

README.md
snopoke: How do changes in the `formplayer` branch get backported to master? Or do they at all? Seems like we could end up diverging if they don't ever get back into master.

benrudolph: that's a good question. i don't know yet. so far we don't really have a process for that besides opening a PR on both branches. any ideas?

README.md
benrudolph: 👍 will update

README.md
snopoke: No, but that seems like the right approach. @ctsims might have an idea?

ctsims: So my long-term hope for formplayer would be for us to get aligned on the normal CommCare release process. Ideally I'd say we'd want to merge formplayer's stable branch into master right before we cut each release for QA at a minimum.  That seems like a good way to ensure that we're getting regular coverage of the previous releases back in each cycle.

benrudolph: i do agree it'd be nice to keep formplayer on the commcare release schedule. maybe this will be easier once it's stabilized a bit. it's a little harder since web really doesn't have any restrictions when releasing things, so whenever we get a ticket, we can either branch off from commcare release cycle and fix it right away or wait a month for the next mobile release. i think we'd just have to set those expectations if that's the route we want to go long term

snopoke:
It seems fine to maintain a mid-release branch as long as we have a process
for getting back onto the released version regularly at the end of each
release cycle.

PR #342, opened 2017-04-05
Wsp/concurrency test script

wpride:
@benrudolph Program I used to hammer Formplayer with concurrent requests. Basically asserts that the response remains unchanged for the same request, which should always be true and would (hopefully) expose any concurrency issues when the wrong response came through. Using this I was able to reproduce one SQLite read issue on the old branch and haven't been able to reproduce any on the one with the new fix.

Can't say that I'm totally convinced, but I can't think of a better methodology for pushing FP.

Also motivated the fixes here: https://github.com/dimagi/commcare-core/compare/914ab903030fabb98875b6b88055127e3486e0de...dc192803566477494591302e053115aacf520579


src/test/java/utils/Requester.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static void main(String[] args) throws Exception {
+        loadTestArgs();
+        startTests();
+    }
+
+    private static void loadTestArgs() {
+        testArgs = new Vector<>();
+        for (String user: testUsers) {
+            testArgs.add(new String[] {"265ab21237afb735076e07efba8368a8", user, "test", "0"});
```
benrudolph: would be great to be able to customize these at some point

benrudolph:
maybe a small readme in that repo to explain how to run it etc

wpride:
@benrudolph Agreed those would be nice changes. I mostly PRed this to show you what I was validating with, not intending to merge (though on second thought its fairly close so probably worth doing)

benrudolph:
ah ok understood

wpride:
@benrudolph Still a bit rough, but pulled out all of the configuration stuff into a JSON file. Think OK to merge as a tool for now.

src/test/java/concurrency_tests/SessionNavigatorThread.java
```diff
@@ -0,0 +1,122 @@
...
+        return makeRequest(sessionNavigationBean, url);
+    }
+
+    static String makeRequest(Object bean, String url) throws IOException {
+        HttpClient httpclient = HttpClients.createDefault();
+        HttpPost httppost = new HttpPost("http://localhost:8090/" + url + "/");
+
+        HttpEntity e = new StringEntity(new ObjectMapper().writeValueAsString(bean));
+        httppost.setEntity(e);
+        httppost.setHeader("Cookie", "sessionid="+ "exrjtza7k4q8xpq9xud9ogkg4x96i8df");
```
benrudolph: this unfortunately won't work

src/test/java/concurrency_tests/SessionNavigatorThread.java
```diff
@@ -0,0 +1,122 @@
...
+        sessionNavigationBean.setSelections(selections);
+        sessionNavigationBean.setAppId(appId);
+        sessionNavigationBean.setDomain(domain);
+        sessionNavigationBean.setUsername(username);
+        return makeRequest(sessionNavigationBean, url);
+    }
+
+    static String makeRequest(Object bean, String url) throws IOException {
+        HttpClient httpclient = HttpClients.createDefault();
+        HttpPost httppost = new HttpPost("http://localhost:8090/" + url + "/");
```
benrudolph: use the application.config for this?

PR #343, opened 2017-04-05
Gradle 3.4.1

wpride:
Giving some serious performance increases 💸


gradlew
```diff
@@ -42,11 +42,6 @@ case "`uname`" in
     ;;
 esac
 
-# For Cygwin, ensure paths are in UNIX format before anything is touched.
```
benrudolph: haha

PR #344, opened 2017-04-05
Do not process all entities for a single entity

benrudolph:
@wpride @ctsims quick POC on the case detail optimization. this loads enikshay case details in less than a second. hold off on merge right now, i want to add some comments and do a little organization

### DIFF
--- a/src/main/java/application/MenuController.java
+++ b/src/main/java/application/MenuController.java
@@ -139,9 +139,11 @@ public class MenuController extends AbstractBaseController{
             throw new RuntimeException("Could not find case with ID " + detailSelection);
         }
 
-        EntityDetailListResponse response = new EntityDetailListResponse(entityScreen,
+        EntityDetailListResponse response = new EntityDetailListResponse(
+                entityScreen,
                 menuSession.getSessionWrapper().getEvaluationContext(),
-                reference);
+                reference
+        );
         return response;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/beans/menus/EntityListResponse.java
```diff
@@ -47,17 +47,23 @@
...
 
         EvaluationContext ec = nextScreen.getEvalContext();
+        EntityDatum datum = (EntityDatum) session.getNeededDatum();
+        if (detailSelection != null) {
+            TreeReference reference = datum.getEntityFromID(ec, detailSelection);
+            entities = new EntityBean[]{processEntity(reference, nextScreen, ec)};
+        } else {
+            Vector<TreeReference> references = ec.expandReference(datum.getNodeset());
+            processEntities(nextScreen, references, ec, offset, searchText);
+        }
```
benrudolph: this is the meat of the optimization, the rest is just parameter passing

src/main/java/beans/menus/EntityListResponse.java
```diff
@@ -47,17 +47,23 @@
...
-    public EntityListResponse(EntityScreen nextScreen, int offset, String searchText, String id) {
+    public EntityListResponse(EntityScreen nextScreen, String detailSelection, int offset, String searchText, String id) {
         SessionWrapper session = nextScreen.getSession();
         Detail shortDetail = nextScreen.getShortDetail();
         nextScreen.getLongDetailList();
 
         EvaluationContext ec = nextScreen.getEvalContext();
+        EntityDatum datum = (EntityDatum) session.getNeededDatum();
+        if (detailSelection != null) {
+            TreeReference reference = datum.getEntityFromID(ec, detailSelection);
```
benrudolph: going to add error handling here if this is null

src/main/java/application/MenuController.java
```diff
@@ -137,9 +139,11 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
             throw new RuntimeException("Could not find case with ID " + detailSelection);
         }
 
-        EntityDetailListResponse response = new EntityDetailListResponse(entityScreen,
+        EntityDetailListResponse response = new EntityDetailListResponse(
```
benrudolph: still feels like we're doing a little extra work than we need to in this route, but don't think it's anything significant

PR #345, opened 2017-04-11
Fix OQPS and Repeat methods

wpride:


wpride:
Hotfixing since these methods are all broken atm

PR #347, opened 2017-04-11
Fix case claim

benrudolph:
@wpride @ctsims needs this guy https://github.com/dimagi/commcare-core/pull/547. default scoping of variables is package level which is what caused the null pointer exception


benrudolph:
@wpride it seems like the `formplayer` branch on commcare-core has diverged. can you please fix?

benrudolph:
ok r2g

PR #349, opened 2017-04-12
Let UnresolvedResourceException through, don't email

wpride:
@benrudolph This should finally stop the annoying UnresolvedResourceException (app config) emails. Just let the typed exception through instead of re-throwing as a runtime


benrudolph:
nice looks good

PR #350, opened 2017-04-12
Check if form index is first in the form

wpride:
We were running into a situation where anytime the FormIndex for the first displayed question wasn't `0` - for example when the first question is in a group it will have index `0,0` - then we would not grey out the getPrevious arrow if we returned to the question. This rectifies that by checking on Formplayer whether there is truly another question before the current one. 

cross-request: https://github.com/dimagi/commcare-hq/pull/15738

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -264,9 +264,8 @@ public class FormController extends AbstractBaseController{
         restoreFactory.configure(requestBean, new DjangoAuth(authToken));
         FormSession formSession = new FormSession(serializableFormSession, restoreFactory);
         formSession.stepToNextIndex();
-        JSONObject resp = formSession.getNextJson();
+        FormEntryNavigationResponseBean responseBean = formSession.getFormNavigation();
         updateSession(formSession, serializableFormSession);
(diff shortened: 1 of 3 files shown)


src/main/java/application/FormController.java
```diff
@@ -280,12 +279,8 @@ public FormEntryNavigationResponseBean getPrevious(@RequestBody SessionRequestBe
         restoreFactory.configure(requestBean, new DjangoAuth(authToken));
         FormSession formSession = new FormSession(serializableFormSession, restoreFactory);
         formSession.stepToPreviousIndex();
-        JSONObject resp = JsonActionUtils.getCurrentJson(formSession.getFormEntryController(),
-                formSession.getFormEntryModel(),
-                formSession.getCurrentIndex());
+        FormEntryNavigationResponseBean responseBean = formSession.getFormNavigation();
         updateSession(formSession, serializableFormSession);
-        FormEntryNavigationResponseBean responseBean = mapper.readValue(resp.toString(), FormEntryNavigationResponseBean.class);
-        responseBean.setCurrentIndex(formSession.getCurrentIndex());
```
benrudolph: nice cleanup

PR #353, opened 2017-04-14
Use UserRestore annotation for configuring RestoreFactory

wpride:


src/main/java/aspects/UserRestoreAspect.java
```diff
@@ -0,0 +1,63 @@
...
+        if (!(args.length > 1 && args[0] instanceof AuthenticatedRequestBean && args[1] instanceof String)) {
+            throw new RuntimeException("Could not configure RestoreFactory with args " + Arrays.toString(args));
+        }
+
+        AuthenticatedRequestBean requestBean = (AuthenticatedRequestBean) args[0];
+        HqAuth auth = getAuthHeaders(requestBean.getDomain(), requestBean.getUsername(), (String) args[1]);
+        restoreFactory.configure((AuthenticatedRequestBean)args[0], auth);
+    }
+
+    private HqAuth getAuthHeaders(String domain, String username, String sessionToken) {
```
benrudolph: can delete the other implementation of this

wpride: Still used in a few places, unfortunately. Auth should also be moved into a Service

PR #356, opened 2017-04-14
AuthService and aspect

wpride:


src/main/java/annotations/RequiresAuth.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for methods that make authenticated requests to HQ
+ */
+@Component
+@Target(value = {ElementType.METHOD, ElementType.TYPE})
+@Retention(value = RetentionPolicy.RUNTIME)
+public @interface RequiresAuth {
```
benrudolph: don't think this is actually used

src/main/java/aspects/AuthAspect.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Aspect to configure the AuthService
+ */
+@Aspect
+public class AuthAspect {
+
+    @Autowired
+    protected AuthService authService;
+
+    @Before(value = "@annotation(org.springframework.web.bind.annotation.RequestMapping)")
```
benrudolph: clever, but do you think this is too broad? will this require auth for form translate or swagger?

wpride: Currently it will attempt to assign authentication for every request, but won't fail if its not an AuthenticatedRequest. It would fail later when the expected `Auth` is null

PR #365, opened 2017-04-19
Move JSON Utils to Formplayer

wpride:
These were only being used by Formplayer and will be for the foreseeable future. Became annoying to keep these in sync.


benrudolph:
looks good once tests are passing
https://github.com/dimagi/commcare-core/pull/552

PR #366, opened 2017-04-20
Profiling

wpride:


ctsims:
Hey Will!

Looks like formplayer literally wasn't using case indexes at all.

Got the query time down to 23-26 seconds (from 334 on my current machine), but the way the indices are injected is a bit gross.

I think we should get the indexing into formplayer _immediately_, though. I expect that it will improve runtimes by an insane amount.

ctsims:
@wpride Quick update: double checked w/profiler, bulkReadIndexToCaseIdMatch(0 is now only 418ms of the remaining 26 seconds, so I think we can close the book on this part.

All remaining runtime appears to be parseIntoSandbox (14s) and installing the app.

ctsims:
Revised: was actually about 10s worth of case list loading stuff. Added another quick optimization to improve bootstrapping the case list.

Before:
Setting locale to: en
Case iterate took: 4645ms
Elasped 31.525

After:
Setting locale to: en
Case iterate took: 92ms
Elasped 22.256

src/main/java/sandbox/JdbcSqlIndexIterator.java
```diff
@@ -0,0 +1,82 @@
...
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Iterator;
+
+/**
+ * Iterator that only processes id indexes, and not real data.
+ * 
+ * @author ctsims
+ */
+public class JdbcSqlIndexIterator<E extends Persistable> extends JdbcSqlStorageIterator<E> {
```
ctsims: Instead of extending this, these two should be either merged, or use a common ancestor, this was just the fastest way to bootstrap things

src/main/java/sandbox/SqlHelper.java
```diff
@@ -54,6 +110,17 @@ public static void createTable(Connection c, String storageKey, Persistable p) {
         try {
             preparedStatement = c.prepareStatement(sqlStatement);
             preparedStatement.execute();
+            if (storageKey.equals("CCCase")) {
```
ctsims: This is the indexing section, it's where most of the gains come from

PR #368, opened 2017-04-21
Fix Case Claim for Login As Users

wpride:
@proteusvacuum @ctsims @benrudolph 

fixes https://manage.dimagi.com/default.asp?252448#1321136

### DIFF
--- a/src/main/java/application/MenuController.java
+++ b/src/main/java/application/MenuController.java
@@ -15,6 +15,7 @@ import beans.menus.EntityDetailListResponse;
 import beans.menus.UpdateRequestBean;
 import exceptions.FormNotFoundException;
 import exceptions.MenuNotFoundException;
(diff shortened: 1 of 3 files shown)


src/main/java/application/MenuController.java
```diff
@@ -374,9 +375,11 @@ private NotificationMessageBean doSync(FormplayerSyncScreen screen, HqAuth auth)
             return new NotificationMessageBean("Session error, expected sync block but didn't get one.", true);
         }
         if(responseEntity.getStatusCode().is2xxSuccessful()){
-            return new NotificationMessageBean("Case claim successful", false);
+            CaseAPIs.forceRestore(restoreFactory);
```
proteusvacuum: @wspride just checking that this is necessary? I thought that webapps already did this when a claim is successful.

PR #369, opened 2017-04-25
Case claim redirect, hide buttons when appropraite

wpride:
cross-request: https://github.com/dimagi/commcare-hq/pull/15877

fixes: https://manage.dimagi.com/default.asp?252280


src/main/java/beans/menus/EntityListResponse.java
```diff
@@ -230,7 +230,13 @@ private void processStyles(Detail detail) {
     }
 
     private void processActions(SessionWrapper session) {
-        Vector<Action> actions = session.getDetail(((EntityDatum) session.getNeededDatum()).getShortDetail()).getCustomActions(session.getEvaluationContext());
+        EntityDatum datum = (EntityDatum) session.getNeededDatum();
+        // HACKY! Uses the fact that case claim results will always be in a "result" instance
```
wpride: @ctsims I couldn't find any cleaner way to determine whether we're on an EntityList as a result of a case claim action or a real case list than looking at the instance name. Any brighter ideas appreciated

benrudolph: would be great to have a test for this with an explanation

proteusvacuum: I think this is actually defined in the `storage-instance` of the query node, so if this is the route you take, this should actually take the name that is defined there. 
https://github.com/dimagi/commcare-core/wiki/Suite20#remote-request

wpride: @proteusvacuum Sadly Mr. Vacuum that SessionDatum isn't available anymore - we've already proceeded on to the next step at this point, so we only know that we're on an entity list and not about the genesis of the nodeset that we're displaying. You did help me remember, however, that we *do* have access to the frame stack and specifically the previous step we took and so can check for the type of that step

proteusvacuum: yay

PR #374, opened 2017-04-25
Allow Formplayer to reuse application dbs

wpride:
Allow Formplayer to reuse application installs between runs. Accomplish this by using Redis to store mappings between the file's GUID and the ZipFile location, rather than in a static map.

cross-request: https://github.com/dimagi/commcare-core/pull/556

@benrudolph


src/main/java/engine/FormplayerConfigEngine.java
```diff
@@ -77,7 +92,8 @@ private void handleInstallError(InputStream errorStream) {
 
     @Override
     protected void setRoots() {
-        super.setRoots();
```
benrudolph: hooray for never ever using static memory to hold persistent data

src/main/java/engine/FormplayerArchiveFileRoot.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Reference derive(String guidPath) throws InvalidReferenceException {
+        try {
+            listOperations.trim(String.format("formplayer:archive:%s", getGUID(guidPath)), 0, MAX_RECENT);
+            String zipName = listOperations.range(String.format("formplayer:archive:%s", getGUID(guidPath)), 0, MAX_RECENT).get(0);
+            return new ArchiveFileReference(new ZipFile(zipName), getGUID(guidPath), getPath(guidPath));
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
benrudolph: what does this look like when it fails to find the zip path in redis? it's very common for redis to just discard keys. we shouldn't count on it to be a permanent store

benrudolph:
is there a corresponding `commcare-core` PR? (i don't subscribe to that repo's PRs)

src/main/java/engine/FormplayerArchiveFileRoot.java
wpride: If this happens on re-initialization, then we'll catch the error here, then wipe the DBs, and reinstall:

https://github.com/dimagi/formplayer/blob/68f564ed704cfd63dcad1cb7ec0c30717ef0513e/src/main/java/services/impl/InstallServiceImpl.java#L49-L49

Are you saying this key could also be cleared out during install (IE within ~60 seconds of its being initialized?) If so then we might want to use a different store

benrudolph: cool, no i don't think it'll be wiped in 60 seconds

wpride:
@benrudolph OK updated this to use the static mapping first, then fall back to Redis if this is missing. I think best to avoid the overhead of spinning up a bunch of ZipFile objects during install

src/main/java/engine/FormplayerArchiveFileRoot.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Created by willpride on 4/25/17.
+ */
+public class FormplayerArchiveFileRoot extends ArchiveFileRoot {
+
+    @Autowired
+    private StringRedisTemplate redisTemplate;
+
+    @Resource(name = "redisTemplate")
+    private ListOperations<String, String> listOperations;
```
benrudolph: why does this use listOperations. we shouldn't need to use any special list stuff if all we're doing is setting a key and retrieving a key

src/main/java/engine/FormplayerArchiveFileRoot.java
```diff
@@ -0,0 +1,50 @@
...
+        String mGUID = super.addArchiveFile(zip);
+        redisTemplate.opsForValue().set(
+                String.format("formplayer:archive:%s", mGUID),
+                zip.getName()
+        );
+        return mGUID;
+    }
+
+    @Override
+    public Reference derive(String guidPath) throws InvalidReferenceException {
```
benrudolph: nit: a comment explaining what this is doing would be useful

PR #376, opened 2017-04-27
handle more errors

benrudolph:
@wpride seeing lots of different types of errors on ICDS. this handles them more gracefully


wpride:
nice. do we want to continue getting emails from these?

benrudolph:
right now i'd like too because the 503 and 504 errors don't occur at the django level (nginx level) and we don't get any error email for them. i think we need to move to sentry soon...

benrudolph:
if it becomes too much we can catch

PR #377, opened 2017-04-27
Purge Cases after Submission

wpride:
fixes https://manage.dimagi.com/default.asp?252309

~@ctsims Not checking for the developer flag yet - any reason to not do this for every submission?~


ctsims:
@wpride do we want to add the feature flag after our conversation this morning before pulling into formplayer?

wpride:
@ctsims added @proteusvacuum note that you will need to add custom property `cc-auto-purge` with value `true`

ctsims:
I have added the custom property to the app, thanks Will

PR #379, opened 2017-04-28
Incremental Sync for Formplayer

wpride:
Implement Incremental Sync for Formplayer. Switch all syncing calls to default to incremental syncing, rather than wiping out DBs and syncing anew.

This PR makes the wipe DBs code path unused. We should implement a new path to this and an obscure but available Clear User Data UI similar to what we have in mobile for when user data gets into a bad state.

This code replicates a good deal of code in `commcare-android`. We should unify this logic in the future.


src/main/java/services/RestoreFactory.java
```diff
@@ -337,22 +329,38 @@ private InputStream getRestoreXmlHelper(String restoreUrl, HqAuth auth) {
...
 
+    public String getRestoreUrl(boolean overwriteCache) {
+        StringBuilder builder = new StringBuilder();
+        builder.append(host);
+        builder.append("/a/");
+        builder.append(domain);
+        builder.append("/phone/restore/?version=2.0");
         if (overwriteCache) {
-            url += "&overwrite_cache=true";
+            builder.append("&overwrite_cache=true");
```
proteusvacuum: this overwrites all of HQs caching of restores. Is there a reason you do this, and not let HQ handle the way it deals with its own caching? I was under the impression that this param was only called for debugging purposes. 
If you hit the restore URL without a `since` param, I think it should return what you expect (i.e. it will generate a new restore payload, or fetch the restore from the cache if it can)

wpride: @proteusvacuum This should only be true for the forceResture path (currently unused) which is meant for a hard wipe/refresh of data. Otherwise should not be using this param.

proteusvacuum: I might just be reading this wrong, but doesn't [this line](https://github.com/dimagi/formplayer/pull/379/files#diff-6b353b482e6dd0481ce5fe61fe333ac7L387) when you are doing a case claim set that param?

wpride: @proteusvacuum you're so right, thanks

proteusvacuum:
Looks good to me, but I don't think I have enough context to merge

src/main/java/services/RestoreFactory.java
benrudolph: the reason this was an issue before was that we didn't actually use sync log ids. restores would then get cached for an hour if they took too long to load. this caused users to not get back a fresh restore and we were getting many tickets about it from icds since their workflow requires that you hit sync at the end of doing some form entry

src/main/java/sandbox/UserSqlSandbox.java
```diff
@@ -192,4 +192,9 @@ public void setLoggedInUser(User user) {
     public Connection getConnection() {
         return handler.getConnection();
     }
+
+    @Override
+    public void setSyncToken(String token) {
+        super.setSyncToken(token);
+    }
```
benrudolph: can we just remove this altogether if it's not changing the functionality?

PR #380, opened 2017-04-28
Redirect properly when case claim is already owned

wpride:
Fixes https://manage.dimagi.com/default.asp?252925#1324484


snopoke:
I don't see anything changing other than renaming the class and adding a `toString`

wpride:
@snopoke looks like you have to expand the diff of `src/main/java/application/MenuController.java` - git is compressing it

src/main/java/application/MenuController.java
```diff
@@ -322,69 +318,59 @@ private MenuSession handlePreviewCommand(SessionNavigationBean sessionNavigation
...
+                queryDictionary);
+        menuSession.updateScreen();
+        nextScreen = menuSession.getNextScreen();
+        log.info("Next screen after query: " + nextScreen);
+        return notificationMessage;
     }
 
-    protected NotificationMessageBean doQuery(FormplayerQueryScreen nextScreen,
-                                              Hashtable<String, String> queryDictionary) {
+    protected NotificationMessage doQuery(FormplayerQueryScreen nextScreen,
```
snopoke: It's a bit confusing to have these both called `doQuery` when they don't do the same thing. Could we rename the first one to something like `doQueryAdvanceToNext`?

Same for `doSync`

PR #381, opened 2017-05-01
Return persistent case tile when asked for detail not on case list

wpride:
Fixes https://manage.dimagi.com/default.asp?252775

cross-request: https://github.com/dimagi/commcare-hq/pull/15959

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -191,7 +191,8 @@ public abstract class AbstractBaseController {
         }
     }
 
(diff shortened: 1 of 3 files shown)


ctsims:
@wpride Does this include a UI element to access the persistent case tile? I thought Sheel had mentioned that it may also not have actually a specific way to call up the persistent case tile (although now that I say that I don't know how he managed to test that it was breaking if he couldn't call it up)

wpride:
@ctsims the entire tile is currently clickable. I included a ping in the ticket to the UI team asking for ideas about a UI clue to add

ctsims:
Got it. Thanks!

PR #385, opened 2017-05-03
Query Encoding

wpride:
@proteusvacuum I think this might do the trick re https://manage.dimagi.com/default.asp?252491


proteusvacuum:
I'm not really sure what this will do, since we will still send the same data down from the server. Also, the content-type of this request is `text/xml`. 

Do you want me to set the charset explicitly in the response header like this? https://github.com/dimagi/commcare-hq/commit/86d18eb9d562670a5f75000ef04c1e09038b9308 
Now that I've looked at this, that's what we do for regular restores, so should probably be done for this too.

PR #386, opened 2017-05-04
Case names in Breadcrumbs

wpride:
Before, case selections would appear in the breadcrumb bar as the previous screen's name (IE the name of the EntityScreen), resulting in that name appearing twice. This changes behavior to put the selected case's name in the breadcrumb bar instead.

The method for doing this is slightly hacky, but I think worth it for the UI/usability gain. The reason this is so tricky in CommCare is because CASE_ID selections on the stack are hard/impossible to differentiate from CASE_IDs that were just popped on there automatically. This PR resolves this by determining the breadcrumbs outside the core CommCare engine. 

cc @ctsims 

screen shot 2017-05-04 at 1 27 01 pm


ctsims:
Hey Will,

this will be super helpful for eNikshay! Thanks for the proactive work on
this

Flagging farid to note that some of our case names as is may actually be
pretty long, actually, which is a consideration. Maybe we should truncate
after some number of characters?

-Clayton

proteusvacuum:
whoa! sweet!
What happens if I click on the case name? Does it take me back to the place immediately after I selected that case? Very nice.
Agree with Clayton that some cases might look pretty weird here, but I think overall this will be very helpful.

wpride:
@proteusvacuum Yeah, exactly.

@ctsims good call on the truncation - should be easy to add in the CSS

ctsims:
I love this feature request. *love* it.

I think I want to push for truncation in the CSS before it's pulled,
though. Right now the view is not great, but there are a lot of
"programatically" generated case names in our app (and others) that are
extremely long and are going to make the breadcrumb bar unusable unless
it's truncated, which is worse than where we started.

-Clayton

benrudolph:
https://github.com/dimagi/commcare-hq/pull/16061

PR #387, opened 2017-05-04
WIP; check if formplayer can talk to couch

sravfeyn:


src/main/java/application/UtilController.java
```diff
@@ -64,6 +65,12 @@ public ServerUpBean serverUp() throws Exception {
         return new ServerUpBean();
     }
 
+    @ApiOperation(value = "Check whether formplayer can connect to couch")
+    @RequestMapping(value = Constants.URL_COUCH_CONNECTION, method = RequestMethod.GET)
+    public String serverUp() throws Exception {
+        return CouchUserRepo.couchIsUp();
```
sravfeyn: How can I wrap this under some kind of HTTP reponse object?

src/main/java/repo/impl/CouchUserRepo.java
```diff
@@ -33,4 +33,8 @@ public CouchUser getUserByUsername(String username) {
         }
         return users.get(0);
     }
+
+    public String couchIsUp() {
+        // Test whether requests to couch are going
```
sravfeyn: I am trying to query `couchDbClient` for `couch_host/couch_db_name` and test whether the response JSON contains the key-val `{'db_name': <couch_db_name>}`.

Can you give an example of how to query `couchDbClient` by providing a URL? And an example of looking up properties of JSON objects in Java.

benrudolph:
agree with @snopoke and then we just need to update the health check in django to hit and parse that endpoint

wpride:
Formplayer no longer talks to couch - 🎉

PR #388, opened 2017-05-05
Case Claim quirkiness fixes

wpride:
- Make case claim redirect work when already owning case
- Notify if query returns successfully with no results
- Don't notify if query is just successful (unnecessary)

Hold off on merge - currently results in a state where you can't use back key over a claim.


benrudolph:
discussed with will and kriti and we decided that the bug that it introduces isn't as bad as the bug that it fixes

PR #389, opened 2017-05-08
Update plugin versions, update deprecated classes

wpride:
This brings us to the latest `spring-boot` version and updates deprecated classes accordingly. Also switched from setting dependency versions manually to using the Spring Boot Gradle plugin's [dependency manager](https://docs.spring.io/spring-boot/docs/current/reference/html/build-tool-plugins-gradle-plugin.html) which is the more canonical way to do this.


wpride:
@snopoke These updates broke the `FormValidatorTests` - I think because the auto-wiring of the `Application.class` dependencies got stricter resulting in some beans getting instantiated with requirements that weren't fulfilled without using the `TestContext` present. Resolved by refactoring the tests to use our standard BaseTestClass which satisfies these requirements with mocks.

PR #390, opened 2017-05-09
Require raven param

benrudolph:
@wpride this should fix up sentry for the most part. there is still one race condition when the logging of one of the breadcrumbs fails on the initial request. Wasn't able to quickly get to the bottom of it, but it works on subsequent requests oddly. For now going to leave as is because I don't think it's terribly high priority

### DIFF
--- a/src/main/java/util/SentryUtils.java
+++ b/src/main/java/util/SentryUtils.java
@@ -6,6 +6,7 @@ import com.getsentry.raven.event.Breadcrumbs;
 import com.getsentry.raven.event.EventBuilder;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 
 /**
  * Created by benrudolph on 4/27/17.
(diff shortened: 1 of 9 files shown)


src/main/java/util/SentryUtils.java
```diff
@@ -6,31 +6,46 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
 
 /**
  * Created by benrudolph on 4/27/17.
  */
 public class SentryUtils {
+    /**
+     * We require a Raven instance in these Utils to ensure that the caller has instantiated
+     * the Raven instance. Raven keeps track of the raven instance internally with
+     * Raven.getStoredInstance(). If this call returns null, the Sentry call will fail.
```
benrudolph: this more or less explains why this is necessary

wpride:
@benrudolph Looks good, guessing you need to mock the Raven in the TestContext

Quoth the raven, Nevermore

benrudolph:
damn you TestContext! thing of evil!

wpride:
Mock.
YEAH
ing.
YEAH
🐦 
YEAH
yeah
YEAH

PR #395, opened 2017-05-12
switch back to forcing restore for both case claim and manual sync

wpride:
@benrudolph think for the time being we should revert back to our previous behavior of restoring fresh 😢  seems like this isn't playing nicely with case claim


benrudolph:
hot fix

PR #397, opened 2017-05-23
Decode uri before sending

benrudolph:
@proteusvacuum @wpride spring already encodes urls so this was getting double encoded

### DIFF
--- a/src/main/java/services/impl/QueryRequesterImpl.java
+++ b/src/main/java/services/impl/QueryRequesterImpl.java
@@ -9,6 +9,9 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.client.RestTemplate;
 import services.QueryRequester;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+
 /**
  * Created by willpride on 4/26/17.
  */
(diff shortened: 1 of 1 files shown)


benrudolph:
https://manage.dimagi.com/default.asp?254058

src/main/java/services/impl/QueryRequesterImpl.java
```diff
@@ -20,10 +23,19 @@
...
-                restTemplate.exchange(uri,
-                        HttpMethod.GET,
-                        new HttpEntity<String>(headers),
-                        String.class);
+                null;
+        try {
+            response = restTemplate.exchange(
+                    // Spring framework automatically encodes urls. This ensures we don't pass in an already
+                    // encoded url.
+                    URLDecoder.decode(uri, "UTF-8"),
```
proteusvacuum: I'm not too familiar with this, but do you know where this is getting encoded in the first place? Is the `restTemplate.exchange` the thing that is re-encoding it? 
Is it possible to stop the original encoding from taking place?

wpride: looks like this is possible https://stackoverflow.com/a/28183190

wpride:
@benrudolph thanks breezus

src/main/java/services/impl/QueryRequesterImpl.java
benrudolph: definitely the better solution. sorry i did a quick search and didn't find anything fruitful and got lazy. thanks will update

src/main/java/screens/FormplayerQueryScreen.java
```diff
@@ -27,14 +29,19 @@ public FormplayerQueryScreen(HqAuth auth){
...
         URL url = getBaseUrl();
         Hashtable<String, String> params = getQueryParams();
         UriComponentsBuilder builder = UriComponentsBuilder.fromHttpUrl(url.toString());
         for(String key: params.keySet()){
             builder.queryParam(key, params.get(key));
         }
-        return builder.toUriString();
+        // In order to get a decoded URI we decode after building the URI.
+        try {
+            return URLDecoder.decode(builder.toUriString(), "UTF-8");
```
benrudolph: ok... this isn't that much better. the solution in the stack overflow doesn't work because java throws an error when you attempt to build a url that you tell it that is encoded and it realizes it isn't. it's also a hack solution because the boolean you flag that you send in to `build` says whether or not the url is encoded or not and it's definitely the case that it is not encoded. did some more searching and it's not obvious to me that this is super easy to do. will any other ideas?

wpride: Hmm no other thoughts. I'm pro just reverting back to your original fix, don't let the perfect be the enemy of the good and all

proteusvacuum: yeah, ditto.

benrudolph:
ok tests finally passed, i think the issue was most likely with jenkins. ✅

PR #398, opened 2017-05-24
Populate sub context with context variables

wpride:
Fixes https://manage.dimagi.com/default.asp?254651

Sub-contexts used by detail tabs weren't being populated with variables from the parent


src/test/java/tests/EntityDetailTabTests.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    protected String getMockRestoreFileName() {
+        return "restores/rmnch.xml";
+    }
+
+    // Regression test confirming that variables are passed to detail tab sub contexts
+    @Test
+    public void testDetailTabChildVariable() throws Exception {
+        // just need to ensure this doesn't error
```
benrudolph: still don't think these types of tests are great. if this ever fails it becomes very difficult to figure out what exactly this was testing since the restore / app is so large

PR #400, opened 2017-05-30
More sentry info

benrudolph:
@snopoke @wpride did some reading of sentry docs to get us to a better place for formplayer error handling this does a few things:

- Better breadcrumbs
- Sets the user
- Add Extras to make it easier to debug

it's still hard to get all the different pieces of data that we need for every request.


src/main/java/aspects/LoggingAspect.java
```diff
@@ -45,19 +43,21 @@ public Object logRequest(ProceedingJoinPoint joinPoint) throws Throwable {
             log.info("Request to " + requestPath + " with no request body.");
         }
 
-        Object result = joinPoint.proceed();
```
benrudolph: this was very important. essentially we'd never get to the below

src/main/java/engine/FormplayerConfigEngine.java
```diff
@@ -57,26 +36,13 @@ public FormplayerConfigEngine(IStorageIndexedFactory storageFactory,
...
-    }
-
     @Override
     protected String downloadToTemp(String resource) {
         try {
             URL url = new URL(resource);
             HttpURLConnection conn = (HttpURLConnection) url.openConnection();
             conn.setInstanceFollowRedirects(true);  //you still need to handle redirect manully.
             HttpURLConnection.setFollowRedirects(true);
-            recordDownloadBreadcrumbs(resource, conn.getResponseMessage());
```
benrudolph: was having trouble auto wiring this class for some reason. even after including it in the component scan. don't think it super important though

PR #403, opened 2017-05-31
Incremental Sync Episode II

wpride:
Re-implements Incremental Restore with a couple fixes.

I think the main one causing the issue we saw was with disagreement between username formatting when getting the syncToken. Specifically, User objects don't have a domained username while the SqlSandbox would. This would cause a mismatch resulting in the `syncToken` being empty and since we weren't overriding the cache this would result in the (stale) cached XML being returned instead of the differential. I was able to reproduce and fix this behavior locally. 

I've also added checks for empty (non-null) syncToken and returned the "Sync" button to forcing a full restore.

### DIFF
--- a/src/main/java/hq/CaseAPIs.java
+++ b/src/main/java/hq/CaseAPIs.java
@@ -77,8 +77,8 @@ public class CaseAPIs {
         // initialize our sandbox's logged in user
         for (IStorageIterator<User> iterator = sandbox.getUserStorage().iterate(); iterator.hasMore(); ) {
             User u = iterator.nextRecord();
-            String wrappedUsername = UserUtils.unwrapUsername(restoreFactory.getWrappedUsername());
-            if (wrappedUsername.equalsIgnoreCase(u.getUsername())) {
+            String unwrappedUsername = UserUtils.unwrapUsername(restoreFactory.getWrappedUsername());
+            if (unwrappedUsername.equalsIgnoreCase(u.getUsername())) {
                 // set last sync token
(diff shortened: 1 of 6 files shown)


src/main/java/services/RestoreFactory.java
```diff
@@ -338,6 +338,9 @@ private String getSyncToken(String username) {
         if (username == null) {
             return null;
         }
+
+        username = UserUtils.unwrapUsername(username);
```
benrudolph: nit: may want to rename this to `unwrappedUsername`

src/main/java/hq/CaseAPIs.java
```diff
@@ -76,7 +77,8 @@ private static UserSqlSandbox restoreUser(RestoreFactory restoreFactory, InputSt
         // initialize our sandbox's logged in user
         for (IStorageIterator<User> iterator = sandbox.getUserStorage().iterate(); iterator.hasMore(); ) {
             User u = iterator.nextRecord();
-            if (restoreFactory.getWrappedUsername().equalsIgnoreCase(u.getUsername())) {
+            String wrappedUsername = UserUtils.unwrapUsername(restoreFactory.getWrappedUsername());
```
benrudolph: should this be `unwrappedUsername`?

src/main/java/services/RestoreFactory.java
wpride: Well when its passed in its potentially *not* unwrapped so I guess its going to be inconsistent at one point in the code

PR #404, opened 2017-06-01
Version sqlite dbs

benrudolph:
@wpride this cleans up the application user db path creation quite a bit. it now all happens in one class. before we were generating the same path in multiple places, making it very hard to actually change the db name. the second commit actually add `V1` to the end of the db file. if we ever want to wipe dbs on deploy, we should just bump change that string

buddy: @mkangia


benrudolph:
✅ @wpride

PR #405, opened 2017-06-05
Serialize nodes

benrudolph:
@wpride @ctsims this serializes nodes returned to the debugger in XML. i've already found it to be very useful. it also adds autocomplete options for instance declarations if it knows them. hold off on merge as it needs to be formatted properly on HQ side


ctsims:
Nice! @wpride Am wondering whether we can shift this code into a place where we will be able to take advantage of it for the CLI as well?

wpride:
This is awesome! Agreed we should get this into `core` too

wpride:
@benrudolph failing. shibby!

benrudolph:
needs this guy: https://github.com/dimagi/commcare-core/pull/589

PR #410, opened 2017-06-06
Add device_id param to Formplayer sync requests

wpride:
cc @ctsims

### DIFF
--- a/src/main/java/services/RestoreFactory.java
+++ b/src/main/java/services/RestoreFactory.java
@@ -348,6 +348,14 @@ public class RestoreFactory implements ConnectionHandler{
         return storage.getMetaDataFieldForRecord(users.firstElement(), User.META_SYNC_TOKEN);
     }
 
+    // Device ID for tracking usage in the same way Android uses IMEI
+    private String getSyncDeviceId() {
+        if (asUsername == null) {
+            return "WebAppsLogin";
+        }
+        return String.format("WebAppsLogin|%s|as|%s", username, asUsername);
+    }
+
     public String getRestoreUrl(boolean overwriteCache) {
         StringBuilder builder = new StringBuilder();
         builder.append(host);
@@ -363,6 +371,9 @@ public class RestoreFactory implements ConnectionHandler{
             builder.append("&since=").append(syncToken);
         }
         */
+
+        builder.append("&device_id=" + getSyncDeviceId());
+
         if( asUsername != null) {
             builder.append("&as=" + asUsername + "@" + domain + ".commcarehq.org");
         }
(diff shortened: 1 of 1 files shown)


src/main/java/services/RestoreFactory.java
```diff
@@ -363,6 +373,9 @@ public String getRestoreUrl(boolean overwriteCache) {
             builder.append("&since=").append(syncToken);
         }
         */
+
+        builder.append("&device_id=" + getSyncDeviceId());
```
ctsims: A bit surprised that there's no need to URLEncode this parameter, but not sure if tehre's magic somewhere in htis process.

Also: as a heads up, Spring has a quite nice Url utilities class you can use to generate url's and add parameters and stuff safely without having to worry about stuff http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/util/UriComponents.html

wpride: The `RestTemplate` class handles URL Encoding for us automatically

Agreed Re: UriComponents, we use that for the claim and query screens but not here.

ctsims: Sweet! Sounds good.

wpride:
@benrudolph this is (according to Trello) due tomorrow

src/main/java/services/RestoreFactory.java
```diff
@@ -351,6 +353,14 @@ private String getSyncToken(String username) {
         return storage.getMetaDataFieldForRecord(users.firstElement(), User.META_SYNC_TOKEN);
     }
 
+    // Device ID for tracking usage in the same way Android uses IMEI
+    private String getSyncDeviceId() {
+        if (asUsername == null) {
+            return DEVICE_ID_SLUG;
+        }
+        return String.format("%s|%s|as|%s", DEVICE_ID_SLUG, username, asUsername);
```
snopoke: Would ti be possible to use the user_id instead of username? Just that we've had edge cases before where a user get's deleted and another one created with the same username.

wpride: Definitely possible, possibly harder. This was to spec though - thoughts on change @ctsims

ctsims: Hey guys,

Definitely no preference on user_id v. username. user_id seems like it has advantages for the reason simon outlined. Don't think it will be an issue for enikshay in particular due to how we are doing user management, and they are the only consumers of the behavior the device_id flag is being set for, so happy to move forward without it for now, but agree that semantically it is the right long-term behavior.

wpride: Getting the `userId` would often require a SQLite DB lookup (since the other parameters are sent in the request) so I'd prefer not to incur that extra cost/point of failure unless necessary.

ctsims: Confirming that this won't block the current app, so let's skip for now but make sure we know about the limitation moving forward.

PR #411, opened 2017-06-08
Remove datetime

benrudolph:
@wpride @ctsims


ctsims:
@benrudolph Just a quick gut check: We don't have a way to be confident that this isn't used by any live deployments in the field right now, right?

i don't think this is a common question type at all, but would be good to know if, say, we actually only turned it on a week or two ago (which I don't have context on currently)

benrudolph:
hmm i don't think we have a way, but the fact that it doesn't work except for when your application is in UTC is at least something

PR #413, opened 2017-06-15
Display Correct Inline Details

wpride:
This expands the correct `inline-detail` when clicked from Web Apps

The small but unavoidable ways these code paths (persistent case tile vs. expanded case tile) diverge is rather unfortunate. Don't love the way this turned out but workable.

Fixes: https://manage.dimagi.com/default.asp?252775
cross-request: https://github.com/dimagi/commcare-hq/pull/16656


benrudolph:
test failure

PR #414, opened 2017-06-15
Get case detail relevancies, sorting

wpride:
Contains code for sorting case detail nodesets and using case detail relevancies (thanks @amstone326 

Most of the PR diff comes from getting us back up to speed with master


wpride:
this line makes us respect case detail relevancies: https://github.com/dimagi/commcare-core/compare/96169d4a07109d3ef459a72c5f77bd95f306b70e...da2b745343bfe4957774a1a690f64397b38c0285#diff-9633adeb2fae35fec66eea6cf8722256R168

src/main/java/beans/menus/EntityListResponse.java
```diff
@@ -181,7 +181,7 @@ private static void sort(List<Entity<TreeReference>> entityList, Detail shortDet
 
     static class LogNotifier implements EntitySortNotificationInterface {
         @Override
-        public void notifyBadfilter(String[] args) {
+        public void notifyBadFilter(String[] args) {
```
benrudolph: heh

wpride: thanks @amstone326

PR #415, opened 2017-06-16
Menu debugger route

benrudolph:
Alright @wpride, this is it! This gives us the ability to evaluate xpath against a menu session. Corresponding HQ PR coming, but that one is much simpler after the refactor

cc: @orangejenny


src/main/java/application/MenuController.java
```diff
@@ -129,6 +128,7 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
                 sessionNavigationBean.getSearchText()
         );
         Screen currentScreen = menuSession.getNextScreen();
+        menuSessionRepo.save(new SerializableMenuSession(menuSession));
```
benrudolph: this'll save many more menu sessions. i think a follow up item for this is to figure out a way to delete them periodically

src/main/java/beans/EvaluateXPathRequestBean.java
```diff
@@ -12,7 +12,6 @@
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class EvaluateXPathRequestBean extends SessionRequestBean {
-    private Map<String, Object> formContext;
```
benrudolph: don't think this was ever being used

PR #418, opened 2017-06-17
Error when adding choice with spaces

wpride:
fixes https://manage.dimagi.com/default.asp?256187


benrudolph:
i wonder if any other special characters cause issues. this one seems the most likely though

PR #421, opened 2017-06-20
Properly send datetime

benrudolph:
@wpride

### DIFF
--- a/src/main/java/api/json/PromptToJson.java
+++ b/src/main/java/api/json/PromptToJson.java
@@ -134,6 +134,7 @@ public class PromptToJson {
             case Constants.DATATYPE_TEXT:
             case Constants.DATATYPE_BINARY:
             case Constants.DATATYPE_TIME:
+            case Constants.DATATYPE_DATE_TIME:
                 obj.put("answer", answerValue.getDisplayText());
                 return;
             case Constants.DATATYPE_INTEGER:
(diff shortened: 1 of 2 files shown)


benrudolph:
actually one sec

benrudolph:
@ctsims @wpride so this solution will work if we remove this line from the xform parser https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xform/parse/XFormParser.java#L286 as you suggested clayton. now we just have so configure that parser for formplayer. is the best solution just to subclass it?

ctsims:
The XFormParser is annoying because a long time ago someone wrote it as a huge static monstrosity and we've been trying to detangle that ever since.

The "safest" option would be to pass in an argument in the constructor (or add a method that sets a a property on the Parser object) which just adds a line to 

https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xform/parse/XFormParser.java#L2943

which ignores a "blacklist" of data types for the current parser object. 

IE:

```
addBlackListedDataType(type) { newBlackListField.add(type)}
...
then replace above
if( !datatypes.contains(string) or blacklist.contains(stringortype) { return UNKNOWN;}
```

It's gross, but I honestly can't guarantee right now that the static semantics of the handlers are clear and it doesn't seem safe to remove/re-add the item from the  static handlers.

PR #426, opened 2017-06-30
Respect LiveQuery application configuration

wpride:
cross-request: https://github.com/dimagi/commcare-hq/pull/16859

cc @ctsims @millerdev

### DIFF
--- a/src/main/java/services/RestoreFactory.java
+++ b/src/main/java/services/RestoreFactory.java
@@ -85,14 +85,16 @@ public class RestoreFactory implements ConnectionHandler{
     private final Log log = LogFactory.getLog(RestoreFactory.class);
 
     private Connection connection;
+    private boolean useLiveQuery;
 
-    public void configure(AuthenticatedRequestBean authenticatedRequestBean, HqAuth auth) {
+    public void configure(AuthenticatedRequestBean authenticatedRequestBean, HqAuth auth, boolean useLiveQuery) {
         this.setUsername(authenticatedRequestBean.getUsername());
         this.setDomain(authenticatedRequestBean.getDomain());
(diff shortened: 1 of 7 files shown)


src/main/java/services/RestoreFactory.java
```diff
@@ -414,4 +420,12 @@ public String getAsUsername() {
     public void setAsUsername(String asUsername) {
         this.asUsername = asUsername;
     }
+
+    public boolean isUseLiveQuery() {
```
proteusvacuum: does this get used anywhere?

src/main/java/services/RestoreFactory.java
```diff
@@ -377,6 +378,11 @@ public String getRestoreUrl(boolean overwriteCache) {
             builder.append("&since=").append(syncToken);
         }
         builder.append("&device_id=").append(getSyncDeviceId());
+
+        if (useLiveQuery) {
+            builder.append("&livequery=true");
```
proteusvacuum: I think this should be `case_sync=livequery`

proteusvacuum: This should be case_sync=livequery

src/main/java/services/RestoreFactory.java
wpride: Its Spring standard to have getters and setters for your all your class variables

src/main/java/services/RestoreFactory.java
wpride: Fixed, thanks

src/main/java/services/RestoreFactory.java
proteusvacuum: should you then use it on line 382?

PR #427, opened 2017-07-06
Always use the sync token, never overwrite cache

wpride:
Removes the `override_cache=true` parameter in favor of using the `since` token parameter in all cases. This brings Formplayer in line with how Android handles restores.

Worth noting that this removes all paths for blowing away user DBs manually - now this will only happen when the DBs have expired (as determined by the sync frequent property).


proteusvacuum:
Cool. I guess you can always add a "clear user data" button somewhere like Android has?

PR #428, opened 2017-07-06
Purge cases after syncing

wpride:
@ctsims Just comparing to the Android code looks like this `purgeCases()` call is possibly the missing element


ctsims:
looks good to merge once tests pass

PR #433, opened 2017-07-17
Remove emails now that Sentry is configured

wpride:


benrudolph:
i wonder if we should leave this logic and just remove the ansible creds

benrudolph:
cool fine with this, should probably remove the asnible logic

PR #439, opened 2017-07-21
Fix graph integer parsing

wpride:
Fixes https://manage.dimagi.com/default.asp?258684

Remembered that I fixed this before, looks like I lost the fix in the merge here https://github.com/dimagi/commcare-core/commit/126334e2de62579c78fca6950a12ed9b5c974358#diff-e65fec0df94c425f16cbe751a2b56a4aR116


wpride:
@benrudolph

PR #441, opened 2017-07-27
Add a few more details to sentry

benrudolph:
@wpride @ctsims 

buddy: @proteusvacuum


src/main/java/services/RestoreFactory.java
```diff
@@ -370,7 +371,7 @@ public String getRestoreUrl() {
         builder.append("/a/");
         builder.append(domain);
         builder.append("/phone/restore/?version=2.0");
-        String syncToken = getSyncToken(getWrappedUsername());
```
wpride: whoops

PR #445, opened 2017-08-03
Better encapsulate SQLite dbs

dannyroberts:
This is a bit of a monster, but I did it over many relatively small commits. I made heavy use of IntelliJ's refactoring commands, and a number of commits are just renames.

A rundown of the new organization:
- `ApplicationDBPath` and `UserDBPath` are more or less static accessor classes for path strings, and both implement the (small) `DBPath` interface.
- `SQLiteDB` is a class that wraps a `DBPath` that you pass it and essentially gives it functionality, allowing you to manipulate the files (create folder structure, see if they exist, get a sqlite connection, etc.). It implements `ConnectorHandler`.
- `ApplicationDB` and `UserDB` are very light extensions of `SQLiteDB` that add custom constructors. `new UserDB(domain, username, asUsername)` is essentially syntactic sugar for `new SQLiteDB(new UserDBPath(domain, username, asUsername))`.
- Because of this syntactic sugar, `UserDBPath` and `ApplicationDBPath` are able to be package-private.
- `FormplayerStorageFactory` and `RestoreFactory` no longer implement `ConnectionHandler`; instead, where they used to have a string `databasePath` they now have a `SQLiteDB` `sqLiteDB` (which implements `ConnectionHandler`), and where they used to pass `this` to instantiate other classes, they now pass `this.sqLiteDB`; other external usages of `factory.getConnection()` and `factory.closeConnection()` have been replaced with `factory.getSQLiteDB().getConnection()` and `factory.getSQLiteDB().closeConnection()`
- all external manipulation of sqlite files by getting the filename, etc is now done through the `sqLiteDB` object instead, e.g. `restoreFactory.getSqLiteDB().createDatabaseFolder()`.
- As a result, I was able to delete `ApplicationUtils`; as well as `TestConnectionHandler`, since `UserDB` is now just as easy to instantiate if not easier.


dannyroberts:
@ctsims FYI since we chatted about this ealier

dannyroberts:
Also FWIW it seems to work—as in I was able to start it up and use Web Apps (including initial sandbox creation) and click around a small app.

src/main/java/dbpath/DBPathConnectionHandler.java
```diff
@@ -60,4 +61,16 @@ public void closeConnection() {
...
+
+    public void deleteDatabaseFolder() {
+        SqlSandboxUtils.deleteDatabaseFolder(dbPath.getDatabaseFile());
+    }
+
+    public boolean createDatabaseFolder() {
+        return new File(dbPath.getDatabaseFile()).getParentFile().mkdirs();
+    }
+
+    public String getDatabaseFileForLoggingPurposes() {
```
benrudolph: the reason this is so specific is that you don't think we'll ever need just the raw path / if we do we're breaking the abstraction?

dannyroberts:
Good tip about testing. I'll make sure to do that.

src/main/java/dbpath/DBPathConnectionHandler.java
dannyroberts: Yeah exactly. In the end I changed it minorly to `getDatabaseFileForDebugPurposes` which loosely includes both logging/printing out for debugging as well as for use in tests, some of which want to be able to test at the system level that a file is actually there or not.

dannyroberts:
> one thing to test out is opening app preview, loading the app, making a change, hitting refresh and then ensuring the change is properly updated. 

just followed these steps exactly working against this branch and everything worked as expected.

wpride:
👍 This is great Danny, thank you!

PR #446, opened 2017-08-04
Skip user restore in formplayer

wpride:
Hacky potential fix for https://manage.dimagi.com/default.asp?259485

Hold off on merge in case they can work around this


wpride:
Merging per Marissa's request

PR #447, opened 2017-08-08
Check database path rather than folder

wpride:
Should be checking whether parent directory exists rather than DB itself


wpride:
merging pre-tests since this is passing locally and breaking live preview

PR #450, opened 2017-08-11
use session details view for auth

snopoke:
Instead of looking up the session in the Django DB we call a service on HQ to get the user details we need.

Can review by commit :fish: 

cc @benrudolph (for kicks)


wpride:
@snopoke Woah, this is awesome! Removing the HQPostgres and Couch dependencies is huge!

snopoke:
There's some more cleanup to do still but I thought I'd leave that for
another PR

snopoke:
Merging now that HQ endpoint has been deployed as well as auth keys changes

snopoke:
deployed this to prod - looking good so far

PR #451, opened 2017-08-11
Fixes for inline and persistent case tiles

wpride:
Makes persistent case tiles without inline details un-clickable

Fixes a bug where inline case tile clicks could result in advancing the session

fixes https://manage.dimagi.com/default.asp?259461

cross-request: https://github.com/dimagi/commcare-hq/pull/17442


src/main/java/application/AbstractBaseController.java
```diff
@@ -253,18 +253,22 @@ protected EntityDetailListResponse getInlineDetail(MenuSession menuSession) {
...
 
-        EvaluationContext ec = session.getEvaluationContext();
-
-        TreeReference ref = entityDatum.getEntityFromID(ec, stepToFrame.getValue());
-        if (ref == null) {
-            return null;
+        EvaluationContext ec;
+        if (inline) {
+            ec = session.getEvaluationContext();
+            return new EntityDetailListResponse(persistentDetail.getFlattenedDetails(),
```
wpride: @ctsims one question I had was about this `getFlattenedDetails()` method. This isn't used in the Android or Formplayer repositories, but seems like the correct function call here - that is, if we have a list of detail tabs we want to display those, if not we just want to use the fields in this detail. Is that assumption correct?

ctsims: I'm not 100% sure what the context is here, but on the inline popup we want to display the tabs, not the flattened detail. The flattened detail was actually just for j2me where we couldn't display tabs.

src/main/java/beans/menus/EntityDetailResponse.java
```diff
@@ -199,4 +200,12 @@ public void setUseNodeset(boolean useNodeset) {
     public void setEntities(EntityBean[] entities) {
         this.entities = entities;
     }
+
+    public boolean isHasInlineTile() {
```
dannyroberts: maybe `getHasInlineTile`?

wpride: Good call, stupid IntelliJ auto fill-in

PR #452, opened 2017-08-23
Gzip archiving

dannyroberts:
This PR provides
- functionality for detecting and unarchiving .gz archive when the original db file is not present, and working transparently from there as if it had been present all along
- a script that can be used in a cron job to archive all .db files in a directory that haven't been accessed for a given period of time. (This will not be "live" until we add such a cron job in ansible.)

Thus the only immediate effect of this PR is to turn on the ability to read from *.db.gz files, but doesn't actually turn on the archiving that will create them yet.

To test this out locally (with the archiving), you can (from the formplayer directory) run (assuming you're storing dbs in `formplayer/dbs`)

```bash
watch './scripts/archive_dbs.sh dbs +10s'
```

in one terminal window to archive all dbs not accessed in the last 10 seconds and

```bash
watch 'find dbs -type f'
```
in another to watch the files in that directory as they change. Then you can spin up formplayer and commcare-hq and play around with webapps and watch what happens to the .db and .db.gz files as you access (or don't access) certain dbs.

I've tested that this works very smoothly even when you're archiving files not used in the last second realtime and clicking around webapps, submitting forms, switching apps, using login-as, etc. Of course, this is with tiny db files where gzipping and ungzipping is instantaneous, so I would recommend a much larger `-atime` of `+15m` or so.

That said, I've carefully designed both processes (java unarchiving and bash archiving) to create a .lock file before any gzip steps, and delete it after (even if the gzip step fails). The only risk I can see here is that if a .lock file _does_ somehow end up there with no process around to delete it, accessing that db will hang forever. The only way I know of that that could happen is if someone manually creates that file, or I suppose if the java or bash process is hard-killed at exactly the wrong time.

### DIFF
--- a/src/main/java/sqlitedb/SQLiteDB.java
+++ b/src/main/java/sqlitedb/SQLiteDB.java
@@ -8,6 +8,7 @@ import services.ConnectionHandler;
 
 import javax.sql.DataSource;
 import java.io.File;
+import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
 
@@ -21,6 +22,11 @@ public class SQLiteDB implements ConnectionHandler {
     }
 
     private Connection getNewConnection() throws SQLException {
+        try {
+            SqlSandboxUtils.unarchiveIfArchived(dbPath.getDatabaseName(), dbPath.getDatabasePath());
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
         DataSource dataSource = SqlSandboxUtils.getDataSource(dbPath.getDatabaseName(), dbPath.getDatabasePath());
         return dataSource.getConnection();
     }
--- a/src/main/java/sqlitedb/SQLiteDB.java
+++ b/src/main/java/sqlitedb/SQLiteDB.java
@@ -3,6 +3,7 @@ package sqlitedb;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.sqlite.SQLiteConnection;
+import sandbox.ArchivableFile;
 import sandbox.SqlSandboxUtils;
 import services.ConnectionHandler;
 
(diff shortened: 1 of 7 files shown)


dannyroberts:
@emord @gcapalbo I think both of you have looked into formplayer space issues recently.

I think you both have context, but in case not, I've tested with some of the larger production sqlitedbs and they have really good compression ratios (as do smaller ones), easily > 90%—i.e. they create files a tenth the size.

src/main/java/sqlitedb/SQLiteDB.java
```diff
@@ -3,25 +3,40 @@
...
+            dbArchivableFile = new ArchivableFile(dbPath.getDatabaseFile());
+        }
     }
 
     private Connection getNewConnection() throws SQLException {
-        DataSource dataSource = SqlSandboxUtils.getDataSource(dbPath.getDatabaseName(), dbPath.getDatabasePath());
+        try {
+            dbArchivableFile.unarchiveIfArchived();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
wpride: The idea is that we only reach this line if we had an archive, but not a DB, and then failed to expand that DB, right? If so, then I'm thinking it might be better to just delete the folder here and re-instantiate a new database connection rather than failing hard

wpride:
Just the one question, otherwise looks great

src/main/java/sqlitedb/SQLiteDB.java
dannyroberts: Yeah, I'm not totally sure. There are a number of method calls within `unarchiveIfArchived` that have the potential to throw an `IOException`, and I'd rather see the error happen and then deal with each kind of situation specifically then silence all potential errors. For example, let's say we silenced the error, and the specific potential error you mentioned does happen,maybe even systematically. Well in that case, we'll be constantly deleting and rebuilding these database, but there will be no indication of that, only a grinding slowness in an already slow system that will be hard to locate or pinpoint.

On the other hand, if we do get an error, we can go straight to the place, see the weird state, and try and figure out exactly why it happened, how widespread it is, etc. and then decide how we want to handle it.

I guess you could make the case to report it as an error, but do automatic cleanup. I'm actually unfamiliar with the tools we'd use in java/formplayer to do that, but I'm also just tempted to leave it be and deal with it if and when it becomes an issue

wpride: Cool, that sounds fine

wpride:
Did you want to test this out more or gtm?

dannyroberts:
OK, I tested and everything works except that `find` on linux doesn't support time units on `-atime` (the access time arg), so the archive_dbs.sh script fails. (I'm going to have to fix it up by using the `-amin` arg instead and probably changing the syntax a bit.)

Given that script isn't hooked up to anything yet, I don't think that needs to block, so I'll merge!

emord:
> OK, I tested and everything works except that find on linux doesn't support time units on -atime (the access time arg), so the archive_dbs.sh script fails. (I'm going to have to fix it up by using the -amin arg instead and probably changing the syntax a bit.)

@dannyroberts it's fairly common practice to mount drives with the option `noatime` to reduce load on the filesystem. I checked hqformplayer1 and we use that option. I think it would be better to use mtime (which is what the tmpreaper in ansible uses)

dannyroberts:
Great, thanks for the tip @emord!

PR #456, opened 2017-08-23
Better exception for hashtable NPE

wpride:
Think this will help track down some of the Enikshay case claim exceptions


wpride:
@dannyroberts confirmed

PR #459, opened 2017-08-25
add duration buckets to datadog requests metric

dannyroberts:
buckets are (in both lexical and numerical order):

- `lt_001s`
- `lt_005s`
- `lt_020s`
- `lt_060s`
- `lt_120s`
- `over_120s`

I chose these buckets to match the ones in https://app.datadoghq.com/dash/74381/hq-vitals?live=true&page=0&is_auto=false&from_ts=1503660835264&to_ts=1503675235264&tile_size=m's "Restore Count by Time Bucket" graph (`commcare.restores.count` metric), and added the `lt_001s` bucket, since it's more relevant in formplayer's case as it supports a realtime web interface.

### DIFF
--- a/src/main/java/aspects/MetricsAspect.java
+++ b/src/main/java/aspects/MetricsAspect.java
@@ -34,22 +34,43 @@ public class MetricsAspect {
             user = bean.getUsernameDetail();
         }
 
+        long startTime = System.nanoTime();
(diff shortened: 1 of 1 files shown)


dannyroberts:
light local testing with the debugger indicates this is doing the right thing (and helped me catch an off-by-a-factor-of-10 error).

PR #461, opened 2017-08-29
Shorten user lock timeout

dannyroberts:
and change response behavior on the lock timeout and lock expiry.

EDIT: I originally was going to change it to 7 seconds, but @ctsims convinced me to try 21 seconds instead, as thousands of requests a day fall into the < 20s bucket.

The impetus here is that the _vast_ majority of requests take under 20 seconds, so if you can't obtain a lock in 21 seconds, chances are you're waiting on a loooong process or a deadlock, and it's relatively likely you won't be able to obtain it in 120 seconds either—so there's no point in presenting that baffling behavior where you do some super simple thing and hangs for 2 minutes, and then fails anyway.

This should only get merged after https://github.com/dimagi/commcare-hq/pull/17639, which improves the behavior of server errors like this one to be less disruptive to the user's workflow.

It will be easy to tell if I've set the timeout too low: my hypothesis is that the lock timemout numbers in https://app.datadoghq.com/dash/232351/formplayer-error-tracking?live=true&page=0&is_auto=false&from_ts=1503867265434&to_ts=1504040065434&tile_size=m&tpl_var_environment=enikshay will rise only slightly; if they spike, then we need to increase the timeout, because it means that a ton of things that were waiting 21+ seconds but eventually obtaining the lock before are now timing out and erroring.

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -1,5 +1,6 @@
 package application;
 
+import aspects.LockAspect;
 import auth.DjangoAuth;
 import auth.HqAuth;
 import auth.TokenAuth;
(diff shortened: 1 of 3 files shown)


snopoke:
Build failure seemingly unrelated 
```
:commcare:test
org.javarosa.core.model.utils.test.LocalizerTest > testLinearSub FAILED
    java.lang.RuntimeException at LocalizerTest.java:695
543 tests completed, 1 failed
:commcare:test FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':commcare:test'.
> There were failing tests. See the report at: file:///home/travis/build/dimagi/formplayer/libs/commcare/build/reports/tests/test/index.html
```

wpride:
Kicked build... that failure seems to occur periodically

wpride:
Looks good to me.

@ctsims - are you thinking of anything more complicated than pointing to this PR and letting them know to revert if something comes up? Alternatively/additionally we could deploy Formplayer on prod early EST (once India quiets down) and then see how it goes

snopoke:
It's quite easy to rollback a formplayer deploy now if it doesn't go well.

PR #463, opened 2017-08-30
Fine-grained Datadog timings logging for expensive operations

wpride:
And refactor to make this cleaner. New timings:

- Submit

~~- Sync~~

- Claim

- Install

Removed sync timings due to @millerdev handling that here https://github.com/dimagi/commcare-hq/pull/17653/files

### DIFF
--- a/src/main/java/services/impl/FormRecordProcessorImpl.java
+++ b/src/main/java/services/impl/FormRecordProcessorImpl.java
@@ -81,7 +81,7 @@ public class FormRecordProcessorImpl extends XmlFormRecordProcessor implements R
      * TODO They should be unified
      *
      */
-    @MethodMetrics(action = "case-purge")
+    @MethodMetrics
     public void purgeCases(UserSqlSandbox sandbox) {
         long start = System.currentTimeMillis();
         //We need to determine if we're using ownership for purging. For right now, only in sync mode
--- a/src/main/java/services/impl/FormRecordProcessorImpl.java
+++ b/src/main/java/services/impl/FormRecordProcessorImpl.java
@@ -1,5 +1,6 @@
 package services.impl;
 
+import annotations.MethodMetrics;
 import com.timgroup.statsd.StatsDClient;
 import database.models.FormplayerCaseIndexTable;
 import engine.FormplayerTransactionParserFactory;
@@ -80,6 +81,7 @@ public class FormRecordProcessorImpl extends XmlFormRecordProcessor implements R
      * TODO They should be unified
      *
      */
+    @MethodMetrics(action = "case-purge")
     public void purgeCases(UserSqlSandbox sandbox) {
         long start = System.currentTimeMillis();
         //We need to determine if we're using ownership for purging. For right now, only in sync mode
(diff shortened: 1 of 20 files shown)


src/main/java/services/impl/FormRecordProcessorImpl.java
```diff
@@ -112,6 +139,13 @@ public static void purgeCases(UserSqlSandbox sandbox) {
         log.info(String.format(
                 "Purged [%d Case, %d Ledger] records in %dms",
                 removedCaseCount, removedLedgers, taken));
-
+        datadogStatsDClient.recordExecutionTime(
+                Constants.DATADOG_TIMINGS,
+                taken,
+                "domain:" + restoreFactory.getDomain(),
+                "username" + restoreFactory.getWrappedUsername(),
+                "request:" + "case-purge",
```
dannyroberts: Should we change this to something like "action" instead of "request"? Wherever we use "request" in formplayer datadog metrics it usually means the url that was hit; I think using that word now to mean essentially which function call we're timing is confusing. (The same comment would apply to other uses of "request" below as well.)

src/main/java/services/impl/FormRecordProcessorImpl.java
```diff
@@ -112,6 +139,13 @@ public static void purgeCases(UserSqlSandbox sandbox) {
...
+                "case-count" + removedCaseCount
```
dannyroberts: I think there should be a `:` here, but also, I'm not sure what we get out of tracking this here. Is there a particular graph you're thinking of? In general, I think tags are really only useful with discreet values, since the way you use them is always essentially "where <tag> = <value>" or "group by <tag> value". I think if you have something in mind with this, we could bucket it (<10, <100, <1000, <10000, >10000, with tag value names that are in alphabetical order so that it splits them nicely), but if you don't have a particular thing you want to track with this, I'd just remove it.

src/main/java/services/impl/FormRecordProcessorImpl.java
```diff
@@ -35,11 +41,32 @@
...
+    protected StatsDClient datadogStatsDClient;
 
-    public static void processXML(FormplayerTransactionParserFactory factory, String fileText) throws IOException, XmlPullParserException, UnfullfilledRequirementsException, InvalidStructureException {
+    @Autowired
+    protected RestoreFactory restoreFactory;
+
+    private final Log log = LogFactory.getLog(FormRecordProcessorImpl.class);
+
+    // This function will only wipe user DBs when they have expired, otherwise will incremental sync
+    public UserSqlSandbox performSync(RestoreFactory restoreFactory) throws Exception {
```
dannyroberts: Rather than moving all of the things that you want to time into the same parent class, could you add a `@ReportTiming` annotation with an aspect that does times it and sends to datadog? Or was there a different reason for reorganizing these functions into a single parent class?

src/main/java/services/impl/FormRecordProcessorImpl.java
snopoke: The format of tags must be 'name:value' (username and case-count are missing ':')

It's not useful to have random numerical values as tags. If you want to have a case-count tag you should bucket.

Is it necessary to have `username` tag - just that it will push up the metric count a lot: https://help.datadoghq.com/hc/en-us/articles/204271775-What-is-a-custom-metric-and-what-is-the-limit-on-the-number-of-custom-metrics-I-can-have-

millerdev: I'd also add that `username` is not likely to be a very useful tag in graphs if there are a high volume of users on your graph. I've worked around this in the past by logging that information to a file rather than sending to datadog.

Another related thing I've done is only tag certain values (like if you only care about certain usernames, put them in a whitelist and only add that tag if the username is in the list). We do that for `domain` in HQ since tagging every domain would push up the custom metric count a lot, but there are certain key domains that we'd like to have tagged (`rec`, for example).

millerdev: Ah, missed @dannyroberts comment above. He said it better.

dannyroberts:
any updates here?

PR #466, opened 2017-09-07
Send lock expiry to sentry

dannyroberts:
and also add an `as_user` tag to sentry reports


dannyroberts:
I'm testing locally, but seems like my "fix" for tests might have just made `request` always `null` which wouldn't be ideal. Going to keep looking.

dannyroberts:
Okay, never mind, I messed up using the debugger. `request` is present and non-null when it needs to be.

PR #467, opened 2017-09-08
send info for long running requests to sentry

dannyroberts:
I'm guessing we'll get about ~15-30 of these from production and ~3-18 of these from enikshay a day at this point. It'll give us a chance to dig into what's causing the slowness in each case, whether there are certain problem users that get a disproportionate number of them, etc.

### DIFF
--- a/src/main/java/aspects/MetricsAspect.java
+++ b/src/main/java/aspects/MetricsAspect.java
@@ -1,6 +1,9 @@
 package aspects;
 
 import beans.AuthenticatedRequestBean;
+import com.getsentry.raven.event.Breadcrumb;
+import com.getsentry.raven.event.BreadcrumbBuilder;
+import com.getsentry.raven.event.Event;
 import com.timgroup.statsd.StatsDClient;
 import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.annotation.Around;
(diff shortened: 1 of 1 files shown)


src/main/java/aspects/MetricsAspect.java
```diff
@@ -55,9 +64,23 @@ public Object logRequest(ProceedingJoinPoint joinPoint) throws Throwable {
...
     }
 
+    private void sendTimingWarningToSentry(final long timeInMs) {
+        raven.recordBreadcrumb(new BreadcrumbBuilder() {{
+            setCategory("long_request");
+            setLevel(Breadcrumb.Level.WARNING);
+            setData(new HashMap<String, String>() {{
+                put("duration", String.format("%.3fs", timeInMs / 1000.));
+            }});
+        }}.build());
```
dannyroberts: @wspride I like this syntax better than the more linear one, where you declare the hashmap, then put a thing into it, then declare the builder, then add things to it, including the hashmap you declared and populated above, and then finally pass `builder.build()` to raven. It's more like what I'm used to doing in python, and more clearly shows that all of these things are really just a single conceptual step of recording a breadcrumb. Do you have thoughts? (And this is supported in the java version we're using, right?)

wpride: I don't really have a problem with this (looks pretty) but doesn't seem to be beloved by Java purists: https://stackoverflow.com/a/14878801. Given how small the scope is here I think no problem

PR #469, opened 2017-09-08
Collapse interface and implementation

dannyroberts:
If anyone wants to advocate for keeping this pattern, I'd love to hear the explanation. Nothing I've read online has been that satisfying and lots of people seem to advocate against it. If and when there's some reason to create an interface that multiple classes can implement, it's pretty easy to do—especially if we've been appropriately marking methods private; until then it's just confusing and makes clicking through to method implementations disorienting and annoying.


dannyroberts:
@gcapalbo I agree it's worth being cautious here, even though tests pass. I'm going to deploy this on staging and then go through some basic operations to make sure nothing's different.

dannyroberts:
OK I've clicked around on staging and haven't seen any issues. Going to go ahead with this before things get too stale

PR #471, opened 2017-09-11
try 2: clean up builder syntax for sentry breadcrumbs

dannyroberts:
Re-did this in a way that doesn't cause https://manage.dimagi.com/default.asp?261639 (failing on null values in data).

I'm going to see if I can write a test into this that would have caught the other bug.

### DIFF
--- a/src/main/java/application/WebAppContext.java
+++ b/src/main/java/application/WebAppContext.java
@@ -3,7 +3,6 @@ package application;
 import aspects.*;
 import com.getsentry.raven.RavenFactory;
 import com.getsentry.raven.dsn.InvalidDsnException;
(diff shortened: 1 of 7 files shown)


dannyroberts:
@gcapalbo I ended up changing the approach to make it more directly testable since your review.

The idea is basically to push more of the building syntax into the relatively "pure" `FormplayerRaven` class and then test that interface more directly. At this point it's feeling like a lot of fuss on my part for a relatively simple thing, but I do think it's worth it being super simple to add and edit sentry breadcrumbs, and it's worth being able to test the code we use to do that. (And it's worth my having learned how to writes tests and use mocks in formplayer...)

dannyroberts:
Tested a bit on staging this time.

PR #472, opened 2017-09-12
log lock timings to datadog and sentry

dannyroberts:
datadog: *.granular.timings metric, category "wait_on_lock";
sentry: breadcrumb, category "wait_on_lock"

### DIFF
--- a/src/main/java/aspects/LockAspect.java
+++ b/src/main/java/aspects/LockAspect.java
@@ -53,7 +53,7 @@ public class LockAspect {
         Lock lock;
 
         try {
-            lock = getLockAndBlock(username);
+            lock = getLockAndBlock(username, joinPoint);
         } catch (LockError e) {
             logLockError(bean, joinPoint, "_timed_out");
             throw e;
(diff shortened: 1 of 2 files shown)


dannyroberts:
Had to merge https://github.com/dimagi/formplayer/pull/471 but now it's good to go.

src/main/java/aspects/LockAspect.java
```diff
@@ -84,8 +84,24 @@ private void logLockError(AuthenticatedRequestBean bean, ProceedingJoinPoint joi
...
 
-    private Lock getLockAndBlock(String username) throws LockError {
+    private Lock getLockAndBlock(String username, ProceedingJoinPoint joinPoint)
+            throws LockError {
+        long startTime = System.nanoTime();
         Lock lock = userLockRegistry.obtain(username);
+        long timeInMs = (System.nanoTime() - startTime) / 1000000;
+        raven.newBreadcrumb()
+                .setCategory(Constants.TimingCategories.WAIT_ON_LOCK)
+                .setMessage(timeInMs > 5 ? "Had to wait to obtain the lock"
```
wpride: Why 5?

dannyroberts: Arbitrary, I just saw locally that it was 0-2m each time. It's probably more on the server because of the network time to redis (locally it's all on one machine), so if I see a lot of "Had to wait to obtain the lock"s with 10-20 seconds, I'll up it. It's not super crucial, just wanted to give someone a cue they could find at a quick glance (or with Cmd-F), so if it takes some calibration/correction later, it's not a big deal.

dannyroberts:
Thanks for the quick review!

PR #477, opened 2017-09-12
Simplify timing

dannyroberts:
Quick prefactor to make adding standardized category timings easier.

The idea with category timings is that they should be _mutually exclusive_, so that all together they add up to less than the total time a request took—i.e. don't time a function and then also time a smaller potion within it.

To add a timing now, it looks like this:

```
CategoryTimingHelper.RecordingTimer timer = categoryTimingHelper.newTimer("my_category");
timer.start();
# do the thing you want to time
timer.end()
        .setMessage("My message which can be a function of timer duration")
        .record();
```

(if there's one already there, if the thing is in a `try`-`catch`, consider putting the `timer.end()...` step in a `finally`).

### DIFF
--- a/src/main/java/services/CategoryTimingHelper.java
+++ b/src/main/java/services/CategoryTimingHelper.java
@@ -1,0 +1,70 @@
+package services;
+
+import com.timgroup.statsd.StatsDClient;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import util.Constants;
+import util.FormplayerHttpRequest;
+import util.FormplayerRaven;
+import util.SimpleTimer;
+
+@Component
+public class CategoryTimingHelper {
+    @Autowired
+    private FormplayerHttpRequest request;
+
+    @Autowired
+    private StatsDClient datadogStatsDClient;
+
+    @Autowired
+    private FormplayerRaven raven;
+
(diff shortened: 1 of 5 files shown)


src/main/java/services/CategoryTimingHelper.java
```diff
@@ -0,0 +1,70 @@
...
+    @Autowired
+    private FormplayerHttpRequest request;
+
+    @Autowired
+    private StatsDClient datadogStatsDClient;
+
+    @Autowired
+    private FormplayerRaven raven;
+
+    public class RecordingTimer extends SimpleTimer {
```
gcapalbo: is there an advantage to putting this class definition within the `CategoryTimingHelper` as opposed to having it be in its own file?

dannyroberts: Yeah I don't know; I guess I was considering it more of an extension of `CategoryTimingHelper` than its own thing; for example when it takes a `CategoryTimingHelper parent` arg, it's doing so to gain access to `CategoryTimingHelper`'s private functions and (indirectly) its `Autowired` globals. You'd never have one without the other so this nested class seemed like the most obvious way to show that.

`SimpleTimer` on the other hand is useful and used elsewhere, so that is its own separate class.

I'm of course carrying a lot of my sensibilities about these things over form python where it's customary to put more things in a single file, but I guess I figured if inner classes are ever acceptable in Java, this would be one such case.

PR #482, opened 2017-09-25
add purge_cases category timings

dannyroberts:
Add timings for https://manage.dimagi.com/default.asp?262477


ctsims:
One thing here: Tehre are definitely two purge code paths that Danny and I found earlier (one is used when syncing, the other when submitting forms). Can we add this for both?

PR #484, opened 2017-09-26
fix request param in datadog

dannyroberts:
the overall timing metric and lock metric were pulling
from the route name rather than the url.
this made navigate_menu vs navigate_menu_start indistinguishable


dannyroberts:
Hmm, looks like I'm going to have to take a different approach. Getting anything to work in `@Bean`-land is always a struggle.

PR #488, opened 2017-09-29
Refactor syncToken management

wpride:
Resolves https://manage.dimagi.com/default.asp?261905

Convoluted but straightforward bug. When looking up the current user in the users database we compare the name strings. ~~The target name is escaped (`-` and `.` become `_`) but the names stored in the database are not. So if we are looking for the user named EG `district.support.test` we will never find this user and thus never set the sync token. This causes subsequent restores to use the cached lookup as we do not have a sync token.~~

Turned this into a bigger refactor. Formplayer borrowed a lot of unnecessary code from Android here - namely, our DBs will only have one user on them (Which we rely on elsewhere in the code) so iterating over the users comparing names is unnecessary.

I've also made a change so that we update the user's sync token directly in storage once the sync token is parsed rather than at the end - @ctsims ~~is there a good reason Android waits until pa~~ duh, because we don't know if the sync has succeeded yet.

### DIFF
--- a/src/main/java/hq/CaseAPIs.java
+++ b/src/main/java/hq/CaseAPIs.java
@@ -67,18 +67,6 @@ public class CaseAPIs {
         ParseUtils.parseIntoSandbox(restorePayload, factory, true, true);
         restoreFactory.commit();
         restoreFactory.setAutoCommit(true);
-        // initialize our sandbox's logged in user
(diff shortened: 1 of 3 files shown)


dannyroberts:
@wpride I'll review this shortly

PR #495, opened 2017-10-04
Support for requests during form entry

wpride:
cross-request/context in https://github.com/dimagi/commcare-core/pull/653/files

cc @ctsims @esoergel

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -23,6 +23,7 @@ import org.commcare.suite.model.Detail;
 import org.commcare.suite.model.EntityDatum;
 import org.commcare.suite.model.StackFrameStep;
 import org.commcare.util.screen.*;
+import org.javarosa.core.model.actions.FormSendCalloutHandler;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.xml.util.InvalidStructureException;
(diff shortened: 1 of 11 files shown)


src/main/java/application/AbstractBaseController.java
```diff
@@ -87,6 +88,9 @@
     protected FormplayerRaven raven;
 
     @Autowired
+    protected FormSendCalloutHandler formSendCalloutHandler;
```
ctsims: I think I'm missing why it doesn't make the most sense for the FormSession to not instantiate this directly. It seems like there's a ton of bootstrapping that's being done to pass this down the chain

src/main/java/util/PrototypeUtils.java
```diff
@@ -93,7 +94,8 @@ public static void setupPrototypes(){
                 SeriesData.class.getName(),
                 Graph.class.getName(),
                 XYSeries.class.getName(),
-                BubbleSeries.class.getName()
+                BubbleSeries.class.getName(),
+                SendAction.class.getName(),
```
ctsims: Whoops, this looks like a bug on the commcare side actually, too, if didn't add this.

src/main/java/application/AbstractBaseController.java
wpride: Autowiring is more idiomatic Spring and much more feature-ful than direct instantiation (giving us access to the entire set of autowired properties/objects). Ideally we'd autowire the FormSession object as well and then we could skip this long handoff chain - good future refactor

src/main/java/util/PrototypeUtils.java
ctsims: Ha, sure enough! https://github.com/dimagi/commcare-android/pull/1851

PR #498, opened 2017-10-10
Rollback SQL writes on exception

wpride:
I * think* this will resolve some of the issues seen [here](https://manage.dimagi.com/default.asp?263442)

I believe that when we encountered certain exceptions (chiefly `InvalidStructureException`) we would error out without closing the SQLite database. This would result in the user being locked out. Until the database was cleared out and could also result in incorrect data being present in the database.


calellowitz:
I am merging this as its urgent for ICDS and that team has decided its worth the risk, that i may miss something in code review. 
@dannyroberts should still review when he comes online.

PR #504, opened 2017-10-23
Change how we load up params map into variables

wpride:
Previous method was loading these as query parameters rather than URL parameters. I must've only tested this with the values loaded directly into the URL.

@ctsims


ctsims:
@wspride have you tested that these adequately URL encode the param and value?

wpride:
@ctsims yes, the Spring framework handles URL encoding automatically for us (and I just tested to be sure)

PR #505, opened 2017-10-24
More comprehensive purge timing

dannyroberts:
Still want to test this locally, but wanted to PR this basic idea.


dannyroberts:
Yeah, I considered that, and agree it's nice to have timings even when the thing you're timing fails (sometimes especially when it fails). Here, it's the fact that I'm passing back a sub-timing in the return value that makes things a bit harrier. I could still try catch the total timing, but now that I'm subtracting purge timing from install timing, if I don't have purge timing that kind of throws off the install timing. It's probably something I could work around with more refactoring, but seems like a rabbit hole of its own.

PR #511, opened 2017-10-31
Log Cleanup

wpride:
Noticed are logs are now getting unwieldy within 24 hours. This trims down on some of the bigger frequent logs (that I've also never found really useful):

- Logging the entire question tree
- Logging all case list entities and menu list commands
- Logging entire case claim query response


dannyroberts:
Good call

PR #512, opened 2017-10-31
Purge stale form sessions

wpride:
This will remove form sessions that are older than one week old. Runs every time Formplayer boots up.


ctsims:
This checks out to me. Will, can you double check with Sheel that he's signed off on this? (just want a verbal)

wpride:
see https://github.com/dimagi/commcare-hq/pull/18418

will merge this in a week

PR #513, opened 2017-10-31
WIP eviction of RedisLocks with dead threads

wpride:
Still lots of joints to oil out but this is the general idea @dannyroberts 

Ugliest part is that the RedisLock and serialization classes are private so had to partially copy those over in order to be able to read the `threadName` from Redis

cc @ctsims

### DIFF
--- a/src/main/java/application/MenuController.java
+++ b/src/main/java/application/MenuController.java
@@ -183,7 +183,6 @@ public class MenuController extends AbstractBaseController {
     @AppInstall
     public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBean sessionNavigationBean,
                                                     @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
-        Thread.sleep(10000);
         String[] selections = sessionNavigationBean.getSelections();
         MenuSession menuSession;
         SimpleTimer timer = new SimpleTimer();
--- a/src/main/java/application/MenuController.java
+++ b/src/main/java/application/MenuController.java
@@ -183,6 +183,7 @@ public class MenuController extends AbstractBaseController {
     @AppInstall
     public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBean sessionNavigationBean,
                                                     @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
+        Thread.sleep(10000);
         String[] selections = sessionNavigationBean.getSelections();
         MenuSession menuSession;
         SimpleTimer timer = new SimpleTimer();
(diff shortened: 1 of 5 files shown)


src/main/java/application/MenuController.java
```diff
@@ -183,6 +183,7 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
     @AppInstall
     public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBean sessionNavigationBean,
                                                     @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
+        Thread.sleep(10000);
```
dannyroberts: Was this just for testing?

src/main/java/aspects/LockAspect.java
```diff
@@ -96,12 +110,42 @@ private void logLockError(AuthenticatedRequestBean bean, ProceedingJoinPoint joi
...
+            if (!threadLives(holdingThreadName)) {
+                log.error("Evicting thread " + holdingThreadName + " from lock with key " + username);
+                evictLock(username);
+                return getLockAndBlock(username);
+            }
             throw new LockError();
         }
     }
 
+    private boolean threadLives(String name) {
```
dannyroberts: And since I'm totally unfamiliar with this, how sure are you that this works? I.e. it'd be a mess if threads that were actually still running didn't show up in the list for some reason.

ctsims:
@wpride is this stalled currently? Want to get a feel for how close this is to being ready to review, since we're still experiencing a lot of performance issues on formplayer.

wpride:
@ctsims Yes - I put this together as a POC and I still have a few big questions/concerns, particularly:

1. What are the potential consequences of directly over-writing the Redis entries and how this could conflict with the library's expectations
2. Given above, maybe it would be better to make our own build of the Redis locking library and just make the necessary classes public/protected
3. Is this sort of thread checking really effective/safe

I'd like to hear your thoughts on above (and maybe @snopoke too) and can prioritize this over the weekend if there's urgency. Seems like a good solution but I'm hesitant to break so many abstraction layers

snopoke:
I think this is getting into dangerous waters and would require significant testing before we could be sure it's working well.

I'm not that familiar with the background of this work. Why are the locks not getting released in the first place? What's causing the threads to die?

The other question I had is whether redis locking is really necessary. If we can guarantee that the same user (session) always goes to the same formplayer service then do we need the distributed locking or can we just use Java's own locking mechanisms?

wpride:
@snopoke To your second question, Clayton and I discussed this briefly last week - Clayton's concern was that we were likely to want the ability to access the lock state in the future, for example for @dannyroberts script that zips up the SQLite databases (which I don't know how that handles locking currently)?

I'm pro moving away from the distributed locking given the problems we're having and how ugly this PR became working around them.

wpride:
closing in favor of https://github.com/dimagi/formplayer/pull/522

PR #514, opened 2017-11-01
Formplayer to attempt to re-use downloaded CCZ files for install

wpride:
This is a pretty nifty refactor wherein app install requests will first check whether the CCZ has already been downloaded to the file system before attempting to download a new copy and re-use that file if so. We accomplish this by updating the `FormplayerArchiveFileRoot` class, which already handled mapping between `jr://` paths and the actual file during install, to use the versioned `app_id` (as passed by the install request URL) of the application as the archive's GUID instead of a randomly generated UUID. Then subsequent requests simply use the FileRoot to check whether a mapping is already available for that UUID in the static mapping or in Redis and re-use the file if so. If the mapping isn't present or it is but the file has been deleted we fail gracefully to the old code path. 

I think there's the potential for pretty significant speed boosts here as well as taking some load off HQ/request IO.

### DIFF
--- a/src/main/java/engine/FormplayerArchiveFileRoot.java
+++ b/src/main/java/engine/FormplayerArchiveFileRoot.java
@@ -25,8 +25,8 @@ public class FormplayerArchiveFileRoot extends ArchiveFileRoot {
     private int MAX_RECENT = 5;
 
     @Override
-    public String addArchiveFile(ZipFile zip) {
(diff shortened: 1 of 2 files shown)


wpride:
@dannyroberts https://github.com/dimagi/commcare-core/pull/669 ... I merged it because I got frustrated with conflicts 😬

PR #516, opened 2017-11-02
Case Purge Refactor

wpride:
Big boy that definitely needs the eyes of @ctsims 

Implements https://github.com/dimagi/commcare-android/pull/1803/files for Formplayer

Includes a refactor of `JdbcSqlStorageIterator` class from using a backing `List` to properly using the `ResultSet`. I think I used the backing list originally because I was concerned about the `PreparedStatement`s and `ResultSet`s being left open, but I think we can handle that by making sure the list is always completely iterated through or the `close` method is called directly. But definitely open for feedback around this. 

Now that I think about it, seems like a good chance we might also get some performance improvements from not copying all these records directly into memory


src/main/java/api/process/FormRecordProcessorHelper.java
```diff
@@ -95,32 +103,46 @@ public static void purgeCases(UserSqlSandbox sandbox) {
...
-        SqliteIndexedStorageUtility<Ledger> stockStorage = sandbox.getLedgerStorage();
-        LedgerPurgeFilter stockFilter = new LedgerPurgeFilter(stockStorage, storage);
-        removedLedgers = stockStorage.removeAll(stockFilter).size();
 
+            SqliteIndexedStorageUtility<Ledger> stockStorage = sandbox.getLedgerStorage();
+            LedgerPurgeFilter stockFilter = new LedgerPurgeFilter(stockStorage, storage);
+            removedLedgers = stockStorage.removeAll(stockFilter).size();
+            sandbox.getConnection().commit();
+        } catch (SQLException e) {
+            e.printStackTrace();
```
ctsims: This should get logged somewhere more accessible, right?

src/main/java/api/process/FormRecordProcessorHelper.java
```diff
@@ -95,32 +103,46 @@ public static void purgeCases(UserSqlSandbox sandbox) {
...
+        } finally {
+            try {
+                sandbox.getConnection().setAutoCommit(true);
+            } catch (SQLException e) {
+                e.printStackTrace();
```
ctsims: Is there a reason this error shouldn't be allowed to bubble up and kill the thread? If this doesn't succeed the sandbox is stuck in an inconsistent state where autocommit isn't enabled, which is a serious issue.

src/main/java/sandbox/JdbcSqlStorageIterator.java
```diff
@@ -1,28 +1,147 @@
...
+    public JdbcSqlStorageIterator(PreparedStatement preparedStatement,
+                                  ResultSet resultSet,
+                                  SqliteIndexedStorageUtility<T> storage,
+                                  String[] metaDataIndexSet) {
+        this.resultSet = resultSet;
+        this.storage = storage;
+        this.metaDataIndexSet = new HashSet<>(ArrayUtilities.toVector(metaDataIndexSet));
+        this.preparedStatement = preparedStatement;
+        try {
+            resultSet.next();
```
ctsims: If this returns false() the resultSet is empty, but it will never get closed, because hasMore() doesn't close a resultSet that is after the last element. 

Either this should close the resultSet if its empty, or hasMore() should close resultSets that have iterated past their last element.,

src/main/java/sandbox/UserSqlSandbox.java
```diff
@@ -172,13 +172,14 @@ public void setIndexedFixturePathBases(String fixtureName, String baseName,
...
             SqliteIndexedStorageUtility<User> userStorage = getUserStorage();
-            AbstractSqlIterator<User> iterator = userStorage.iterate();
+            JdbcSqlStorageIterator<User> iterator = userStorage.iterate();
             if (iterator.hasMore()) {
                 // should be only one user here
                 user = iterator.nextRecord();
             } else {
                 user = null;
             }
+            iterator.close();
```
ctsims: this should be in a finally.

src/main/java/services/RestoreFactory.java
```diff
@@ -299,12 +299,13 @@ private InputStream getRestoreXmlHelper(String restoreUrl, HqAuth auth) {
...
         SqliteIndexedStorageUtility<User> storage = getSqlSandbox().getUserStorage();
-        AbstractSqlIterator<User> iterator = storage.iterate();
-        //should be exactly one user
+        JdbcSqlStorageIterator<User> iterator = storage.iterate();
         if (!iterator.hasNext()) {
             return null;
         }
-        return iterator.next().getLastSyncToken();
+        String syncToken = iterator.next().getLastSyncToken();
+        iterator.close();
```
ctsims: should be in a finally.

src/main/java/services/RestoreFactory.java
```diff
@@ -299,12 +299,13 @@ private InputStream getRestoreXmlHelper(String restoreUrl, HqAuth auth) {
 
     public String getSyncToken() {
         SqliteIndexedStorageUtility<User> storage = getSqlSandbox().getUserStorage();
-        AbstractSqlIterator<User> iterator = storage.iterate();
-        //should be exactly one user
+        JdbcSqlStorageIterator<User> iterator = storage.iterate();
         if (!iterator.hasNext()) {
             return null;
```
ctsims: return path doesn't close iterator

src/main/java/util/FormplayerSentry.java
```diff
@@ -50,6 +54,7 @@
 
     public FormplayerSentry(SentryClient sentryClient) {
         this.sentryClient = sentryClient;
+        ILoggerFactory loggerFactory = getILoggerFactory();
```
ctsims: this is a local variable. Is this statement not a no-op?

wpride:
@ctsims this is cleaned up

dannyroberts:
@ctsims Looks like will addressed your comments and this is just waiting on final review from you

src/main/java/sandbox/JdbcSqlStorageIterator.java
```diff
@@ -1,28 +1,154 @@
...
+        }
     }
 
-    E getCurrentRecord() {
-        return ((E)backingList.get(index));
+    @Override
+    public int peekID() {
+        try {
+            return resultSet.getInt(DatabaseHelper.ID_COL);
+        } catch (SQLException e) {
```
ctsims: Wow, result sets... really suck. I can't believe this is a strongly typed exception that needs to be handled ~everywhere rather than an untyped exception.

src/main/java/sandbox/UserSqlSandbox.java
```diff
@@ -171,13 +171,20 @@ public void setIndexedFixturePathBases(String fixtureName, String baseName,
...
+                SqliteIndexedStorageUtility<User> userStorage = getUserStorage();
+                iterator = userStorage.iterate();
+                if (iterator.hasMore()) {
+                    // should be only one user here
+                    user = iterator.nextRecord();
+                } else {
+                    user = null;
+                }
+            } finally {
+                if (iterator != null) {
```
ctsims: Generally speaking this pattern should be

```
iterator = iterate();
try {
...
} finally {
    iterator.close();
}
```

Which is cleaner. There's no reason trying to catch errors from "userStorage.iterate();" since you can't do anything useful with the catch anyway.

PR #522, opened 2017-11-06
JVM-level locking

wpride:
Move locking off of Redis and onto the JVM since we know user requests are on one machine

Also add ability to evict locks owned by dead threads

### DIFF
--- a/src/main/java/application/AbstractBaseController.java
+++ b/src/main/java/application/AbstractBaseController.java
@@ -373,6 +373,12 @@ public abstract class AbstractBaseController {
         return new ExceptionResponseBean("User lock timed out", req.getRequestURL().toString());
     }
 
+    @ExceptionHandler({InterruptedRuntimeException.class})
+    @ResponseBody
+    public ExceptionResponseBean handleInterruptException(FormplayerHttpRequest req, Exception exception) {
+        return new ExceptionResponseBean("Request interrupted, please try again.", req.getRequestURL().toString());
+    }
+
     @ExceptionHandler(Exception.class)
     @ResponseBody
     public ExceptionResponseBean handleError(FormplayerHttpRequest req, Exception exception) {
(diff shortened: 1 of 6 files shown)


src/main/java/application/AbstractBaseController.java
```diff
@@ -373,6 +373,12 @@ public ExceptionResponseBean handleLockError(FormplayerHttpRequest req, Exceptio
         return new ExceptionResponseBean("User lock timed out", req.getRequestURL().toString());
     }
 
+    @ExceptionHandler({InterruptedRuntimeException.class})
+    @ResponseBody
+    public ExceptionResponseBean handleInterruptException(FormplayerHttpRequest req, Exception exception) {
+        return new ExceptionResponseBean("Request interrupted, please try again.", req.getRequestURL().toString());
```
ctsims: Is this going to show up to a user? If so, this isn't an OK message, it won't mean anything to them.

The user facing message should avoid using words that they won't understand in favor of something like

"An issue prevented us from processing your previous action, please try again"

src/main/java/exceptions/InterruptedRuntimeException.java
```diff
@@ -0,0 +1,10 @@
+package exceptions;
+
+/**
```
ctsims: Should have a description on how to use this exception and what it represents

src/main/java/services/FormplayerLockRegistry.java
```diff
@@ -0,0 +1,112 @@
...
+        this.lockTable = new FormplayerReentrantLock[arraySize];
+        this.lockTableLocks = new Object[arraySize];
+        for (int i = 0; i < arraySize; i++) {
+            this.lockTable[i] = new FormplayerReentrantLock();
+            this.lockTableLocks[i] = new Object();
+        }
+    }
+
+    private FormplayerReentrantLock setNewLock(Integer lockIndex) {
+        this.lockTable[lockIndex] = new FormplayerReentrantLock();
```
ctsims: Strictly speaking, this block should synchronize around the lock, in case it's called in ways you don't anticipate.

PR #523, opened 2017-11-07
Add download restore and parse restore timings

wpride:
This adds separate timings for the download and parsing of the restore file


wpride:
@dannyroberts actually maybe don't merge in favor of https://github.com/dimagi/formplayer/pull/524

PR #524, opened 2017-11-07
Fold most CaseAPIs fucntionality into RestoreFactory

wpride:
After dealing with the timing objects in https://github.com/dimagi/formplayer/pull/523 decided it might be better just to centralize this logic in the `RestoreFactory` where we already have the `CategoriesTiming` class Autowired. Downside is this makes `RestoreFactory` even more of a god class. I think the upside is worth it though; removing the whole `SyncTimings` class, removing the un-DRY calls to `recordTimings`. The CaseAPIs static functions all took a `RestoreFactory` as an argument anyways which just seems silly/un-static.

### DIFF
--- a/src/main/java/application/MenuController.java
+++ b/src/main/java/application/MenuController.java
@@ -194,6 +194,7 @@ public class MenuController extends AbstractBaseController {
 
         categoryTimingHelper.recordCategoryTiming(appInstallTiming, Constants.TimingCategories.APP_INSTALL);
         categoryTimingHelper.recordCategoryTiming(menuSession.getPurgeCasesTiming(), Constants.TimingCategories.PURGE_CASES);
+        categoryTimingHelper.recordCategoryTiming(menuSession.getParseRestoreTiming(), Constants.TimingCategories.PARSE_RESTORE);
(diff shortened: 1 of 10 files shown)


dannyroberts:
I like that it's simpler.

The only drawback is I'd been trying to make it so that for any given endpoint, a given timing is either _always_ sent (even if it's 0) or _never_ sent. I'm not sure I was doing a great job of it in practice, but the goal of that is so that averages include all the times a code path wasn't called at all, so that the average of the subtimings add up to the average total timing (again, in theory).

I think this kind of abandons that. That might be OK. The way I was doing it before wasn't sustainable. And it only matters for datadog; for sentry, it's useful to get it in the breadcrumbs once per time it's called (and not at all if it's not called). I think there's a better way to do what I wanted to do, and that's to have each end point declare which datadog subtimings they support and to have categoryTimings start with that value at 0 and then add to it during the course of the request, and then send it once at the end. That automates this guarantee rather than making the programmer think about it every time (and probably get it wrong).

In sum, I'm pro this change, but we should be clear that in the interim the datadog graph doesn't mean exactly what it used to

PR #528, opened 2017-11-07
Remove dependecies from commcare-core, exlude jmh

wpride:
@amstone326 This codifies some of the IDE changes we had to make in Gradle. Specifically

1. Exclude `jmh` directory in `commcare-core`
2. Defer to `commcare-core` library versions

@dannyroberts should clear up recurring `regex.me` dependency issue

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,12 @@ ext {
 sourceCompatibility = 1.7
 targetCompatibility = 1.7
 
+idea {
+    module {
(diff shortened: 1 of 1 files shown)


amstone326:
@wpride Question about the 2nd point-- I thought you were saying yesterday that formplayer isn't capable of inferring dependencies from core and you have to repeat them in the formplayer `build.gradle` file. How is what you did here different from that?

wpride:
@amstone326 Apparently I was wrong, 🎉

PR #530, opened 2017-11-08
Ensure autoCommit always re-enabled after restores

wpride:


ctsims:
Hey Will, 

One quick question: Is there not a concern that things might happen between "commit()" and the finally execution which may not be commited? (not sure if setting autocommit to true commits any pending stuff)

wpride:
@ctsims hmm good point. I think in this case we're fine because `commit` will set the transaction mode back to auto-commit and `setAutoCommect(true)` would commit any pending transactions. Still probably worth adding this for clarification though.

PR #531, opened 2017-11-08
Parse Restore response stream directly

wpride:
In local testing I observed that we were spending a lot of time converting the restore response from an InputStream to a resource and then back to an InputStream in [ResourceHttpMessageConverter](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/converter/ResourceHttpMessageConverter.html). In addition to the time spend I'm sure reading all this into memory wasn't good either.

This PR handles parsing the restore response (or the async retry response) directly in a new `MessageConverter` sub class


wpride:
1. BufferedStream
2. Don't blow away DBs for incremental

wpride:
@ctsims Changes you requested have been made

dannyroberts:
@wpride Looks like this has fallen behind and has conflicts that need to be resolved.

wpride:
@dannyroberts thanks, adding a Don't Merge tag until I can performance test this

PR #533, opened 2017-11-08
Thread local cache table

wpride:
This cache should probably be ThreadLocal rather than shared across all threads


dannyroberts:
Can you make this PR too? https://github.com/dimagi/commcare-core/compare/formplayer...thread-local-cache-table

wpride:
@dannyroberts https://github.com/dimagi/commcare-core/pull/674

PR #535, opened 2017-11-10
Add authToken args to MenuDebugger routes

wpride:
If the `authToken` argument isn't specified in the route then it isn't available to us in the `UserRestoreAspect`. This was breaking the MenuDebugger routes. 

Product announce: Menu debugger functional again


dannyroberts:
For posterity, how did this break? Was it fully broken or only in some edge case?

wpride:
Fully broken. I think what happened was that this route didn't require auth (since you're using a `MenuSession`) but in the past 1-2 months I made having Auth a hard requirement for all routes with the `UserRestore` annotation. That must've broken this

PR #537, opened 2017-11-14
Add distinct-values function

wpride:
Adds `distinct-values` function in core

cross-request: https://github.com/dimagi/commcare-core/pull/676

@ctsims not sure how you want to handle announce or this?


ctsims:
@wpride  Custom / invisible is totally fine.

PR #543, opened 2017-11-16
RESTful end of form navigation

wpride:
Instead of passing a `menuSessionId` between the browser and Formplayer after end of form navigation, determine the list of steps the user "would have taken" to get to the current position. Such a set of steps must always exist after end of form navigation. Then, return this set of steps to the browser to re-set the URL.

The Menu debugger previously relied on the menuSessionId; instead, re-create the menuSession for each request. 

Contains a couple of refactors: centralize the creation of `MenuSession` objects in a `MenuSessionFactory` class and centralize the logic around menu session navigation in the `MenuSessionRunnerService` class.

cross-request: https://github.com/dimagi/commcare-hq/pull/18569

**Product Note**: breadcrumbs now work after end of form navigation


src/main/java/services/MenuSessionFactory.java
```diff
@@ -0,0 +1,97 @@
...
+    private String host;
+
+    private static final Log log = LogFactory.getLog(MenuSessionFactory.class);
+
+    /**
+     * Rebuild the MenuSession from its stack frame. This is used after end of form navigation.
+     * By re-walking the frame, we establish the set of selections the user 'would' have made to get
+     * to this state without doing end of form navigation. Such a path must always exist in a valid app.
+     */
+    public void rebuildSessionFromFrame(MenuSession menuSession) {
```
wpride: @amstone326 @ctsims this function does the heavy lifting and could use some verification of the assumptions I've made

dannyroberts:
@wpride awesome that you did this! Can you run me through how we're sure (or the plan for how we can get sure) this won't break anything in production?

wpride:
@dannyroberts Thank you for asking that very important question. I agree that that's an extremely important issue and I can assure you that we're currently looking into that at a full capacity. 

I've managed to convince myself that this is safe through test coverage. We had existing tests around double management EOF navigation, case claim navigation, and one Enikshay workflow. In this PR I'd ended another test covering two Enikshay private form navigation workflows, and I just added a test from our mobile QA plan that covers five "standard" types of end of form navigation. These are all passing (and rather smoothly). What's "allowed" in end of form navigation is actually pretty limited, so I feel pretty good about our level of coverage.

The failure mode here is also just people getting sent to the home screen (and us getting a Sentry message) without any data loss

How does that sound to you?

dannyroberts:
Thanks for the explanation @wpride, sounds good to me.

In other news i think the conflict is just with the submodule

src/test/java/tests/GeoTests.java
```diff
@@ -28,6 +28,6 @@ protected String getMockRestoreFileName() {
     // validate that we don't crash on here() function
     @Test
     public void testHereOverride() throws Exception {
-        sessionNavigate(new String[]{"5", "1"}, "basic", EntityListResponse.class);
+        sessionNavigate(new String[]{"6", "1"}, "basic", EntityListResponse.class);
```
amstone326: @wpride why did this change?

src/test/java/tests/EnikshayEndOfFormTests.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import utils.TestContext;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+
+/**
+ * Regression tests for fixed behaviors
```
amstone326: Oh interesting I had thought this was purely a refactor; what issues did it fix?

src/main/java/services/MenuSessionFactory.java
```diff
@@ -0,0 +1,102 @@
...
+                    for (StackFrameStep step : steps) {
+                        if (step.getId().equals(options[i].getCommandID())) {
+                            currentStep = String.valueOf(i);
+                        }
+                    }
+                }
+            } else if (screen instanceof EntityScreen) {
+                EntityScreen entityScreen = (EntityScreen) screen;
+                SessionDatum neededDatum = entityScreen.getSession().getNeededDatum();
+                Set<String> caseIds = entityScreen.getReferenceMap().keySet();
```
ctsims: noting that "caseIds" is a misnomer here.

amstone326: @ctsims what should it be?

wpride: Also wondering this - would you prefer `entityIds` ?

src/test/java/tests/GeoTests.java
wpride: I updated to the most recent CCQA CCZ file and this module was moved

src/test/java/tests/EnikshayEndOfFormTests.java
wpride: I'm going to have to get better at my class comments with you reviewing...

src/main/java/services/MenuSessionFactory.java
ctsims: entityIds is better, yeah.

PR #544, opened 2017-11-17
Dockerfile

kaapstorm:
This PR adds a Dockerfile that pulls the latest build from Jenkins, and uses the example application properties as a default config.

The idea of compiling as part of the build is definitely up for debate. I chose this approach because I figured it would be the quickest way to get a docker image for developers.

(I'm not sure whether the couch connection is still necessary.)

cc @millerdev @ctsims

### DIFF
--- a/config/application.docker.properties
+++ b/config/application.docker.properties
@@ -1,5 +1,5 @@
 # set to the URL of your CCHQ instance
-commcarehq.host=http://__WEB_HOST__:8000
+commcarehq.host=http://webhost:8000
 commcarehq.environment=development
 commcarehq.formplayerAuthKey=secretkey
 user.suffix=commcarehq.org
--- a/config/application.docker.properties
+++ b/config/application.docker.properties
@@ -1,0 +1,40 @@
+# set to the URL of your CCHQ instance
+commcarehq.host=http://__WEB_HOST__:8000
+commcarehq.environment=development
+commcarehq.formplayerAuthKey=secretkey
+user.suffix=commcarehq.org
+logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=INFO
+server.port=8080
+redis.hostname=redis
+# sqlite.dataDir must end in /
+sqlite.dataDir=dbs/
+# sentry.dsn only needed for production environments
+sentry.dsn=
+
+# Properties for HQ db
+datasource.hq.driverClassName=org.postgresql.Driver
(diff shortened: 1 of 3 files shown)


wpride:
@kaapstorm this is awesome!

My only question is why do the new settings in `application.docker.properties` need to be in a separate file (rather than just `application.properties`0?

kaapstorm:
When users pull these changes, I don't want git to try to overwrite their existing `application.properties` file. So I reckoned I should either change the defaults in `application.example.properties` to be right for Docker, or create a new file for Docker values and leave the existing example file as it is. At first I went with changing the existing file, and then figured it would be less disruptive to make a new file.

But maybe I should go back to changing the existing example file, because when defaults need to be added or changed then we only need to do that in one place.

config/application.docker.properties
```diff
@@ -0,0 +1,47 @@
...
+user.suffix=commcarehq.org
+logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=INFO
+server.port=8080
+redis.hostname=redis
+# sqlite.dataDir must end in /
+sqlite.dataDir=dbs/
+# sentry.dsn only needed for production environments
+sentry.dsn=
+
+couch.username=commcarehq
```
snopoke: our current docker setup uses admin party for couch

kaapstorm:
@millerdev suggested using an environment variable to determine whether FormPlayer is running with HQ running locally or in another Docker container. 

If HQ is running locally (i.e. on the Docker host machine), then we need to update `/etc/hosts` with the IP address of the docker host. In Docker, `/etc/hosts` is actually a mount point, not a normal file, so we have to make changes to a temporary copy of it, and then copy the temporary file back again.

We also need to set the `application.properties` file to point to the container named "web" if HQ is also running in Docker, or to the Docker host if it is running locally.

I chose to write the `application.properties` file at build time, and set an alias in `/etc/hosts` to wherever HQ is running at runtime, considering we are changing `etc/hosts` anyway. Alternatively I could change `application.properties` at runtime too, but I figured using just one script, and limiting runtime changes to the hosts file, was a simpler approach.

I removed the Couch settings from `application.properties` because it turns out they are unused. And I'm happy to use `application.example.properties` instead of having a separate `application.docker.properties` file.

See also [HQ PR 18584](https://github.com/dimagi/commcare-hq/pull/18584) for details on how to run HQ.

kaapstorm:
cc code buddy @dannyroberts, this was half my Hackathon project.

dannyroberts:
nice!

PR #546, opened 2017-11-17
Improved Debugger Responses

ctsims:
Allow formplayer to respond to xpath eval requests with structured trace responses.

Requires:
https://github.com/dimagi/commcare-core/pull/678

### DIFF
--- a/src/main/java/beans/EvaluateXPathResponseBean.java
+++ b/src/main/java/beans/EvaluateXPathResponseBean.java
@@ -5,6 +5,10 @@ import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.AbstractTreeElement;
 import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.model.trace.AccumulatingReporter;
+import org.javarosa.core.model.trace.EvaluationTraceReporter;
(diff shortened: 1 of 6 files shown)


src/main/java/beans/EvaluateXPathResponseBean.java
```diff
@@ -31,17 +35,31 @@
...
     //Jackson requires the default constructor
     public EvaluateXPathResponseBean(){}
 
-    public EvaluateXPathResponseBean(EvaluationContext evaluationContext, String xpath) throws XPathSyntaxException {
+    public EvaluateXPathResponseBean(EvaluationContext evaluationContext, String xpath, String debugTraceLevel) throws XPathSyntaxException {
         status = Constants.ANSWER_RESPONSE_STATUS_POSITIVE;
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
         KXmlSerializer serializer = new KXmlSerializer();
 
+        EvaluationContext subcontext = evaluationContext.spawnWithCleanLifecycle();
```
wpride: Why do we need to spawn a new `subContext` here? Just being a good citizen?

ctsims: The  debugger we set after this point is set on an eval context, so we have to spin up a new context to prevent leaking anything from outside of this method/process

PR #548, opened 2017-11-17
Delete form session right after submit success

wpride:
Delete `FormSession` from Postgres immediately after getting a successful sync. Previously, we retrieved the next screen (for end of form navigation) before deleting the row. If we encountered an error here, we would have submitted the form but *not* have deleted the `FormSession`. The user would be redirected back to the home screen by the failed end of form nav, but in theory they could go to the Incomplete Forms page and re-open and re-submit that form, as appears to have happened here:

https://manage.dimagi.com/default.asp?265683

@ctsims turned this into a two-fer to include rolling back on failed submissions

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -161,9 +161,11 @@ public class FormController extends AbstractBaseController{
                 }
                 // Only delete session immediately after successful submit
                 deleteSession(submitRequestBean.getSessionId());
+                restoreFactory.commit();
 
             }
             catch (InvalidStructureException e) {
+                restoreFactory.rollback();
                 submitResponseBean.setStatus(Constants.ANSWER_RESPONSE_STATUS_NEGATIVE);
                 submitResponseBean.setNotification(new NotificationMessage(e.getMessage(), true));
                 log.error("Submission failed with structure exception " + e);
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -128,41 +128,51 @@ public class FormController extends AbstractBaseController{
             submitResponseBean.setStatus(Constants.ANSWER_RESPONSE_STATUS_NEGATIVE);
         } else {
             try {
+                restoreFactory.setAutoCommit(false);
+
                 SimpleTimer purgeCasesTimer = FormRecordProcessorHelper.processXML(
                         new FormplayerTransactionParserFactory(restoreFactory.getSqlSandbox(),
                                 PropertyUtils.isBulkPerformanceEnabled()),
                         formEntrySession.submitGetXml()
                 ).getPurgeCasesTimer();
+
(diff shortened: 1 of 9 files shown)


src/main/java/services/SubmitService.java
```diff
@@ -4,23 +4,37 @@
...
  * Service that handles form submission to CommCareHQ
  */
-public class SubmitService {
+public class SubmitService extends DefaultResponseErrorHandler {
 
     @Autowired
     RestoreFactory restoreFactory;
 
     public ResponseEntity<String> submitForm(String formXml, String submitUrl) {
+        restoreFactory.setAutoCommit(false);
```
ctsims: I think these always need to be wrapped in try{} finally{} blocks

ctsims: Er, not the setAutoCommit(False) part, the part after (including the potential rollback). The general pattern of

```
setNotAutoCommit
try{
   dostuff
} fail {
rollback
} finally {
  autocommitbackon
}
```

ctsims:
Also: Is it possible to write a unit / regression for this?

src/main/java/application/FormController.java
```diff
@@ -130,39 +128,53 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
...
+
+            }
+            catch (InvalidStructureException e) {
+                restoreFactory.rollback();
+                submitResponseBean.setStatus(Constants.ANSWER_RESPONSE_STATUS_NEGATIVE);
+                submitResponseBean.setNotification(new NotificationMessage(e.getMessage(), true));
+                log.error("Submission failed with structure exception " + e);
                 return submitResponseBean;
             }
+            finally {
```
ctsims: @wpride Currently there are only two rollback conditions: Bad responses from the server, and Invalid structure.

I _think_ it's the case that we should be ready to handle any other deviant behavior as well though, right?

The pattern that we follow in android (thanks to good library support) is actually

```
transaction()
try{
   transaction.Succeesful();
} [potential catch blocks] 
finally {
   transaction.close();
}
```

Where ```transaction.close``` _automatically_ rolls back the commit without needing any specific error handler to do ti for us. The transaction essentially starts 'dirty' and has to be marked clean by the `try{}` block in order for ending the transaction to be a commit instead of a rollback.

Here, we should either be more proactive about handling unexpected situations, or we should adopt a similar approach where a rollback is the assumption, rather than an exceptional output.

wpride:
@ctsims good call, updated

ctsims:
commit() re-enables autocommit by default, I imagine?

I don't think this should block the PR, but before adopting this specific code pattern I'd suggest we see if there's better built-in transactional support, or at the very least rename the methods to reflect the role they're playing in the transaction, ie: commit() should be called "commitAndMarkSuccesful()" or something, since we've overloaded the meaning of the autocommit mode.

dannyroberts:
@wpride is this good to go now?

wpride:
@ctsims This is just waiting on your confirming the changes you requested

PR #559, opened 2017-12-04
Revert "Purge stale form sessions"

wpride:
Reverts dimagi/formplayer#512


wpride:
"Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'formSessionRepo': Invocation of init method failed; nested exception is org.springframework.dao.DataIntegrityViolationException: StatementCallback; SQL [delete from formplayer_sessions where dateopened::timestamptz < now() - interval '7 days']; ERROR: date/time field value out of range: "Mon May 02 24:38:59 UTC 2016"; nested exception is org.postgresql.util.PSQLException: ERROR: date/time field value out of range: "Mon May 02 24:38:59 UTC 2016"

PR #561, opened 2017-12-07
Saved Forms

wpride:
For ICDS (and probably helpful on Enikshay) implement Saved Forms for Formplayer. ICDS requested this feature so that Web Apps will continue working when HQ (but not Formplayer) is down which apparently is a common situation for them. We could potentially also spin form submission into a non-blocking asyncronous thread (as we do in Android) so that on Enikshay when couch is backed up Formplayer will no longer be blocked on submission (in fact this PR should just do that).

This is a fairly big conceptual switch with some not-great consequences (namely, that forms could be "submitted" but not show up for other users) but seems like ICDS will need (already needs) this. Open for comments.


src/main/java/application/UtilController.java
```diff
@@ -38,16 +42,32 @@
...
+    private SubmitService submitService;
+
     @ApiOperation(value = "Sync the user's database with the server")
     @RequestMapping(value = Constants.URL_SYNC_DB, method = RequestMethod.POST)
     @UserLock
     @UserRestore
     public SyncDbResponseBean syncUserDb(@RequestBody SyncDbRequestBean syncRequest,
                                          @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
         restoreFactory.performTimedSync();
+        submitUserForms(restoreFactory.getUsername());
```
snopoke: shouldn't this happen prior to the restore?

ctsims:
Hey Will,

Can we sync up on this before this gets integrated (potentially with the
ICDS team members)? I'm not 100% sure I think this is something we should
accept as a technical limitation. It's a massive increase in complexity.

-Clayton

wpride:
@ctsims Yeah, sorry I should have been clearer this is definitely a POC and not something I'm certain we should merge. However, the ICDS team were pretty insistent that something along these lines would be needed given that the hardware situation doesn't look likely to improve and I imagine this would be very helpful on Enikshay recently too. Can cover in our sync.

ctsims:
Ok. Let's talk in sync. I am 100% not signed off that it makes sense to introduce an eventual consistency between our own internal servers in our own cloud this way, and we might need to sit down and figure out the implications with the larger ICDS team.

PR #562, opened 2017-12-07
Update README for FormPlayer in Docker

kaapstorm:
@wpride @dannyroberts cc @nickpell


wpride:
👍 This is great Norman, thank you1

PR #564, opened 2017-12-07
Remove migrated forms

wpride:
Almost a year's gone by, I think its time we cut the cord with CloudCare form sessions.

Once this is merged we can drop this table in PG.


dannyroberts:
Feels like a dumb question, but—What's a migrated form?

wpride:
@dannyroberts A form that came over from CloudCare 🙀

PR #570, opened 2017-12-09
Fix CLI src code importing

wpride:
This should fix the issues we see in our IDEs occasionally (though not building from the command line) where classes from these packages weren't being imported automatically. I think the imports removed here was making IntelliJ alternatively believe it should use the full or shortened path and so bugging out. Essentially we were importing from two locations. This clears that up.


dannyroberts:
@wpride assume you meant to make a PR for the submodule? Also FYI test failing

PR #572, opened 2017-12-15
Support for here() in entity screen (take 2)

amstone326:
cross-request: https://github.com/dimagi/commcare-hq/pull/18924, https://github.com/dimagi/commcare-core/pull/695

**Product Note:** Adds support for the `here()` function in formplayer, which enables the 'sort by distance from current location' feature in the case list to work on Web Apps. When `here()` is evaluated in Web Apps, it will return the location provided by the web browser.


src/main/java/services/MenuSessionRunnerService.java
```diff
@@ -371,21 +370,13 @@ private static StackFrameStep getStepToFrame(SessionWrapper session) {
         return stepToFrame;
     }
 
-    private static void setPersistentCaseTile(MenuSession menuSession, NewFormResponse formResponse) {
```
amstone326: got rid of these because they seemed pointless (one-line helpers with the same name as the functions they call)

src/main/java/beans/menus/EntityDetailListResponse.java
```diff
@@ -29,12 +27,15 @@ public EntityDetailListResponse(EntityDetailResponse entityDetailResponse) {
         this.isPersistentDetail = true;
     }
 
-    public EntityDetailListResponse(EntityScreen screen, EvaluationContext ec, TreeReference treeReference) {
+    public EntityDetailListResponse(EntityScreen screen, EvaluationContext ec, TreeReference treeReference,
+                                    MenuSession menuSession) {
+        ec.addFunctionHandler(new FormplayerHereFunctionHandler(menuSession, menuSession.getCurrentBrowserLocation()));
```
wpride: I don't love this pattern of passing the `MenuSession` all the way down the chain to add the function handler here. In all four usages the `EvaluationContext` comes from a call to `menuSession.getSessionWrapper().getEvaluationContext()` - what if we replaced these calls with a call to a new method `menuSession.getEvaluationContext()` that got the EC form the wrapper, then added the `here` handler immediately?

wpride: I don't love this pattern of passing the MenuSession all the way down the chain to add the function handler here. In all four usages the EvaluationContext comes from a call to menuSession.getSessionWrapper().getEvaluationContext() - what if we replaced these calls with a call to a new method menuSession.getEvaluationContext() that got the EC form the wrapper, then added the here handler immediately?

src/main/java/application/MenuController.java
```diff
@@ -147,7 +144,7 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
             if (detail == null) {
                 throw new RuntimeException("Tried to get details while not on a case list.");
             }
-            return new EntityDetailListResponse(detail);
+            return LocationRelevantResponseBean.setLocationNeeds(new EntityDetailListResponse(detail), menuSession);
```
wpride: This block is only for persistent case tiles, which I don't think would contain `here()`? Though I could be wrong

src/main/java/application/MenuController.java
```diff
@@ -122,7 +119,7 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
             if (detail == null) {
                 throw new RuntimeException("Could not get inline details");
             }
-            return detail;
+            return LocationRelevantResponseBean.setLocationNeeds(detail, menuSession);
```
wpride: This `LocationRelevantResponseBean` pattern seems un-DRY to me and obscure the meaning of the code (IE this looks like we're returning some location response rather than a case list response that possibly has location details). All of the responses being set here makes calls to `MenuSessionRunnerService` which in turn all call `getDetails` - could this code be centralized there?

src/main/java/application/MenuController.java
```diff
@@ -7,10 +7,7 @@
 import beans.NewFormResponse;
 import beans.NotificationMessage;
 import beans.SessionNavigationBean;
-import beans.menus.BaseResponseBean;
-import beans.menus.EntityDetailListResponse;
-import beans.menus.EntityDetailResponse;
-import beans.menus.UpdateRequestBean;
+import beans.menus.*;
```
wpride: This is slightly annoying, but Android Studio (and hence the `commcare-core` code) has been defaulting to never use `*` for imports, while IntelliJ will compress them like this. Recently I've switched IntelliJ to never compress as well to prevent unnecessary divergence in the `commcare-core` branch Can configure that in Preferences >Editor>Code Style>Java>Class count to use import with '*' set to 99 (see picture) 
screen shot 2017-12-21 at 2 28 00 pm

src/main/java/util/FormplayerHereFunctionHandler.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Override
+    public Object eval(Object[] args, EvaluationContext ec) throws XPathArityException {
+        alertOnEval();
+        if (browserLocation == null) {
+            return "";
+        }
+        String[] locationData = browserLocation.split(",");
+        double lat = Double.parseDouble(locationData[0]);
```
wpride: Seems like this could throw some `NumberFormatException`s or `ArrayIndexOutOfBoundsException`s

src/main/java/services/MenuSessionRunnerService.java
wpride: :+1: very nice, thanks

src/main/java/application/MenuController.java
amstone326: I mean it's just up to us whether we want to support it in that place. My perspective is that it would be confusing to support `here()` in some details but not others, so I'd rather be consistent.

src/main/java/application/MenuController.java
amstone326: @wpride I didn't follow the alternative you proposed, but in the most recent commit I moved where the `setLocationNeeds()` helper lives. I think that should resolve the issue you're talking about where it looks like we're returning a different type of response bean. If you want to discuss further what you were suggesting, let me know.

amstone326:
@wpride Ok I think I addressed all of your feedback on here and https://github.com/dimagi/commcare-hq/pull/18924. Thanks for the thorough review!

src/main/java/util/FormplayerHereFunctionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+        if (browserLocation == null) {
+            return "";
+        }
+        try {
+            String[] locationData = browserLocation.split(",");
+            double lat = Double.parseDouble(locationData[0]);
+            double lon = Double.parseDouble(locationData[1]);
+            return (new GeoPointData(new double[]{lat, lon})).getDisplayText();
+        } catch (NumberFormatException | IndexOutOfBoundsException e) {
+            throw new XPathTypeMismatchException("The browser location provided for evaluation of here() was of an unexpected format");
```
wpride: NIT, but would be nice to have the `locationData` in the error message

src/main/java/application/DebuggerController.java
```diff
@@ -85,12 +83,12 @@ public MenuDebuggerContentResponseBean menuDebuggerContent(
         MenuSession menuSession = getMenuSessionFromBean(debuggerMenuRequest);
         runnerService.advanceSessionWithSelections(menuSession, debuggerMenuRequest.getSelections());
 
-        return new MenuDebuggerContentResponseBean(
+        return MenuController.setLocationNeeds(new MenuDebuggerContentResponseBean(
```
wpride: Should have thought of this on my previous review, but I don't think the front end component of this will work for the debugger routes. They are handled in a separate `js` file outside of Marionette [here](https://github.com/dimagi/commcare-hq/blob/d07d1209f938dac9dbbd004cc800c02df89d23af/corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js#L458-L467) so I don't think the callback hooks will work. IMO I don't think we need that working as a first pass

src/main/java/application/MenuController.java
```diff
@@ -156,10 +141,9 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
             throw new RuntimeException("Could not find case with ID " + detailSelection);
         }
 
-        return new EntityDetailListResponse(
-                entityScreen,
-                menuSession.getSessionWrapper().getEvaluationContext(),
-                reference
+        return setLocationNeeds(
```
wpride: What I was trying to say here was that it would be nice to move these `setLocationNeeds` calls from being called at the end of [getDetails](https://github.com/dimagi/formplayer/blob/0a3d04e32cad9cf195c828673235764c091cced9/src/main/java/services/MenuSessionRunnerService.java#L318-L318) and [advanceSessionWIthSelections](https://github.com/dimagi/formplayer/blob/0a3d04e32cad9cf195c828673235764c091cced9/src/main/java/services/MenuSessionRunnerService.java#L147-L147) rather than after these methods return. However, I'm now noticing that the call paths for these four usages are all slightly different in a way that would make this not clean... so maybe worth worrying about in a future refactor of this (messy) code

PR #573, opened 2017-12-19
Add blocked time to request increment

wpride:
Honestly I don't really understand the Datadog side of this well enough to understand when we want the duration vs. the duration bucket - if we could offline about that I'd appreciate it.


dannyroberts:
Tags (which is what all of the things you're changing are) in datadog are meant to have discrete values, and the way they're useful for building graphs reflects that: you can filter for e.g. "domain:chadbrochill" but not for *"duration_ms>10000"; you can do breakdown by tag, so like stacked bars where each segment is a different tag value, again something that doesn't make sense when there are continuous tag values.

In addition to being discrete, it's also nice _pricing_-wise to have a _small_ number of values. For example, request url is discrete (each value is distinct with no continuous spectrum between them the way there is for numbers) but there are many, many values; domains are another example—discrete but with hundreds of possible values. So those _make sense_ as tag values but are expensive ($/mo). Something like `restore_blocked_time` on the other hand isn't just expensive, it also _doesn't make sense_ as a tag value, because you never want to filter on "restore_blocked_time:2542" or something, so putting it there is both expensive and totally useless.

The place where putting continuous values (like time) _does_ make sense is as the "value" of a metric. Reporting that as its own timing makes total sense.

So why do we use timing buckets? Timing buckets turn a continuous quantity (ms) into one of 6 values. This in turn _is_ useful to filter on and (even more so) to do breakdown by. Does that make sense?

In this PR, you're going to want to turn all of those timings that you're adding as tags into buckets. _Or_ you need to submit a different metric for each of those. _Or_ (any maybe this is the best for what you want to do with them?) you want to submit a new metric that's organized like the categorical timings metric, where you report a the timing as the value and what it's measuring (e.g. `"unblocked_time"`) as a tag (e.g. `"category:unblocked_time"`).

wpride:
@dannyroberts Thanks! Updated

PR #575, opened 2018-01-02
Localize to browser timezone

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/709, https://github.com/dimagi/commcare-hq/pull/19045
  
Requests to formplayer from web apps now include a `tz_offset_millis` argument that communicates the timezone offset in milliseconds of the browser that sent the request. This value gets stored in the `FormSession` and is used in 2 distinct places:

1. When a request is made to answer a question, the timezone offset is passed down the chain to `JsonActionUtils.getAnswerData()` so that if we are resolving an `UncastData` value to a `TimeData` or `DateTimeData` value, we can take the timezone offset into account when doing so.

2. Whenever `FormSession.getInstanceXml()` is called, the timezone offset is passed down the chain to the various implementations of `XFormAnswerDataSerializer.serializeAnswerData()` so that if we are writing out a `DateData`, `DateTimeData` or `TimeData` value, we can take the timezone offset into account when doing so.

I added tests for the changes I made in `DateUtils` to both parsing dates and formatting dates with the timezone offset taken into account.


amstone326:
Closing in favor of a new strategy

PR #577, opened 2018-01-08
Align MenuDisplayable with Android

wpride:
Another simple refactor to get `formplayer` aligned with `commcare-core` `master`. 

In this case the `Observable` pattern for badge text doesn't really make sense for Formplayer, so we could either:

1. Use the extant `MenuDisplayable` function to get the Observable, then call the `blockingGet` function from the observable to obtain the string
2. Add another function to `MenuDisplayable` that gives the String back directly

I opted for (1) but don't feel strongly 

cross-request: https://github.com/dimagi/commcare-core/pull/723

### DIFF
--- a/src/main/java/sandbox/SqlHelper.java
+++ b/src/main/java/sandbox/SqlHelper.java
@@ -259,11 +259,9 @@ public class SqlHelper {
         }
     }
 
-    public static void basicInsert(Connection c, String storageKey,
-                                   Map<String, String> contentVals) {
+    private static void performInsert(Connection c,
+                                     Pair<List<String>, String> valsAndInsertStatement) {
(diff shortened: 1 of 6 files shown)


amstone326:
@wpride I'm confused by the PR description here. It looks like everything that's changed in the core PR is to make things the same as they are on master, but your explanation makes it sound like you had to make a decision to do something differently for formplayer. What am I missing?

wpride:
@amstone326 Sorry, poor description - you're absolutely right that this is just aligning Formplayer back with `master`. I had separated getting the raw text and getting the async object in `formplayer` and my description was elaborating on why I decided to switch back to using the one method.

PR #581, opened 2018-01-08
Revert demo user restores on Formplayer

wpride:
@amstone326 remembered why I had removed this to begin with 😨 the demo user restores are too large to store in SQLite and so crash during string serialization. Added a comment about this in the PR, but in order to implement demo user restore we'll need to implement some version of Android's `FileSystemInstaller` on Formplayer


wpride:
Created a PR with test demonstrating the failure here https://github.com/dimagi/formplayer/pull/582

PR #583, opened 2018-01-09
Fix App Fixture Storage lookup

wpride:
fixes https://manage.dimagi.com/default.asp?267173#1442608

One of those ones where I'm not sure how this wasn't broken previously. Essentially, we install app fixture to the application database, but here we were only looking in the user database for the app fixture (where it was not). This allows you to override the app fixture location when performing initialization.
  
cross-request: https://github.com/dimagi/commcare-core/pull/714


amstone326:
@wpride build failure

PR #584, opened 2018-01-09
Fix for FixtureIndex update failing

wpride:
Noticed this in Sentry, looks like this statement should be an insert-or-replace rather than a basic insert. Added a line to the tests that would trigger this failure.

### DIFF
--- a/src/main/java/sandbox/SqlHelper.java
+++ b/src/main/java/sandbox/SqlHelper.java
@@ -284,10 +284,37 @@ public class SqlHelper {
         }
     }
 
+    public static void insertOrReplace(Connection c,
+                                       String storageKey,
+                                       Map<String, String> contentValues) {
+        PreparedStatement preparedStatement = null;
+        Pair<List<String>, String> valsAndInsertStatement =
+                buildInsertOrReplaceStatement(storageKey, contentValues);
+        try {
+            preparedStatement = c.prepareStatement(valsAndInsertStatement.second);
+            int i = 1;
+            for (String val : valsAndInsertStatement.first) {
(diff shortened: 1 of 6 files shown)


src/test/java/tests/Enikshay2bTests.java
```diff
@@ -10,6 +10,7 @@
 import org.json.JSONObject;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.Mockito;
```
amstone326: unused

src/main/java/sandbox/SqlHelper.java
```diff
@@ -284,10 +284,37 @@ public static void basicInsert(Connection c, String storageKey,
         }
     }
 
+    public static void insertOrReplace(Connection c,
```
amstone326: Everything in this function is identical to `basicInsert()` besides calling `buildInsertOrReplaceStatement` instead of `buildInsertStatement()`. Would be nice to unify these and just have a boolean arg that controls which of those helpers gets used

wpride:
@amstone326 Thanks!

src/main/java/sandbox/SqlHelper.java
```diff
@@ -259,11 +259,9 @@ public static PreparedStatement prepareTableSelectStatement(Connection c,
         }
     }
 
-    public static void basicInsert(Connection c, String storageKey,
-                                   Map<String, String> contentVals) {
+    public static void performInsert(Connection c,
```
amstone326: can be `private`

PR #585, opened 2018-01-09
Localize date calcs to browser timezone

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/715, https://github.com/dimagi/commcare-hq/pull/19045
  
 **Product Note**: Fixes an issue where all date, time, and datetime computations in Web Apps could only be conducted relative to the timezone of the formplayer server. Now, any evaluations of dates/times/datetimes will take into account the timezone of the browser that submitted the request


src/main/java/services/BrowserValuesProvider.java
```diff
@@ -0,0 +1,22 @@
+package services;
+
+import beans.AuthenticatedRequestBean;
+import org.javarosa.core.model.utils.TimezoneProvider;
+
+/**
+ * Created by amstone326 on 1/8/18.
+ */
+public class BrowserValuesProvider extends TimezoneProvider {
```
amstone326: The reason I gave this a more generic name is because once this PR is merged, I think it makes sense to refactor the work I did for sending the browser location to formplayer to use this same pattern, rather than passing it down the chain in a million places.

src/main/java/aspects/SetBrowserValuesAspect.java
```diff
@@ -0,0 +1,32 @@
...
+    @Autowired
+    private BrowserValuesProvider browserValuesProvider;
+
+    @Before(value = "@annotation(org.springframework.web.bind.annotation.RequestMapping)")
+    public void setValues(JoinPoint joinPoint) throws Throwable {
+        Object beanArg = joinPoint.getArgs()[0];
+        if (beanArg instanceof AuthenticatedRequestBean) {
+            browserValuesProvider.setTimezoneOffset((AuthenticatedRequestBean)beanArg);
+        }
+        DateUtils.setTimezoneProvider(browserValuesProvider);
```
wpride: What happens if there isn't a value for `tz_offset_millis`?

src/main/java/aspects/SetBrowserValuesAspect.java
```diff
@@ -0,0 +1,32 @@
...
+public class SetBrowserValuesAspect {
+
+    @Autowired
+    private BrowserValuesProvider browserValuesProvider;
+
+    @Before(value = "@annotation(org.springframework.web.bind.annotation.RequestMapping)")
+    public void setValues(JoinPoint joinPoint) throws Throwable {
+        Object beanArg = joinPoint.getArgs()[0];
+        if (beanArg instanceof AuthenticatedRequestBean) {
+            browserValuesProvider.setTimezoneOffset((AuthenticatedRequestBean)beanArg);
```
wpride: If this condition isn't met looks like we'll still set the timezone provider but with the default timezone offset of `timezoneOffsetMillis`? I'd think if this condition isn't met we should log the event and not set the timezone provider?

src/main/java/aspects/SetBrowserValuesAspect.java
```diff
@@ -0,0 +1,32 @@
...
+@Aspect
+@Order(1)
+public class SetBrowserValuesAspect {
+
+    @Autowired
+    private BrowserValuesProvider browserValuesProvider;
+
+    @Before(value = "@annotation(org.springframework.web.bind.annotation.RequestMapping)")
+    public void setValues(JoinPoint joinPoint) throws Throwable {
+        Object beanArg = joinPoint.getArgs()[0];
```
wpride: I believe this will crash with `ArrayIndexOutOfBoundsException` when the request doesn't have a body, should check as in https://github.com/dimagi/formplayer/blob/17dc83b30b8ab72c480eaf239f2999c68b20fc91/src/main/java/aspects/MetricsAspect.java#L51-L51

src/main/java/aspects/SetBrowserValuesAspect.java
```diff
@@ -0,0 +1,32 @@
...
+import org.javarosa.core.model.utils.DateUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.annotation.Order;
+import services.BrowserValuesProvider;
+
+/**
+ * Created by amstone326 on 1/8/18.
+ */
+@Aspect
+@Order(1)
```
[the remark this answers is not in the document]
amstone326: Ah I'm sorry you're right, I wasn't reading closely enough when I changed this

src/main/java/aspects/SetBrowserValuesAspect.java
amstone326: We need to set the provider every time because otherwise a request with a bean of some other type would end up using whatever `BrowserValuesProvider` had last been set as the provider in `DateUtils` by a previous request. But it's harmless to do so because the default value of `timezoneOffsetMillis in `BrowserValuesProvider` is -1, which is the condition under which we ignore the timezone offset and don't use it in `DateUtils`

src/main/java/aspects/SetBrowserValuesAspect.java
amstone326: The default value of `timezoneOffsetMillis` in `AuthenticatedRequestBean` is also -1, so if the request body never sets a different value, then the call to `bean.getTzOffset()` in `BrowserValuesProvider.setTimezoneOffset()` will return -1, thus resulting in the same situation I described in the above question.

src/main/java/aspects/SetBrowserValuesAspect.java
amstone326: thanks

amstone326:
@wpride Thanks for the review and bug catches. I've either addressed or responded to all your comments on this and the 2 supporting PRs.

PR #589, opened 2018-01-12
Update Formplayer to use non-singleton PropertyManager

wpride:
Implementation of changes in https://github.com/dimagi/commcare-android/pull/1928 for Formplayer. Changes the `PropertyManager` to use the `FormplayerStorageFactory` as the delivery vector since the lifecycle of both is the same (IE an app install)


amstone326:
@wpride I'm a little confused by this PR. It seems like now there are 2 separate `PropertyManager`s in play that can be provided by 2 sources: 1 from the `CommCarePlatform` and 1 from the `FormplayerStorageFactory`. Why is that necessary/correct?

wpride:
@amstone326 Good question, the three places I use the `PropertyManager` from `FormplayerStorageFactory` did not have the `CommCarePlatform` object readily available; in `RestoreFactory` in particular there's not guarantee that an application will even be associated with the request (IE we could just be doing a `sync-db` request) so I needed a new vector. Given that the `PropertyManager` class is essentially a wrapper around the storage layer and lives for the lifecycle of the application, the `FormplayerStorageFactory` class seemed like a logical fit. Note that we can get away with instantiating different `PropertyManager` classes because they're all going to read from the same underlying SQLite DB for their values so nothing will be stored at Java-class level. 

I agree that this isn't ideal cleanliness/clarity but I have another [PR](https://github.com/dimagi/formplayer/pull/591) that will change how we use the `StorageManager` and will clean this up.

amstone326:
Ah also this has submodule conflicts now

wpride:
@amstone326 thanks, updated

PR #593, opened 2018-01-17
Check for sync frequency null

wpride:
If sync frequency returns null, default to not requiring a sync

fixes https://manage.dimagi.com/default.asp?268659

Thanks @kaapstorm


wpride:
Minor change + tests passing locally, merging to hotfix due to build backlog

PR #600, opened 2018-01-23
Attempt to parse message in submission response and send to browser

wpride:
Parse the `OpenRosaResponse` message block from submissions (currently always a `√`) and send to the browser

@orangejenny this ends up in the extant notifications JSON attribute (which we use for pop-ups), if you don't want to re-purpose this for this case we can add a new attribute:

`{..."notification":{"message":"   √   ","error":false}, ...}`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,6 +77,7 @@ dependencies {
     compile fileTree(dir: 'libs', include: '*.jar')
     compile project(path: ':commcare', configuration: 'api')
     compile project(path: ':commcare', configuration: 'translate')
+    compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.9.0'
     testCompile project(path: ':commcare', configuration: 'api')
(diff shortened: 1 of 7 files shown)


orangejenny:
I'll leave it to Aliza to actually review, but this is excellent!

amstone326:
@wpride What would happen if there was a notification to be sent and 1 of these messages at the same time?

wpride:
@amstone326 Good question that I thought about for a minute. Currently, the only notification sent would be the last one shown. However, for the submission endpoint the only time we're using `setNotification` is when there's a blocking error so that wouldn't conflict with this. So, I decided YAGNI when thinking about building something more elaborate/robust

amstone326:
@wpride hm, I'm still not thrilled about that. If someone comes along later and adds another notification getting set at a timing that does conflict, they'll have no idea that they're interfering with behavior that HQ is relying upon, right?

wpride:
@amstone326 OK you convinced me, moved this to its own location (and adde some test coverage)

PR #606, opened 2018-01-30
CommCareConfigEngine use PrintStream, Formplayer passes in dummy

wpride:
cross-request: https://github.com/dimagi/commcare-core/pull/742

Pass in a dummy `PrintStream` to `CommCareConfigEngine` since we do not care about the (mostly diagnostic) logging done there. Hard failures should be sent as Exceptions and logged at the Sentry level.


ctsims:
Hey,

So the method you overwrote in outputstream isn't the only method in an outputstream and I think it's dangerous to build no-op objects which don't account for the full behavior of the underlying classes.

Is there a meaningful disadvantage to printing this stuff out to stdout? If so, I think we should make a first-class object which no-ops an output stream (or find one online), which has full tests to make sure nothing akward happens when it is used

wpride:
@ctsims fair enough, updated.

The only "disadvantage" was that I don't find the `CommCareConfigEngine` log output that useful, but happy to address that separately

PR #607, opened 2018-01-30
SMS for Formplayer

wpride:
Updates to Formplayer to handle SMS requests

A few large changes:

1. **Authentication** Touchforms interacts directly with Formplayer on behalf of devices that often have no mobile worker associated with them. Previously, the Jython server used the credentials special `TOUCHFORMS_USER` CommCareHQ superuser to make requests to HQ. We copy that model here, so Formplayer will now require the `touchforms.username` and `touchforms.password` properties. Further, Formplayer will make use of the existing `formplayer_auth_key` shared with HQ to validate that a request came from HQ. 

2. **Prompt Navigation Mode** SMS uses a separate form entry mode where instead of receiving the entire set of questions all at once as in Web Apps you receive them one at a time similar to OQPS. Additionally, each answer of a question *automatically advances* to the next question and returns that question to the user. This implements that mode in the `FormSession`

3. **Restore as Case** SMS allows users to fill out a form *as a case* so that you do not need an associated mobile worker to perform form entry. Touchforms used a custom end point to fetch all the cases related to the relevant case and create an abstracted CaseDB. This PR uses a similar (but new) endpoint written by @esoergel that allows you to get a proper restore payload given a CaseDB and then use Formplayer as usual. 

This set of changes **should not** change the behavior of Web Apps. My current plan, once this is merged, is to set up an experiment to compare the results of Formplayer and Touchforms when handling the same inputs. 

cc @gcapalbo


dannyroberts:
This is awesome!

PR #609, opened 2018-01-31
Update submodule again

wpride:
Remove more unused functions towards `commcare-core` parity


amstone326:
@wpride I don't see a PR for the core change. Can you make one or point to where it is if I'm missing it?

wpride:
@amstone326 I pushed them directly to `formplayer` in my haste, you can view the diff here https://github.com/dimagi/commcare-core/compare/41a371b16614997a6163695463b28662c22715d1...065507cf92ae978def16d736d380d17815342a35

dannyroberts:
In addition to removing some methods, you also changed the signature of `getNonMainInstances`. Is that intentional?

wpride:
@dannyroberts Nice catch, and yes - that's the signature in `master`

PR #616, opened 2018-02-15
Add caseId property to FormSession to enable filling out a form as a case in SMS

wpride:
And use this caseId to restore on subsequent requests.

The `FormSession` class has become overloaded, eventually I'd like to split the functionality into `SMSFormSession`, `OQPSFormSession`, and standard `FormSession` classes

cc @gcapalbo

### DIFF
--- a/src/main/java/aspects/UserRestoreAspect.java
+++ b/src/main/java/aspects/UserRestoreAspect.java
@@ -82,7 +82,12 @@ public class UserRestoreAspect {
         } else {
             // SMS users don't submit username and domain with each request, so obtain from session
             SerializableFormSession formSession = formSessionRepo.findOneWrapped(requestBean.getSessionId());
-            restoreFactory.configure(touchformsUsername, formSession.getDomain(), formSession.getAsUser(), auth);
+
+            if (formSession.getCaseId() != null) {
+                restoreFactory.configure(formSession.getDomain(), formSession.getCaseId(), auth);
+            } else {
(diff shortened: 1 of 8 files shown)


gcapalbo:
@wpride it looks like any time a case id is present you're going to use the "restore as case" endpoint instead of the "restore as user" endpoint, is that right?

if so, then i think that's going to present an issue because we might pass case id when it's a user filling out a survey for a case, in which case we'd ideally want that user to be able to have the same normal experience as they would over webapps and restore as themselves.

i think i can add a param when we start the session named `restoreAsCaseId`, which will be the case id of the case to restore as, only when it's a case filling out the survey. this way, you'd only have `restoreAsCaseId` or `restoreAsUser` (but never both) and you can determine from that which endpoint to use.

would that work for you?

wpride:
@gcapalbo In Formplayer (and I believe in Touchforms since this was modeled on Touchforms) any `caseIds` that are part of the session data would be nested in the `session-data` parameter dictionary, not at the top-level of the request: 

https://github.com/dimagi/formplayer/blob/810f1357e3ac6db1eadafb193dbff5a378867bbf/src/main/java/beans/NewSessionRequestBean.java#L62-L62

So I wouldn't expect this to be an issue (good to confirm for SMS though, and good call)

wpride:
@gcapalbo Disregard above, realized you're totally right due to the change here https://github.com/dimagi/touchforms/pull/292/files#r168726955

wpride:
@gcapalbo Pushed the formplayer change, lmk when you have a chance to implement this and I'll test it out

PR #622, opened 2018-02-20
ReferenceManagerHandler refactor

wpride:
cross-request: https://github.com/dimagi/commcare-core/pull/761

implements static/thread-local shared pattern for `ReferenceManager`


ctsims:
I'm not sure I agree with the naming conventions. If, say, someone was using an external codebase relying on CommCare this would drop things out from under them for naming reasons, which seems like an antipattern

Given that the ReferenceManager is _always_ accessed from the singleton context, I'd just rename ReferenceManager to something different entirely (ReferenceHandler seems fine...) so that code that was working before keeps working seemlessly.

PR #623, opened 2018-02-20
Impement FormplayerSuiteParser to override demo user restore

wpride:
Enable Formplayer to skip user-restore parsing during app install

- override `getSuiteParser` to return `FormplayerSuiteParser`
- override `handleTag` in `FormplayerSuiteParser` to skip `user-restore`

cross-request: https://github.com/dimagi/commcare-core/pull/762

### DIFF
--- a/src/main/java/installers/FormplayerSuiteInstaller.java
+++ b/src/main/java/installers/FormplayerSuiteInstaller.java
@@ -1,17 +1,21 @@
 package installers;
 
+import org.commcare.resources.model.ResourceTable;
 import org.commcare.resources.model.installers.SuiteInstaller;
 import org.commcare.suite.model.Suite;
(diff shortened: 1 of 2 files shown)


ctsims:
Hey Will,

So it looks like the write to the DB actually happens at *install* time, using this installer:

https://github.com/dimagi/commcare-core/blob/76a7f08a8118f265c5975e79229ab069a366bee1/src/main/java/org/commcare/resources/model/installers/OfflineUserRestoreInstaller.java

I think it would be easier / more fitting with the existing modularity to provide a different platform installer for that action in formplayer, since it's already compartmentalized here:

https://github.com/dimagi/commcare-core/pull/762/files

That means you'd just need to override that installer with some sort of "NoOpInstaller" which just returns true, rather than changing the inheritance structure for the parsing.

PR #624, opened 2018-02-20
Answer response isn't always positive, dummy

wpride:
I think I messed this up in my SMS PR - when you get the *next* answer in prompt mode you know that the previous answer was `accepted`, but not for all requests (duh) 

addresses https://manage.dimagi.com/default.asp?270677

### DIFF
--- a/src/main/java/session/FormSession.java
+++ b/src/main/java/session/FormSession.java
@@ -513,6 +513,10 @@ public class FormSession {
         return responseBean;
     }
 
+    /**
+     *  Automatically advance to the next question after an answer when in "prompt" mode
+     *  (currently only used by SMS)
+     */
     public FormEntryNavigationResponseBean getNextFormNavigation() throws IOException {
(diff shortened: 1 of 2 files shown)


src/main/java/session/FormSession.java
```diff
@@ -526,6 +526,7 @@ public FormEntryNavigationResponseBean getNextFormNavigation() throws IOExceptio
         responseBean.setCurrentIndex(currentIndex);
         responseBean.setInstanceXml(null);
         responseBean.setTree(null);
+        responseBean.setStatus(Constants.ANSWER_RESPONSE_STATUS_POSITIVE);
```
amstone326: Is this method only used for SMS? Otherwise I'm a little confused why this would get set here. If you can add a bit of context on what this method does in general, that would probably help (or @dannyroberts if you have enough context on this from reviewing the SMS PR that feel free to ignore me and merge).

wpride:
@amstone326 Added comment, the short answer is yes this is only used by SMS

PR #625, opened 2018-02-20
Skip user restores in OfflineUserRestoreInstaller

wpride:
cross-request: https://github.com/dimagi/commcare-core/pull/763

Skip demo user restore parsing in `FormplayerOfflineUserRestoreInstaller`

@ctsims only returning `true` from install resulted in an infinite loop [here](https://github.com/dimagi/commcare-core/blob/a1a5d09e87985a85f0c208007e49b9dc47c2c35a/src/main/java/org/commcare/resources/model/ResourceTable.java#L366-L366) - parsing in a dummy `OfflineUserRestore` and then skipping initialization gives the desired result. Please let me know if I'm missing something

### DIFF
--- a/src/main/java/installers/FormplayerOfflineUserRestoreInstaller.java
+++ b/src/main/java/installers/FormplayerOfflineUserRestoreInstaller.java
@@ -1,12 +1,20 @@
 package installers;
 
+import org.commcare.resources.model.Resource;
+import org.commcare.resources.model.ResourceLocation;
+import org.commcare.resources.model.ResourceTable;
+import org.commcare.resources.model.UnresolvedResourceException;
 import org.commcare.resources.model.installers.OfflineUserRestoreInstaller;
 import org.commcare.suite.model.OfflineUserRestore;
 import org.commcare.util.CommCarePlatform;
+import org.javarosa.core.reference.Reference;
 import org.javarosa.core.services.storage.IStorageUtilityIndexed;
 import org.javarosa.core.util.externalizable.DeserializationException;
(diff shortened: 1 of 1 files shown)


src/main/java/installers/FormplayerOfflineUserRestoreInstaller.java
```diff
@@ -46,6 +59,22 @@ public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOExcep
 
     }
 
+    @Override
+    public boolean install(Resource r, ResourceLocation location,
+                           Reference ref, ResourceTable table,
+                           CommCarePlatform platform, boolean upgrade)
+            throws UnresolvedResourceException, UnfullfilledRequirementsException {
+        OfflineUserRestore offlineUserRestore = new OfflineUserRestore();
```
amstone326: Am I following that this just creates an empty/dummy user restore object? That seems fine, but what is the reason that we can't just have this method do nothing?

ctsims:
@wpride 

You can't just return true, you need to write the resource status as installed, IE:
```
+        if (upgrade) {
+            table.commit(r, Resource.RESOURCE_STATUS_INSTALLED);
+        } else {
+            table.commit(r, Resource.RESOURCE_STATUS_UPGRADE);
+        }

```

But you shouldn't need to do the parse at all

ctsims:
hey will,

this isn't tagged as ready for review again, is it? It looks fine to me now. (consider this approval if it's done, just didn't want to mark it as approved if there were still changes coming)

PR #627, opened 2018-02-21
Use ReferenceManager.instance() instead of ReferenceHandler

wpride:
For consistency reasons, switch to using `ReferenceManager.instance()` (as in `commcare-android`) rather than `ReferenceHandler`

cross-request: https://github.com/dimagi/commcare-core/pull/766


ctsims:
Thanks, this will make it way easier to do basic stuff (like copy/paste code between environments and have it be clear that it's the same thing)

PR #631, opened 2018-03-07
Add `next` and `prev` endpoints

wpride:
These endpoints are named differently for SMS, add the new namings

cc @gcapalbo

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -314,7 +314,7 @@ public class FormController extends AbstractBaseController{
     }
 
     @ApiOperation(value = "Get the questions for the next index in OQPS mode")
-    @RequestMapping(value = Constants.URL_NEXT_INDEX, method = RequestMethod.POST)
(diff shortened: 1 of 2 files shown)


src/main/java/application/FormController.java
```diff
@@ -337,7 +337,7 @@ public FormEntryNavigationResponseBean getNext(@RequestBody SessionRequestBean r
     }
 
     @ApiOperation(value = "Get the questios for the previous index in OQPS mode")
-    @RequestMapping(value = Constants.URL_PREV_INDEX, method = RequestMethod.POST)
+    @RequestMapping(value = {Constants.URL_PREV_INDEX, Constants.URL_PREV}, method = RequestMethod.POST)
```
gcapalbo: i don't think prev is actually used, did you see that being used anywhere?

gcapalbo:
@wpride is this just temporary while both formplayer and touchforms are being used for smsforms? i think once we're exclusively on formplayer we could just update the client side to use the right name

wpride:
@gcapalbo Yes temporary for now, but unfortunately SMS and Live Preview expect the responses in slightly different formats fat the moment so we need two different routes (encoded in new PR)

PR #632, opened 2018-03-07
Fix SMS next behavior

wpride:
Follow-up to https://github.com/dimagi/formplayer/pull/631

Does a few things:

- Remove `prev` route 
- Split out SMS and Live Preview `next` functionality since response format is different
- Add test

cc @gcapalbo

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -1,7 +1,6 @@
 package application;
 
-import annotations.AppInstall;
-import annotations.ConfigureStorage;
+import annotations.ConfigureStorageFromSession;
 import annotations.UserLock;
 import annotations.UserRestore;
 import api.json.JsonActionUtils;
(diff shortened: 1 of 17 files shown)


src/main/java/application/FormController.java
```diff
@@ -336,8 +336,30 @@ public FormEntryNavigationResponseBean getNext(@RequestBody SessionRequestBean r
...
+    @UserLock
+    @UserRestore
+    public FormEntryNavigationResponseBean getNextSms(@RequestBody SessionRequestBean requestBean,
+                                                   @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
+        SerializableFormSession serializableFormSession = formSessionRepo.findOneWrapped(requestBean.getSessionId());
+        storageFactory.configure(serializableFormSession.getUsername(),
+                serializableFormSession.getDomain(),
+                serializableFormSession.getAppId(),
+                serializableFormSession.getAsUser()
+        );
```
gcapalbo: confirming that all this works when it's a case filling out the survey as well as when it's a user filling it out?

wpride:
@gcapalbo Good question - because SMS passes the `form-xml` in directly we never actually perform an install so don't use the `StorageFactory`. However, this was resulting in bunch of `unknown` directories so I've cleaned this up and centralized the logic in an aspect

wpride:
@amstone326 Yes, once I realized that I'd need to add [this conditional logic](https://github.com/dimagi/formplayer/blob/ac9ffb032f0bdef0e430458ef5170651f2376e24/src/main/java/services/FormplayerStorageFactory.java#L48-L48) for the restore as case option I was confronted with either copying that to all 11 places, or respecting human decency and centralizing the call in one aspect.

PR #640, opened 2018-03-22
Allow POST requests to get-instance as well

wpride:
Fixes https://sentry.io/dimagi/commcarehq/issues/492026734/?referrer=slack

Seems like technically this should be a `GET` request but for the sake of the SMS migration accept `POST` too


gcapalbo:
@wpride if it was only accepting GET requests before does anything need to change with where it looks for its params now (like session id)?

gcapalbo:
also it looks like the client side used with touchforms expects the response to be a json result that looks like `{'output': '...'}` with `...` filled by the xml. is that what this is returning?

PR #644, opened 2018-04-16
App updates for Formplayer

wpride:
Allow Formplayer to update an application instead of always deleting DBs and re-installing. PR appears strange because a lot of this code was merge but unused already; majority of code changes are error handling. 

cross-request: https://github.com/dimagi/commcare-core/pull/785

### DIFF
--- a/src/main/java/session/MenuSession.java
+++ b/src/main/java/session/MenuSession.java
@@ -1,5 +1,6 @@
 package session;
 
+import beans.NotificationMessage;
 import engine.FormplayerConfigEngine;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -7,6 +8,7 @@ import org.apache.http.client.utils.URIBuilder;
 import org.commcare.modern.database.TableBuilder;
 import org.commcare.modern.session.SessionWrapper;
 import org.commcare.modern.util.Pair;
+import org.commcare.resources.model.UnresolvedResourceException;
 import org.commcare.session.CommCareSession;
 import org.commcare.session.SessionFrame;
 import org.commcare.suite.model.FormIdDatum;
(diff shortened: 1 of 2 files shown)


src/main/java/session/MenuSession.java
```diff
@@ -127,8 +130,26 @@ private void initializeBreadcrumbs() {
...
+                return new NotificationMessage("Application updated successfully.", false);
+            } else {
+                return new NotificationMessage("Application up to date.", false);
+            }
+        } catch (UnresolvedResourceException e) {
+            String message = "Update Failed! Couldn't find or install one of the remote resources";
+            log.error(message, e);
+            return new NotificationMessage(message, true);
+        } catch (UnfullfilledRequirementsException e) {
+            String message = "Update Failed! This CLI host is incompatible with the app";
```
amstone326: This seems like a strange message to show in formplayer; I don't think we want to mention the CLI here

src/main/java/session/MenuSession.java
```diff
@@ -127,8 +130,26 @@ private void initializeBreadcrumbs() {
...
+            log.error(message, e);
+            return new NotificationMessage(message, true);
+        } catch (Exception e) {
+            String message = "Update Failed! There is a problem with one of the resources";
```
amstone326: Should this message be more generic, given that we really have no idea where this Exception came from if we get to this point?

src/main/java/session/MenuSession.java
wpride: Good call

src/main/java/session/MenuSession.java
wpride: Good point; I copied this code and it looks like this catch was needlessly generic, the only typed exception remaining was `InstallCancelledException` which I'm now catching explicitly

amstone326:
@wpride Approved but the build seems to be failing on both this and the `core` PR

PR #645, opened 2018-04-19
Build target default to "save"

wpride:


src/main/java/beans/menus/UpdateRequestBean.java
```diff
@@ -23,16 +23,12 @@
     private String sessionId;
 
     public String getUpdateMode() {
-        return updateMode;
+        // Default to 'save' since this is only accessible from App Preview
+        return updateMode == null ? "save" : updateMode;
     }
 
     public void setUpdateMode (String updateMode) {
-        if(!VALID_UPDATE_MODES.contains(updateMode)) {
```
amstone326: why do we want to get rid of the check for whether the passed-in value is valid?

PR #651, opened 2018-05-07
Add route for changing language of form session

wpride:
For easier debugging in App Preview. Not wired into anything in the front end at the moment but safe to merge on its own.

@orangejenny 

The languages should be available at form start in the 'data' object [here](https://github.com/dimagi/commcare-hq/blob/b322c7ce0e00790b81e4937b6fef19b03737a9cb/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js#L144-L144)  under the `langs` key

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -7,6 +7,7 @@ import api.json.JsonActionUtils;
 import api.process.FormRecordProcessorHelper;
 import api.util.ApiConstants;
 import beans.AnswerQuestionRequestBean;
+import beans.ChangeLocaleRequestBean;
 import beans.FormEntryNavigationResponseBean;
 import beans.FormEntryResponseBean;
 import beans.InstanceXmlBean;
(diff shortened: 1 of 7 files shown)


src/main/java/application/FormController.java
```diff
@@ -94,6 +95,20 @@ public NewFormResponse newFormResponse(@RequestBody NewSessionRequestBean newSes
...
     }
 
+    @ApiOperation(value = "Change the form session locale")
+    @RequestMapping(value = Constants.URL_CHANGE_LANGUAGE, method = RequestMethod.POST)
+    @UserLock
+    @UserRestore
+    @ConfigureStorageFromSession
+    public FormEntryResponseBean changeLocale(@RequestBody ChangeLocaleRequestBean changeLocaleBean,
+                                                @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
+        SerializableFormSession serializableFormSession = formSessionRepo.findOneWrapped(changeLocaleBean.getSessionId());
```
amstone326: It looks like you don't really worry about this in other similar routes, but should we not protect against the possibility of this endpoint getting hit with an invalid or no-longer-present session ID, and therefore having no form session returned here?

wpride: @amstone326 We handle that explicitly [here](https://github.com/dimagi/formplayer/blob/1258b69e83b887638881e815fbdc3ecd803faaff/src/main/java/repo/impl/PostgresFormSessionRepo.java#L192-L192) and reset the session in the front end

amstone326: Ah sorry, I should have looked a bit more at that. Thanks.

PR #655, opened 2018-06-11
Fix ShortestCycleAlgorithm bug

wpride:
addresses https://manage.dimagi.com/default.asp?277481

Check for nullity of edges before running DFS

Also, log when the `validate` call crashes

cross-request https://github.com/dimagi/commcare-core/pull/812


amstone326:
@wpride build failure

PR #656, opened 2018-06-19
Implement HMac verification for Touchforms requests

wpride:
Implements HMac Verification of requests from Touchforms

Partner of https://github.com/dimagi/commcare-hq/pull/20916

### DIFF
--- a/src/main/java/application/FormplayerAuthFilter.java
+++ b/src/main/java/application/FormplayerAuthFilter.java
@@ -59,7 +59,7 @@ public class FormplayerAuthFilter extends OncePerRequestFilter {
                 return;
             }
 
-            if (request.getHeader(Constants.HMAC_HEADER) != null) {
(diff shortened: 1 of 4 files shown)


src/main/java/application/FormplayerAuthFilter.java
```diff
@@ -48,21 +47,37 @@
...
             // These are order dependent
             if (getSessionId(request) == null) {
                 setResponseUnauthorized(response, "Invalid session id");
                 return;
             }
 
-            if (authKey != null && authKey.equals(getSessionId(request))) {
-                logger.info("Logging in as touchforms_user");
-                setSmsRequestDetails(request);
+            if (request.getHeader(Constants.HMAC_HEADER) != null) {
```
gcapalbo: can we also check that the formplayer auth key being used actually has a value (not null or empty string)?

PR #659, opened 2018-06-21
Don't require sessionId cookie

wpride:
Don't make `sessionId` cookie required in methods that SMS uses, only check for `sessionId` when not doing SMS authentication.

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -90,7 +90,7 @@ public class FormController extends AbstractBaseController{
     @UserLock
     @UserRestore
         public NewFormResponse newFormResponse(@RequestBody NewSessionRequestBean newSessionBean,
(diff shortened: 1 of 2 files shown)


src/main/java/application/FormController.java
```diff
@@ -90,7 +90,7 @@
     @UserLock
     @UserRestore
         public NewFormResponse newFormResponse(@RequestBody NewSessionRequestBean newSessionBean,
-                                           @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
+                                           @CookieValue(name=Constants.POSTGRES_DJANGO_SESSION_ID, required=false) String authToken) throws Exception {
```
gcapalbo: what's the difference here for when the view is accessed outside the context of mac digest validation? Will it still return an error if no session id is passed in?

PR #661, opened 2018-06-23
Add HMAC auth to case restores

wpride:
@gcapalbo @snopoke

PR #664, opened 2018-06-25
Remove authKey check

wpride:
Per https://github.com/dimagi/commcare-hq/pull/20984

Now that SMS will be using HMAC auth rather than Django auth this value is not guaranteed to be present, so remove the check from this aspect.

### DIFF
--- a/src/main/java/aspects/UserRestoreAspect.java
+++ b/src/main/java/aspects/UserRestoreAspect.java
@@ -57,8 +57,9 @@ public class UserRestoreAspect {
     @Before(value = "@annotation(annotations.UserRestore)")
     public void configureRestoreFactory(JoinPoint joinPoint) throws Throwable {
         Object[] args = joinPoint.getArgs();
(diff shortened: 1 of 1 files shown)


src/main/java/aspects/UserRestoreAspect.java
```diff
@@ -98,9 +99,7 @@ public void closeRestoreFactory(JoinPoint joinPoint) throws Throwable {
 
     private HqAuth getAuthHeaders(String domain, String username, String sessionToken) {
         HqAuth auth;
-        if (sessionToken != null && sessionToken.equals(authKey)) {
-            auth = new BasicAuth(touchformsUsername, touchformsPassword);
```
gcapalbo: is this removing the touchforms user auth for formplayer requests to hq? we need to keep that around until all formplayer requests are using the mac digest auth

gcapalbo: actually, aren't these changes already present in https://github.com/dimagi/formplayer/pull/661?

wpride:
closing in favor of https://github.com/dimagi/formplayer/pull/661

PR #668, opened 2018-09-04
fixes end of form navigation bug

jmtroth0:
determines whether or not it's a clmi in the same way [commcare-android](https://github.com/dimagi/commcare-android/blob/fa8737adac20d1e2b378313313164fa323934bda/app/src/org/commcare/activities/EntityDetailActivity.java#L99) does it. then it steps back to the case list.

https://manage.dimagi.com/default.asp?281324

@emord


jmtroth0:
Me neither. But it works! And seems pretty clean

ctsims:
This looks reasonable to me. thanks Jordan

jmtroth0:
Thanks @ctsims. I'll tag you on these in the future

PR #671, opened 2018-10-23
Just kidding, this now fixes the tests

jmtroth0:


jmtroth0:
Some tests call the javarosa DateUtils class, which was using the BrowserValuesProvider, which tried to get the timezone offset used a request bean, which wasn't available in the tests. This makes a mocked class, similar to how the javarosa DateUtils tests mock it so it doesn't need to check that.

Let me know if this isn't a kosher test change or if I should try to make it more specific. I did it at the base level since this seemed like something that would break any test that tried to use DateUtils

@ctsims @shubham1g5 buddy: @Rohit25negi

ctsims:
No specific objections from my side. I think there's some chance there's more than one place this offset is reflected, but they'd also be broken as-is, so feels fine to address as they come up.

PR #674, opened 2019-01-11
use long detail for detail screen

jmtroth0:
@ctsims @dannyroberts

### DIFF
--- a/src/main/java/beans/menus/EntityListResponse.java
+++ b/src/main/java/beans/menus/EntityListResponse.java
@@ -66,6 +66,7 @@ public class EntityListResponse extends MenuBean {
                 throw new ApplicationConfigException(String.format("Could not create detail %s for case with ID %s " +
                         " either because the case filter matched zero or multiple cases.", detailSelection, detail));
             }
+            detail = nextScreen.getLongDetailList(reference)[0];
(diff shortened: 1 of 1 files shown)


jmtroth0:
https://github.com/dimagi/commcare-core/pull/835 needs to be pulled in before merging this

jmtroth0:
bumping this for review. or maybe @shubham1g5

ctsims:
@jmtroth0 What's the user facing behavior as a results of this change when there's no long detail defined? It doesn't look like it falls back intentionally to the short detail, does it skip it entirely?

jmtroth0:
It would have just skipped it entirely. Now it will fall back to the short detail, although I'm not sure how to test that since I'm not sure how to create the case where there is no LongDetail

ctsims:
@jmtroth0 you can create a "no long detail" scenario on HQ by just deleting all of the fields on the Case Detail config to test your scenario. 

Signed off as soon as it's confirmed that that works.

ctsims:
Hm, wait, I think I misunderstood.

"Skip entirely" (IE: "I click on the row and then immediately am done with detail selection") is the right behavior for a missing long detail, so if that's what happens when it's missing we should do that (after confirming)

jmtroth0:
Ok, that works now, skipping the detail page if there isn't any data on it.

PR #678, opened 2019-04-25
check if RestoreFactory configured before using it

snopoke:
NullPointerException on staging caused by null username
in restore factory since it isn't configured when the
exception is raised

buddies @kaapstorm @emord

### DIFF
--- a/src/main/java/util/FormplayerSentry.java
+++ b/src/main/java/util/FormplayerSentry.java
@@ -161,7 +161,7 @@ public class FormplayerSentry {
         String username = "unknown";
         String synctoken = "unknown";
         String sandboxPath = "unknown";
-        if (restoreFactory == null || not restoreFactory.isConfigured()) {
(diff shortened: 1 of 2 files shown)


src/main/java/util/FormplayerSentry.java
```diff
@@ -158,17 +158,25 @@ private String getAppDownloadURL() {
     }
 
     private EventBuilder getDefaultBuilder() {
+        String username = "unknown";
+        String synctoken = "unknown";
+        String sandboxPath = "unknown";
+        if (restoreFactory == null || not restoreFactory.isConfigured()) {
```
shubham1g5: Should it be `if (restoreFactory != null  && restoreFactory.isConfigured())`

snopoke:
```suggestion
        if (restoreFactory != null && restoreFactory.isConfigured()) {
```

snopoke:
@shubham1g5 updated

PR #681, opened 2019-07-08
Adds Support for Root level attribute lookup for Indexed Fixtures 

shubham1g5:
Also adds support for non-string values in sql insert operations. 

Corresponding Android PR: https://github.com/dimagi/commcare-android/pull/2085
cross-request: https://github.com/dimagi/commcare-core/pull/858

### DIFF
--- a/src/main/java/engine/FormplayerIndexedFixtureInstanceTreeElement.java
+++ b/src/main/java/engine/FormplayerIndexedFixtureInstanceTreeElement.java
@@ -1,0 +1,51 @@
+package engine;
+
+import org.commcare.cases.instance.IndexedFixtureInstanceTreeElement;
+import org.commcare.cases.model.StorageIndexedTreeElementModel;
+import org.commcare.core.interfaces.UserSandbox;
+import org.javarosa.core.model.IndexedFixtureIdentifier;
+import org.javarosa.core.model.instance.AbstractTreeElement;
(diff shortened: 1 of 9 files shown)


src/main/java/engine/FormplayerIndexedFixtureInstanceTreeElement.java
```diff
@@ -0,0 +1,51 @@
...
+            return new FormplayerIndexedFixtureInstanceTreeElement(instanceBase, storage, indexedFixtureIdentifier);
+        }
+    }
+
+    protected synchronized TreeElement loadAttributes() {
+        if (attributes == null) {
+            try {
+                attributes = SerializationUtil.deserialize(attrHolder, TreeElement.class);
+            } catch (Exception e) {
+                if (e.getCause() instanceof DeserializationException) {
```
ctsims: It seems really scary to me that this swallows errors that aren't deserialization exceptions. What's the reason for that?

shubham1g5: @ctsims Hey thanks for catching this, that's definitely a mistake. Corrected it.

ctsims: Cool, looks good to me now.

PR #682, opened 2019-07-13
Formplayer 2.47

shubham1g5:


ctsims:
This is just  the backmerge, right?

shubham1g5:
@ctsims  More like an update PR, though no need to review since all the changes have been reviewed in individual PRs .

A question before I merge this though - Do we need to do some communication internally on possible side effects of updating the User DB version here ? Incomplete forms should not get deleted on just changing user DB version right or they do ? I did test it though just wanna confirm.

ctsims:
Updating the UserDB version should be sufficient.

Incomplete forms can get deleted on a UserDB update, but we shouldn't be shy about that, as we've communicated fairly consistently that incomplete forms are not a way to store long term data, and warning people or treating it like an issue for them to get wiped out reinforces the opposite.

PR #684, opened 2019-07-22
Adds timeouts on download requests

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/HI-776
Sentry: https://sentry.io/organizations/dimagi/issues/1037055401/?project=142105&query=is%3Aunresolved

In the sentry trace error above, the app profile download always seems stuck while reading the network response from stream. Since we were not setting any explicit timeouts on these request, it can cause the thread to get stuck and hold the lock infinitely. 

cross-request: https://github.com/dimagi/commcare-core/pull/863


shubham1g5:
Closing PR since app download requests generally take substantially more time than a minute and defining timeout can be troublesome here.

PR #688, opened 2019-10-16
Add HQ request timeouts

millerdev:
Read timeout of 15 minutes matches the expiration period of locks used around restore requests to hopefully prevent those requests from hanging indefinitely and subsequently locking a user out of formplayer (until the next restart). While it's good to have this in place as a fallback, 15 minute lock periods make for a bad user experience, and this does not fix the root cause of that issue.

### DIFF
--- a/src/main/java/util/Constants.java
+++ b/src/main/java/util/Constants.java
@@ -108,6 +108,8 @@ public class Constants {
     public static final int USER_LOCK_TIMEOUT = 21;
     // 15 minutes in milliseconds
     public static final int LOCK_DURATION = 60 * 15 * 1000;
+    public static final int CONNECT_TIMEOUT = 60 * 1000;
+    public static final int READ_TIMEOUT = LOCK_DURATION;
 
     //Misc
     public static String HMAC_HEADER = "X-MAC-DIGEST";
--- a/src/main/java/application/WebAppContext.java
+++ b/src/main/java/application/WebAppContext.java
@@ -32,6 +32,7 @@ import org.springframework.data.redis.core.RedisTemplate;
 import org.springframework.data.redis.core.StringRedisTemplate;
 import org.springframework.data.redis.serializer.GenericJackson2JsonRedisSerializer;
 import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.web.client.RestTemplate;
 import org.springframework.web.servlet.ViewResolver;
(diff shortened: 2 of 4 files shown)


src/main/java/util/Constants.java
```diff
@@ -108,6 +108,8 @@
     public static final int USER_LOCK_TIMEOUT = 21;
     // 15 minutes in milliseconds
     public static final int LOCK_DURATION = 60 * 15 * 1000;
+    public static final int CONNECT_TIMEOUT = 60 * 1000;
```
ctsims: The documentation implies that this connection timeout is for "establishing a connection," which sounds correct to me although I'm not confident in when Java has decided that has occurred . Did you test this against an HQ restore which took more than 60 seconds to return to ensure that it doesn't timeout if it hasn't received a response code within this duration?

millerdev: I did not test it against an actual restore. My understanding of [the documentation](https://docs.oracle.com/javase/8/docs/api/java/net/URLConnection.html#setConnectTimeout-int-)

> If the timeout expires before the connection can be established, a java.net.SocketTimeoutException is raised.

is that it is the amount of time allowed for the connection to reach the ESTABLISHED state, which happens to be the state of the hung connections. This [SO answer](https://stackoverflow.com/a/3069450/10840) seems pretty clear.

It sounds like you're wondering if `connectTimeout` might be the total amount of time to complete the request, but nothing I've read has implied anything like that. Are you concerned that the implementation differs from the documentation?

millerdev:
Note on tests: I ran them locally and 4 `FormValidatorTests` failed, but those same tests were also failing for me before this PR was created. I merged since this PR did not cause any new tests to fail in my local environment.

The travis failure looks unrelated, possibly something to do with java version detection?
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine java version from '11.0.2'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

The command "eval ./gradlew assemble " failed 3 times.
```

dannyroberts:
I confirmed by putting it on staging that this PR does appear to cause the 401 sync issue. Reproduce by opening webapps and hitting the sync button

ctsims:
@millerdev I ran this locally while testing another branch and it looks like the 401 is coming from when I try to sync /start without "Login As" active. With Login As it works as expected.

PR #690, opened 2019-10-18
Revert "Add HQ request timeouts"

sravfeyn:
The deploy of this code coincides with formplayer 401 errors from HQ on restore URLs. 

Notice the time of the first occurrance of the current formplayer errors
```
(production) cchq@formplayer2-production:~/www/production/log$ grep "HttpClientErrorException: 401" formplayer-spring.log | head
2019-10-17 19:30:16.851 ERROR 4660 --- [nio-8181-exec-3] application.AbstractBaseController       : Exception org.springframework.web.client.HttpClientErrorException: 401 Unauthorized making external request util.FormplayerHttpRequest@5864d204.
2019-10-17 19:39:22.577 ERROR 4660 --- [nio-8181-exec-2] application.AbstractBaseController       : Exception org.springframework.web.client.HttpClientErrorException: 401 Unauthorized making external request util.FormplayerHttpRequest@7efa5cf0.
```
Notice the deploy time of formplayer, matches with those exceptions
```
lrwxrwxrwx 1 root root   68 Oct 17 19:23 current -> /home/cchq/www/production/formplayer_build/releases/2019-10-17_19.23/
```

Notice the nginx 401s for webapps requests
```
35.175.35.59 - - - - - [18/Oct/2019:09:04:32 +0000] "GET /a/sravan-test/phone/restore/?version=2.0&device_id=WebAppsLogin HTTP/1.1" 401 639 0 "-" "Apache-HttpClient/4.5.2 (Java/1.8.0_222)"
```

I am not sure how [this PR](https://github.com/dimagi/formplayer/pull/688) would cause these issues, so I am attempting a revert of this PR since entire webapps is down due to this issue. The context also available on https://dimagi.slack.com/archives/C0B44C0Q2/p1571391889050300

@calellowitz

PR #697, opened 2019-11-05
Add HQ request timeouts (second try)

millerdev:
Reverts https://github.com/dimagi/formplayer/pull/690

Changes since https://github.com/dimagi/formplayer/pull/688: https://github.com/dimagi/formplayer/commit/8501145de6fb84e7bcd767311ac69e105cd153c0

The default scope is singleton, so the previous version (#688) created a single `RestTemplate` instance that was used for all restore requests.

This has been deployed to and tested on staging. Testing actions taken:
- Navigate to [app](https://staging.commcarehq.org/a/dmiller-staging/apps/view/a6ee6bcff4f11ba66e7a6b0af966daf0/)
- Sync (success)
- Navigate to form
- Submit [form](https://staging.commcarehq.org/a/dmiller-staging/reports/form_data/833abd19-3ec5-4fef-9a32-6da470fea448/)
- Sync (success)

According to @dannyroberts a similar test workflow had failed on staging when he deployed and tested #688 there.

### DIFF
--- a/src/main/java/application/WebAppContext.java
+++ b/src/main/java/application/WebAppContext.java
@@ -32,6 +32,7 @@ import org.springframework.data.redis.core.RedisTemplate;
 import org.springframework.data.redis.core.StringRedisTemplate;
 import org.springframework.data.redis.serializer.GenericJackson2JsonRedisSerializer;
 import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.web.client.RestTemplate;
 import org.springframework.web.servlet.ViewResolver;
 import org.springframework.web.servlet.config.annotation.DefaultServletHandlerConfigurer;
 import org.springframework.web.servlet.config.annotation.EnableWebMvc;
@@ -60,6 +61,7 @@ import services.RestoreFactory;
 import services.SubmitService;
 import services.SyncRequester;
 import services.XFormService;
+import util.Constants;
 import util.FormplayerSentry;
 
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


millerdev:
> I guess the singleton had baked in the wrong auth at some point?

That's my theory, yes. However, I do not see how `RestTemplate` is stateful, so I'm not sure I can explain it.

> Is the "request" scope referring to a single request to the formplayer spring server, or each request that is made from formplayer?

My understanding it that it's a single request to the formplayer spring server. Not each request made by formplayer.

> In principle if formplayer makes different outgoing requests with different auth while processing a single incoming request, we might see the same problem again?

In theory, yes? However, [`RestoreFactory`](https://github.com/dimagi/formplayer/blob/e66321cc7dbab050f7336ffcd381b7a575ea09b0/src/main/java/application/WebAppContext.java#L254-L256), which is making the restore requests has the same "request" scope. I assume we'd want to change that if we ever made multiple requests with different auth in the context of a single request to formplayer. Assuming there is some auth or session caching happening somewhere, I think it makes sense to keep it scoped on the spring request for now.

dannyroberts:
Cool sounds good.

> However, RestoreFactory, which is making the restore requests has the same "request" scope.

Does this mean that if there were an issue with request-scoped outgoing request–builders, we'd likely have seen those issues already because of the request-scoped `RestoreFactory`? Or is that a leap that I made there?

millerdev:
> Or is that a leap that I made there?

I think you made a leap there. But since I still don't fully understand the failure mode you could be correct.

PR #700, opened 2019-11-06
let spring manage the DB datasource

snopoke:
This will need to wail until the config changes have been pushed out: https://github.com/dimagi/commcare-cloud/pull/3377


snopoke:
@ctsims thanks for the suggestion.

PR #701, opened 2019-11-06
Lock in Gradle dependency versions for modernized Spring libs

ctsims:
It's best practice for production Gradle builds to have explicit version dependencies, to ensure that build products are consistent.

This locks in the most up to date versions of the release libs, using spring-boot's new dependency manager for the libraries covered under its POM, and setting explicit dependencies for any other libraries without one.

To accept spring-boot's new versions, I also had to update a couple of JDBC connector details. They are tested locally, but have not been tested in a staging environment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,46 +39,46 @@ idea {
 }
 
 dependencies {
-    compile group: 'org.springframework.boot', name: 'spring-boot'
-    compile group: 'org.springframework.boot', name: 'spring-boot-starter-web'
-    compile group: 'org.springframework.boot', name: 'spring-boot-starter-actuator'
-    compile group: 'org.springframework.boot', name: 'spring-boot-starter'
-    compile group: 'org.springframework.boot', name: 'spring-boot-starter-jdbc'
-    compile group: 'org.springframework.boot', name: 'spring-boot-autoconfigure'
-    compile group: 'org.springframework.data', name: 'spring-data-jpa'
-    compile("org.springframework:spring-web")
-    compile group: 'org.springframework.data', name: 'spring-data-redis'
-    compile group: 'org.springframework.boot', name: 'spring-boot-starter-data-jpa'
+    compile group: 'org.springframework.boot', name: 'spring-boot', version: '2.2.0.RELEASE'
+    compile group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: '2.2.0.RELEASE'
+    compile group: 'org.springframework.boot', name: 'spring-boot-starter-actuator', version: '2.2.0.RELEASE'
+    compile group: 'org.springframework.boot', name: 'spring-boot-starter', version: '2.2.0.RELEASE'
+    compile group: 'org.springframework.boot', name: 'spring-boot-starter-jdbc', version: '2.2.0.RELEASE'
+    compile group: 'org.springframework.boot', name: 'spring-boot-autoconfigure', version: '2.2.0.RELEASE'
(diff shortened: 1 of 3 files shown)


snopoke:
The versions should be set by the spring boot plugin according to the version of spring boot you are using:
https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/html/#managing-dependencies (edited) 
more info on Spring BOM: https://www.baeldung.com/spring-maven-bom

snopoke:
Having said that - it looks like you may have set the versions for some libraries that are outside the Spring BOM. I think this is what's in the BOM: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-dependencies/pom.xml

ctsims:
Hey @snopoke ,

I read a bit more and you are correct that the spring-boot managed dependencies only cover the spring boot libraries, effectively. I isolated the boot dependencies after reviewing the POM and removed their versions to let the plugin version take control. 

There is a [more generic Spring dependency management](https://docs.spring.io/dependency-management-plugin/docs/current-SNAPSHOT/reference/html/) plugin that will let you import POM files manually, but the other libs have very few managed/coupled dependencies, so I don't think we'll get much from it.

Should be ready for review again. This required me to make some more core changes to address some changes to the new JDBC library after accepting spring-boot's new version of it, so I agree that it would be good to put these through staging first.

build.gradle
```diff
@@ -39,51 +39,63 @@ idea {
...
-    compile group: 'org.hibernate', name: 'hibernate-core', version: '5.4.7.Final'
-    compile group: 'org.springframework.data', name: 'spring-data-commons'
+    compile group: 'io.springfox', name: 'springfox-swagger2', version: '2.4.0'
+    compile group: 'io.springfox', name: 'springfox-swagger-ui', version: '2.4.0'
     compile group: 'javax.mail', name: 'mail', version: '1.5.0-b01'
-    compile group: 'com.datadoghq', name: 'java-dogstatsd-client', version: '2.3'
-    compile group: 'org.springframework.integration', name: 'spring-integration-redis'
-    compile 'org.xerial:sqlite-jdbc:3.16.1'
+    compile group: 'org.apache.commons', name: 'commons-email', version: '1.2'
+    compile group: 'org.apache.directory.studio', name: 'org.apache.commons.io', version: '2.0.1'
```
snopoke: do we use this? I don't think so.

build.gradle
```diff
@@ -39,51 +39,63 @@ idea {
...
+    compile group: 'org.json', name: 'json', version: '20131018'
+    compile group: 'org.lightcouch', name: 'lightcouch', version: '0.1.6'
+    compile group: 'org.simpleframework', name: 'simple-xml', version: '2.7.1'
+    compile group: 'org.springframework', name: 'spring-web', version: '5.2.0.RELEASE'
+    compile group: 'org.springframework', name: 'spring-orm', version: '5.2.0.RELEASE'
+    compile group: 'org.springframework', name: 'spring-jdbc', version: '5.2.0.RELEASE'
+    compile group: 'org.springframework.data', name: 'spring-data-jpa', version: '2.2.0.RELEASE'
```
snopoke: is it necessary to specify this at all? It's already a dependency of `spring-boot-starter-data-jpa`

ctsims:
@snopoke Agreed on the dependencies. I can likely take another look when I update formplayer to get back on cc-core master. Thanks!

PR #704, opened 2019-11-08
Clear the userDB and retry when we get a 412 on restore

dannyroberts:
I ran into similar issues to what @millerdev fixed in https://github.com/dimagi/formplayer/pull/697, so I reworked the code to use a fresh `RestTemplate` object each time which fixed the issues.

This PR makes Web Apps support a 412 restore response from the server. Previously the behavior was that formplayer threw an error and Web Apps showed a generic error message, and all subsequent clicks showed the same error message, making it impossible to move past it until the db file got deleted.

Now the behavior is that when formplayer gets a 412, it deletes the userDB and syncs again without a sync token (i.e. without `since=...` in the URL). Other than incomplete forms getting silently deleted, and the restore possibly taking longer than normal, there's no sign that anything special happened to the user. I think this is the same as what happens on a 412 on the phone.

### DIFF
--- a/src/main/java/services/RestoreFactory.java
+++ b/src/main/java/services/RestoreFactory.java
@@ -27,6 +27,7 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Component;
+import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 import org.w3c.dom.Document;
 import org.w3c.dom.NamedNodeMap;
@@ -160,8 +161,20 @@ public class RestoreFactory {
         if(getSqlSandbox().getLoggedInUser() != null){
             getSQLiteDB().createDatabaseFolder();
         }
-        UserSqlSandbox sandbox = restoreUser();
-        if (shouldPurge) {
+        UserSqlSandbox sandbox;
+        try {
+             sandbox = restoreUser();
+        } catch (HttpClientErrorException e) {
+            if (e.getStatusCode().value() == 412) {
+                getSQLiteDB().deleteDatabaseFile();
+                // this line has the effect of clearing the sync token
+                // from the restore URL that's used
+                sqLiteDB = new UserDB(domain, username, asUsername);
+                return performTimedSync(shouldPurge);
+            }
+            throw e;
+        }
+        if (shouldPurge && sandbox != null) {
             SimpleTimer purgeTimer = new SimpleTimer();
             purgeTimer.start();
             FormRecordProcessorHelper.purgeCases(sandbox);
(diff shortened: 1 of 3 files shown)


src/main/java/services/RestoreFactory.java
```diff
@@ -159,8 +161,20 @@ public UserSqlSandbox performTimedSync(boolean shouldPurge) throws Exception {
...
-        if (shouldPurge) {
+        UserSqlSandbox sandbox;
+        try {
+             sandbox = restoreUser();
+        } catch (HttpClientErrorException e) {
+            if (e.getStatusCode().value() == 412) {
+                getSQLiteDB().deleteDatabaseFile();
+                // this line has the effect of clearing the sync token
+                // from the restore URL that's used
+                sqLiteDB = new UserDB(domain, username, asUsername);
```
dannyroberts: @ctsims @shubham1g5 My testing cycle locally was just
1. Delete synclogs to force a 412 (using this new button: https://github.com/dimagi/commcare-hq/pull/25850)
2. Go to WebApps and hit sync

Previously the behavior was that it raised an error and prevented you from syncing again. Now it syncs seamlessly.

Do these two lines seem like a reasonable way of clearing user state in before resyncing in response to a 412? On a phone there may be unsent forms, but I think WebApps doesn't do unsent forms, just incomplete forms. Does that sound right? Anything else I'm overlooking?

ctsims: Hi Danny,

On device we do this with an atomic transaction (IE: Try to download the sync, and only clear user data once it's received in full, and if the download fails don't break the state), but I agree that  WebApps doesn't need that level of care and we generally consider it to be a safe action to wipe out someone's user sandbox and force them to resync (it's how we handle ~all migrations).

I might double check that there's no utility method somewhere which wraps 
```
getSQLiteDB().deleteDatabaseFile();
```
in a more semantically meaningful action which might come with some other cleanup or considerations, but the approach itself seems fine

dannyroberts:
> I might double check that there's no utility method somewhere which wraps `getSQLiteDB().deleteDatabaseFile();` in a more semantically meaningful action which might come with some other cleanup or considerations, but the approach itself seems fine

Yeah I looked for this but there doesn't seem to be. It's used directly in a handful of places without otherwise much overlap in the surrounding code, so I followed that example.

dannyroberts:
This is a side note from https://github.com/dimagi/commcare-hq/pull/25850#issuecomment-553160901 so that I can make sure to document the feature correctly.

> @ctsims 
> The new feature will not delete incomplete forms. 412'ing does not result in incomplete forms being removed from the DB, it just clears the cases, ledgers, and lookup tables in the user sandbox.

Because of this PR, a 412 will result in deleting incomplete forms _on formplayer_, right? I don't think that deeply affects the other conversation, but I just wanted to make sure I get it right in the docs.

ctsims:
Hi Danny,

You are correct based on the implementation of 412'ing in your recent PR.

It would be fairly straightforward to re-implement by just deleting the
same tables as mobile
<https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/DataPullTask.java#L552>
rather
than the entire DB, if that's a concern.

-Clayton

PR #706, opened 2019-11-19
Modernize Database Connection Management

ctsims:
Updates to the way that we are managing jdbc connections and objects. 

Ensures that all connection objects are closed out to try and fix the issue with the system leaking paged sqlite journal files.

Replaces (where cleanly possible) meaningless try{} finally{} blocks with cleaner [try with resource](https://www.baeldung.com/java-try-with-resources) patterns, now that we are on a compatible version.

### DIFF
--- a/src/main/java/database/models/EntityStorageCache.java
+++ b/src/main/java/database/models/EntityStorageCache.java
@@ -108,30 +108,21 @@ public class EntityStorageCache {
 
     // Currently unused
     public String retrieveCacheValue(String entityKey, String cacheKey) {
-        Connection connection = null;
-        PreparedStatement preparedStatement = null;
-        try {
-            preparedStatement = SqlHelper.prepareTableSelectStatement(connection,
+        Connection connection = handler.getConnection();
+        try (PreparedStatement preparedStatement = SqlHelper.prepareTableSelectStatement(connection,
                     TABLE_NAME,
                     new String[]{COL_CACHE_NAME, COL_ENTITY_KEY, COL_CACHE_KEY},
-                    new String[]{mCacheName, entityKey, cacheKey});
-            ResultSet resultSet = preparedStatement.executeQuery();
+                    new String[]{mCacheName, entityKey, cacheKey})) {
+            try (ResultSet resultSet = preparedStatement.executeQuery()) {
 
-            if (resultSet.next()) {
-                return resultSet.getString(resultSet.findColumn(COL_VALUE));
-            } else {
-                return null;
+                if (resultSet.next()) {
+                    return resultSet.getString(resultSet.findColumn(COL_VALUE));
+                } else {
+                    return null;
+                }
             }
         } catch (SQLException e) {
             throw new RuntimeException(e);
-        } finally {
-            if (preparedStatement != null) {
-                try {
-                    preparedStatement.close();
-                } catch (SQLException e) {
-                    log.debug("Exception closing prepared statement ", e);
-                }
-            }
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/main/java/database/models/FormplayerCaseIndexTable.java
```diff
@@ -131,38 +131,29 @@ public void clearCaseIndices(Collection<Integer> idsToClear) {
...
-                String relationship = resultSet.getString(COL_INDEX_RELATIONSHIP);
-                Pair<String, String> index  = new Pair<> (targetCase, relationship);
-
-                Vector<Pair<String, String>> indexList;
-                if (!caseIndexMap.containsKey(caseRecordId)) {
-                    indexList = new Vector<>();
-                } else {
-                    indexList = caseIndexMap.get(caseRecordId);
+                    projection)) {
+            try (ResultSet resultSet = selectStatement.executeQuery()) {
```
dannyroberts: I could be confused, but instead of nesting `try`, could you use two resources in one try-with-resource?

```
try (
    PreparedStatement selectStatement = SqlHelper.prepareTableSelectProjectionStatement(
        connectionHandler.getConnection(), TABLE_NAME, projection);
    ResultSet resultSet = selectStatement.executeQuery()
) {
    ...
} catch(...) {
    ...
}
```

I'm getting this from the example towards the bottom of https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

There are about a dozen places in this PR where this comment might apply, but I'll let you figure out if it's even a good suggestion

src/main/java/sandbox/SqlHelper.java
```diff
@@ -89,20 +80,11 @@ private static void dumpResultSetRow(ResultSetMetaData metaData, ResultSet resul
 
     public static void dropTable(Connection c, String storageKey) {
         String sqlStatement = "DROP TABLE IF EXISTS " + storageKey;
-        PreparedStatement preparedStatement = null;
-        try {
-            preparedStatement = c.prepareStatement(sqlStatement);
+        try (PreparedStatement preparedStatement = c.prepareStatement(sqlStatement)) {
+            ;
```
dannyroberts: stray `;`?

src/main/java/database/models/EntityStorageCache.java
```diff
@@ -108,30 +108,21 @@ public void cache(String entityKey, String cacheKey, String value) throws SQLExc
 
     // Currently unused
     public String retrieveCacheValue(String entityKey, String cacheKey) {
-        Connection connection = null;
-        PreparedStatement preparedStatement = null;
-        try {
-            preparedStatement = SqlHelper.prepareTableSelectStatement(connection,
+        Connection connection = handler.getConnection();
```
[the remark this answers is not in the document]
ctsims: This code was broken before and couldn't have worked, as far as I can tell. I believe someone copied it from another place where they were trying to use the "establish connection inside the try" pattern, and just missed step 2.

src/main/java/database/models/FormplayerCaseIndexTable.java
ctsims: You can declare more than one resource in the block., yeah. I didn't do it often since there were a few places where we were looping one of the objects or it wasn't super obvious whether intermediate code could be re-arranged. 

I think it would be the better pattern of the two in a few places, and will try to come back and apply if I have time.

src/main/java/sandbox/SqlHelper.java
ctsims: Thanks. Fixing.

PR #707, opened 2019-11-19
Fix Issues with Dangling DB Connections

ctsims:
The code which actually builds and establishes the SQLite connections was attempting to set the autoCommit behavior of the connection do a default state before returning it to the caller, and in doing so was unintentionally creating a hanging connection which was never closed.

This code doesn't push down the `autoCommit` portion of the change to the actually created connection, since the attempt to set the behavior of the connection to autoCommit was failing since it was only changing the behavior of the orphaned connection. It may be worth exploring whether actually setting the autoCommit value would work and/or have performance gains, but I think the system expects autoCommit to be active by default now and can't see how it would work otherwise.


dannyroberts:
Yeah nice

PR #710, opened 2020-03-25
Fix HTTP Connection issues due to re-use of Rest Template

ctsims:
Locally, Formplayer had issues for me when I made requests too quickly back to back because the RestTemplate configured to make the auth request was being re-used, including the underlying connection object, resulting in

`localhost:8000 failed to respond; nested exception is org.apache.http.NoHttpResponseException: localhost:8000 failed to respond`

This ensures that a new RestTemplate is used for each auth attempt and fixes the issue for me. It looks like the only reason the old behavior was configured was to support a mock for testing, so this should be a safe change.

### DIFF
--- a/src/main/java/services/HqUserDetailsService.java
+++ b/src/main/java/services/HqUserDetailsService.java
@@ -20,6 +20,7 @@ import util.RequestUtils;
 public class HqUserDetailsService {
     private final Log log = LogFactory.getLog(HqUserDetailsService.class);
     private final RestTemplate restTemplate;
+    private final RestTemplateBuilder builder;
 
     @Value("${commcarehq.host}")
     private String host;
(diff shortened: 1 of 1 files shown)


ctsims:
@shubham1g5 Not sure if this might fix some of the issues you and @sravfeyn have been looking into?

shubham1g5:
@ctsims  For my understanding, Are `restTemplates` not meant to be reused across requests ? I was kind of treating them as reusable connection objects.

ctsims:
@shubham1g5 RestTemplate's are really minimal objects. They don't handle connection pooling, retries, etc, they really just wrap a Java UrlConnection object. I think the general pattern is to use the RestTemplateBuilder() to provide the reusable connection details. 

And to be clear - I'm not sure what the exact request lifecycle for this object was. I don't think it's meant to be cached the way the code was caching it, even if RestTemplates are ok to reuse in other contexts.

PR #711, opened 2020-03-25
Basic post-form autosync

ctsims:
Spec: https://github.com/dimagi/commcare-hq/issues/26940

Probably won't be useful for a lot of domains where syncs take a while.

Shouldn't be used broadly in production until better tested

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -195,6 +195,13 @@ public class FormController extends AbstractBaseController{
                 }
             }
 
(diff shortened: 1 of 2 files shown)


ctsims:
Pulling because travis actually succeeded a while ago. Not sure why the CI hook shows it as still pending
https://travis-ci.org/github/dimagi/formplayer/builds/666898591?

benrudolph:
nice, 13 LOCs for this feature 👌

PR #715, opened 2020-04-06
Permit more aggressive autosync

ctsims:
When in "sync after submit"mode, force a sync on session refresh if state by 20 minutes.

### DIFF
--- a/src/main/java/services/RestoreFactory.java
+++ b/src/main/java/services/RestoreFactory.java
@@ -310,7 +310,7 @@ public class RestoreFactory {
         try {
             return storageFactory.getPropertyManager().isSyncAfterFormEnabled();
         } catch (RuntimeException e) {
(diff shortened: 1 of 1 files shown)


src/main/java/services/RestoreFactory.java
```diff
@@ -305,6 +306,16 @@ public String getSyncFreqency() {
         }
     }
 
+    public boolean useAggressiveSyncTiming() {
+        try {
+            return storageFactory.getPropertyManager().isSyncAfterFormEnabled();
+        } catch (RuntimeException e) {
+            // In cases where we don't have access to the PropertyManager, such sync-db, this call
```
millerdev, line 313: ?
```suggestion
            // In cases where we don't have access to the PropertyManager, such as sync-db, this call
```

PR #716, opened 2020-04-17
Form Volatility Indicators

ctsims:
Allows forms to declare a volatility key for their case entities. When volatile forms are begun of completed for a specific case a record is made, and users are warned if they open a form which was recently opened by another user, or submitted by another user (unless they have synced since after the form was submitted). 

image

This is hidden for now, and shouldn't affect most domains. I don't love the fact that the volatility indicators are stored inside the form record, I think it would be way better if this was generalized more heuristically against the session, but it will be easy to build on top of later.

### DIFF
--- a/src/main/java/application/FormController.java
+++ b/src/main/java/application/FormController.java
@@ -13,6 +13,8 @@ import org.simpleframework.xml.core.Persister;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.data.redis.core.RedisTemplate;
+import org.springframework.data.redis.core.ValueOperations;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.CookieValue;
 import org.springframework.web.bind.annotation.RequestBody;
@@ -24,6 +26,8 @@ import org.springframework.web.bind.annotation.RestController;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+
+import javax.annotation.Resource;
 
 import annotations.ConfigureStorageFromSession;
 import annotations.UserLock;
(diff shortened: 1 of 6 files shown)


src/main/java/session/FormSession.java
```diff
@@ -237,6 +248,55 @@ private void initialize(boolean newInstance, Map<String, String> sessionData, St
...
+        if(value != null) {
+            return formDef.fillTemplateString(
+                    value, TreeReference.rootRef());
+        }
+        return null;
+    }
+
+    private void setVolatilityIndicators()
+    {
+        String volatilityKey = getPragma("Pragma-Volatility-Key");
```
shubham1g5, line 266: Not sure if I understand where these keys are getting set. I have not seen these keys in Form xml, is this a new change on HQ ?

src/main/java/application/FormController.java
```diff
@@ -143,6 +156,8 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
                 formEntrySession.getFormEntryModel(),
                 submitRequestBean.getAnswers());
 
+        FormVolatilityRecord volatilityRecord = formEntrySession.getSessionVolatilityRecord();
```
shubham1g5, line 159: Nit: would be good to move this down to where it ends up getting used. (above L213)

src/main/java/session/FormSession.java
ctsims: Right now, it's manual. I'm adding questions to the form explicitly with these itext keys

ctsims: In the future I think these might get set in the session instead, since otherwise we can't do things like sync **before** loading the form cleanly. Wanted to figure that out before coding in HQ.

snopoke: Just for my understanding this would be like a case ID and `Pragma-Volatility-Entity-Title` would be the case name or something right?

ctsims: Yeah, in the base case. If the form selected more than one thing, those would both need to be more complex to reflect the compound selection, but for most HQ use cases it's just case_id and case_name

PR #719, opened 2020-05-07
Exposing Error Notifications to Sentry

shubham1g5:
* Exposes error notifications to sentry and datadog counters
* Sends all kind of exceptions to Sentry

Notes: 
1. Would be great to get this deployed to staging to test first (As my HQ setup is still not great, I was not able to fully test it)

2. Is there a better Spring way to be able to avoid multiple `logNotification` calls here ?

### DIFF
--- a/src/main/java/services/InstallService.java
+++ b/src/main/java/services/InstallService.java
@@ -56,8 +56,8 @@ public class InstallService {
                     engine.initEnvironment();
                     return new Pair<>(engine, false);
                 } catch (Exception e) {
-                    log.warn("Warning: An error occurred while trying to use the old DB file for app. CommCare will now try to install the new DB file for this app. Error details: Got exception "
-                            + e + " while reinitializing at path " + sqliteDB.getDatabaseFileForDebugPurposes());
+                    log.debug("An error occurred while trying to use the old DB file for app. Error details: Got exception "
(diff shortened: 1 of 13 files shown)


dannyroberts:
@shubham1g5 awesome! I can't tell if the build failures are "real", but once you're ready just let me know and I can put this up on staging

shubham1g5:
@dannyroberts ugh, some tests were failing. Thanks for flagging. @ShivamPokhriyal  tested this locally and it's working good.

src/main/java/services/InstallService.java
```diff
@@ -56,8 +56,8 @@
                     engine.initEnvironment();
                     return new Pair<>(engine, false);
                 } catch (Exception e) {
-                    log.warn("Warning: An error occurred while trying to use the old DB file for app. CommCare will now try to install the new DB file for this app. Error details: Got exception "
-                            + e + " while reinitializing at path " + sqliteDB.getDatabaseFileForDebugPurposes());
+                    log.debug("An error occurred while trying to use the old DB file for app. Error details: Got exception "
+                            + e + " while reinitializing at path " + sqliteDB.getDatabaseFileForDebugPurposes() + " Reinitializing new DB ..");
```
dannyroberts, line 60: do `log.warn` and `log.debug` show up to the user at all?

shubham1g5: No, this only shows in console. This is a totally normal error and was leading developers to incoorectly believe that there is something wrong with the Formplayer, so making it slightly less intruisive.

PR #721, opened 2020-05-08
Use time in seconds for getDurationBucket method call

ShivamPokhriyal:
JIRA: https://dimagi-dev.atlassian.net/browse/SAAS-10832

### DIFF
--- a/src/main/java/aspects/MetricsAspect.java
+++ b/src/main/java/aspects/MetricsAspect.java
@@ -92,7 +92,7 @@ public class MetricsAspect {
     }
 
     private String getUnblockedTimeBucket(SimpleTimer timer) {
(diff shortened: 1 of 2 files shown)


src/main/java/aspects/MetricsAspect.java
```diff
@@ -115,7 +115,7 @@ private long getRestoreBlockedTime() {
         if (restoreFactory.getDownloadRestoreTimer() == null) {
             return 0;
         }
-        return restoreFactory.getDownloadRestoreTimer().durationInMs();
```
shubham1g5, line 118: this method gets used in the calculation above for `getBlockedTime` which seems to be expecting this value in ms.

shubham1g5: wrong flag. resolving.

PR #722, opened 2020-05-09
Lock Eviction Endpoint

ctsims:
Many long running request operations in formplayer can leave user sandboxes locked for very long amounts of time with no recourse. This makes operations like load testing especially painful.

This introduces an endpoint which forces the same 'lock eviction' process as would naturally occur if a request had come after the long timeout. 

Includes: https://github.com/dimagi/commcare-core/pull/893 to increase coverage of interruptable requests

### DIFF
--- a/src/main/java/application/UtilController.java
+++ b/src/main/java/application/UtilController.java
@@ -3,6 +3,7 @@ package application;
 import annotations.NoLogging;
 import annotations.UserLock;
 import annotations.UserRestore;
+import aspects.LockAspect;
 import beans.*;
 import hq.CaseAPIs;
 import io.swagger.annotations.Api;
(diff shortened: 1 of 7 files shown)


src/main/java/application/UtilController.java
```diff
@@ -97,6 +108,36 @@ public NotificationMessage clearUserData(
...
+        Integer secondsLocked = userLockRegistry.getTimeLocked(key);
+        if(secondsLocked == null) {
+            return new LockReportBean(false, 0);
+        } else{
+            return new LockReportBean(true, secondsLocked);
+        }
+    }
+
+    @ApiOperation(value = "Breaks any currently locked requests for the current user")
+    @RequestMapping(value = Constants.URL_BREAK_LOCKS, method = RequestMethod.POST)
```
shubham1g5, line 126: Is there an user action associated with this on HQ side ? Or HQ is just goign to poll Formplayer periodically to check for any long running locks and subsequently timeout these ?

ctsims: There's a manual action currently in this PR that shows up under the settings menu: 
https://github.com/dimagi/commcare-hq/pull/27430

but not sure what the final presentation will be (like a popup in a new screen trying to do a request that says "Waiting for another action, do you want to cancel it?"), there. My guess is that we'll need to split out accessors to the session by "read only" or something different to allow some long running requests to be made in parallel.

PR #723, opened 2020-05-14
Enable fuzzy search

ShivamPokhriyal:


src/test/java/tests/CasePaginationTests.java
```diff
@@ -95,4 +111,15 @@ public void testSearchAndPagination() throws Exception {
         assert entityListResponse.getPageCount() == 2;
         assert entityListResponse.getCurrentPage() == 1;
     }
+
+    public class FormPlayerPropertyManagerMock extends FormplayerPropertyManager {
```
shubham1g5, line 115: should be abstracted into it's own file as it's useful for all tests.

src/test/java/tests/CasePaginationTests.java
```diff
@@ -95,4 +111,15 @@ public void testSearchAndPagination() throws Exception {
...
+        public FormPlayerPropertyManagerMock(IStorageUtilityIndexed properties) {
+            super(properties);
+        }
+
+        @Override
+        public boolean isFuzzySearchEnabled() {
```
shubham1g5, line 121: can we add a setter for it and return that value instead instead of always returning `false`.

PR #726, opened 2020-05-26
Lower lock timeout to 30seconds

ShivamPokhriyal:
Jira https://dimagi-dev.atlassian.net/browse/SAAS-10840

### DIFF
--- a/src/main/java/util/Constants.java
+++ b/src/main/java/util/Constants.java
@@ -109,8 +109,8 @@ public class Constants {
     ));
 
     public static final int USER_LOCK_TIMEOUT = 21;
-    // 15 minutes in milliseconds
-    public static final int LOCK_DURATION = 60 * 15 * 1000;
+    // 30 seconds in milliseconds.
+    public static final int LOCK_DURATION = 30 * 1000;
     public static final int CONNECT_TIMEOUT = 60 * 1000;
     public static final int READ_TIMEOUT = LOCK_DURATION;
 


ctsims:
@ShivamPokhriyal How was this timing considered?

There are a significant number of operations in Web Apps that take longer than 30 seconds, so this will potentially expose quite a few unsafe locks

ShivamPokhriyal:
> There are a significant number of operations in Web Apps that take longer than 30 seconds, so this will potentially expose quite a few unsafe locks

I wanted to lower the lock-timing from the current 15 minutes, which appears to be too long. But agree that 30 seconds might be too short for some of the operations. 
Now, when I see the datadog metrics for [`granular.timings`](https://app.datadoghq.com/metric/explorer?from_ts=1588002991130&to_ts=1590594991130&live=true&page=0&is_auto=false&tile_size=m&exp_metric=formplayer.metrics.granular.timings.max&exp_group=category&exp_agg=max&exp_row_type=metric#category:app_install), it seems like the max time is 10 minutes. 
And only a couple operations(like app installation and restore) require longer time while other operations typically finish within 30 seconds. Though, I'm not sure if that's the correct way to figure out the max time required for certain operations.

ctsims:
@ShivamPokhriyal Yeah, I agree that the current timings are pretty crazy, but the timeout applies to ~all actions including initial app install and restore, and we've had circumstances where people perform very long initial setup operations before fast steady state usage.

I don't think it's good design, but we also can't dial it back this drastically for now without an alternative

PR #736, opened 2020-06-11
request logging

knguyen142:
Print out logs for each request with the following:
- request path
- request body
- project space
- username 
- restore as
- response 

## Design
Uses Spring Aspects to tap into before and after lifecycle of a request to grab data. This implementation adds code to (and refactors) the existing logging aspect instead of creating a new class because a lot of the data needed is shared.

## Limitations
This currently only logs from paths that have the `HTTPServletRequest` parameter. The aspect searches for this parameter to grab the request body. One of the paths was changed to add this to test before and after having the parameter.

## Log to file
To only log these lines to a separate file, a logback config file is needed. Add `logging.config=config/logback-spring.xml` to application.properties and use something like this for the `logback-spring.xml`

```
<configuration>

    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
        <!-- encoders are assigned the type
             ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
        <encoder>
            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
        </encoder>
    </appender>

    <appender name="requests" class="ch.qos.logback.core.rolling.RollingFileAppender">
        <file value="requests.log"/>
        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
            <!-- daily rollover -->
            <fileNamePattern>logFile.%d{yyyy-MM-dd}.log</fileNamePattern>

            <!-- keep 30 days' worth of history capped at 3GB total size -->
            <maxHistory>30</maxHistory>
            <totalSizeCap>3GB</totalSizeCap>

        </rollingPolicy>
        <encoder>
            <pattern>%-4relative [%thread] %-5level %logger{35} - %msg%n</pattern>
        </encoder>
    </appender>

    <root level="warn">
        <appender-ref ref="STDOUT" />
    </root>

    <logger name="org.springframework.web.filter.CommonsRequestLoggingFilter">
        <level value="INFO" />
        <appender-ref ref="STDOUT" />
    </logger>

    <logger name="org.springframework.web">
        <level value="INFO" />
        <appender-ref ref="STDOUT" />
    </logger>

    <logger name="org.hibernate">
        <level value="ERROR" />
        <appender-ref ref="STDOUT" />
    </logger>

    <logger name="aspects.LoggingAspect">
        <level value="ALL" />
        <appender-ref ref="requests" />
    </logger>

</configuration>
```

# TODO
There are a few things that I didn't get to work because of Spring noobness.

1. Getting the request body as JSON instead of JSON string.
2. Grabbing the response as JSON (currently gets the response bean)

# TESTING
tested locally
Before adding the httpservletrequest param to answer path
```
47637 [http-nio-8080-exec-4] INFO  aspects.LoggingAspect - {"responseBean":"FormEntryResponseBean: [status=accepted, seq_id: 1]","request_path":"answer","project_space":"test-project","username":"knguyen@dimagi.com"}
```
after
```
13400 [http-nio-8080-exec-2] INFO  aspects.LoggingAspect - {"responseBean":"FormEntryResponseBean: [status=accepted, seq_id: 1]","request_body":"{\"action\":\"answer\",\"ix\":\"0\",\"answer\":\"asdf\",\"oneQuestionPerScreen\":false,\"domain\":\"test-project\",\"username\":\"knguyen@dimagi.com\",\"restoreAs\":null,\"session-id\":\"5b87c0d8-b94f-4954-996d-40ccfca7b207\",\"session_id\":\"5b87c0d8-b94f-4954-996d-40ccfca7b207\",\"debuggerEnabled\":true,\"tz_offset_millis\":-14400000}","request_path":"answer","project_space":"test-project","username":"knguyen@dimagi.com"}
```

More logs
```
16083 [http-nio-8080-exec-4] INFO  aspects.LoggingAspect - {"responseBean":"SubmitResponseBean, [status=success, errors: {}, next screen= null]","request_body":"{\"action\":\"submit-all\",\"answers\":{\"0\":\"asdf\"},\"prevalidated\":true,\"domain\":\"test-project\",\"username\":\"knguyen@dimagi.com\",\"restoreAs\":null,\"session-id\":\"5b87c0d8-b94f-4954-996d-40ccfca7b207\",\"session_id\":\"5b87c0d8-b94f-4954-996d-40ccfca7b207\",\"debuggerEnabled\":true,\"tz_offset_millis\":-14400000}","request_path":"submit-all","project_space":"test-project","username":"knguyen@dimagi.com"}
16212 [http-nio-8080-exec-6] INFO  aspects.LoggingAspect - {"responseBean":"CommandListResponseBean [commands=[Command [index=0, text=Surveys, audioUri=null, imageUri=null]]MenuBean= MenuBean [title=Untitled Application, breadcrumbs: [Untitled Application]super=BaseResponseBean [title=Untitled Application, notification=null, clearSession=false, selections=null]]]","request_body":"{\"username\":\"knguyen@dimagi.com\",\"restoreAs\":null,\"domain\":\"test-project\",\"app_id\":\"8927275bbca183ed6b0f0ca0900033fb\",\"locale\":\"en\",\"offset\":null,\"oneQuestionPerScreen\":false,\"preview\":true,\"geo_location\":null,\"tz_offset_millis\":-14400000}","request_path":"navigate_menu","project_space":"test-project","username":"knguyen@dimagi.com"}
702179 [http-nio-8080-exec-10] INFO  aspects.LoggingAspect - {"responseBean":"CommandListResponseBean [commands=[Command [index=0, text=Survey, audioUri=null, imageUri=null]]MenuBean= MenuBean [title=Surveys, breadcrumbs: [Untitled Application, Surveys]super=BaseResponseBean [title=Surveys, notification=null, clearSession=false, selections=null]]]","request_body":"{\"username\":\"knguyen@dimagi.com\",\"restoreAs\":null,\"domain\":\"test-project\",\"app_id\":\"8927275bbca183ed6b0f0ca0900033fb\",\"locale\":\"en\",\"selections\":[0],\"offset\":0,\"search_text\":null,\"menu_session_id\":null,\"query_dictionary\":null,\"oneQuestionPerScreen\":false,\"sortIndex\":null,\"preview\":true,\"geo_location\":null,\"tz_offset_millis\":-14400000}","request_path":"navigate_menu","project_space":"test-project","username":"knguyen@dimagi.com"}
```

Verified that the request body was the same as what was being sent from Chrome using inspector tool.


src/main/java/aspects/LoggingAspect.java
```diff
@@ -29,33 +34,80 @@
...
-        log.info("Request to " + requestPath + " returned result " + result);
+
+        JSONObject logLine = (new JSONObject())
+                .put("request_path", requestPath)
+                .put("request_body", requestBody)
+                .put("project_space", domain)
+                .put("username", username)
+                .put("restore_as", restoreAs)
+                // TODO: Figure out a way to get response JSON.
+                .put("responseBean", result);
```
dannyroberts, line 74: According to https://stackoverflow.com/questions/11692303/turning-a-bean-to-json-in-java you can use

```
new JSONObject(result)
```

Not sure how much I trust that, but it's worth a shot

ctsims: This is pretty fraught, because Spring uses the incoming requested content type to determine the serialization of downstream body. Provided an alternative above.

src/main/java/application/FormplayerAuthFilter.java
```diff
@@ -57,7 +59,15 @@
     protected void doFilterInternal(HttpServletRequest req, HttpServletResponse response, FilterChain filterChain)  {
         try {
             FormplayerHttpRequest request = getRequestIfAuthorized(req);
-            filterChain.doFilter(request, response);
+            ContentCachingResponseWrapper responseWrapper = new ContentCachingResponseWrapper(response);
```
ctsims, line 62: @knguyen142 Added a quick wrapper setup here which I've tested does what you'd expect.

It provides a clear copy of the concrete response body in a context where the user details should be available to be tacked onto the request, and the URL should be available for the log.

I think this is the cleanest entry point. My only concern is how this handles the edge cases around a couple of the large payload responses (Form beans in particular), since you get a pretty large response back for every change, but that's more of an architecture question for danny

knguyen142: Thanks @ctsims. I'm playing around with it and do see that it gets the response. However, I'm having issues getting access to the response body in the logging aspect. It appears that when line 67 is run, the logging aspect has already completed. 

Are you suggesting we should move the logging logic from the logging aspect to the intercepting filter chain, in which case we should create another filter?

knguyen142: I found https://github.com/dimagi/formplayer/blob/2cd847a09ac05afc88687d8771b95006ee760d77/src/main/java/application/Application.java#L49, would it be something like that?

knguyen142:
@calellowitz I reverted my changes and went with your approach of using filters instead of aspects. I haven't done the actual logging, but let me know if this is the right approach. 
Also, I can't seem to get `restore_as` from the servlet request and response.

src/main/java/application/FormplayerAuthFilter.java
ctsims: Hey, sorry for the confusion, I was in fact suggesting replacing the aspect driven approach with using the filter I sent over as the entry point for the logging. Will catch up on your changes since now.

ctsims:
@knguyen142 Ah, I see the confusion. Yeah, getting access to the internal state components is a bit tricky, since they are turned into beans for most of the java layer.

In the setup I sent over we do actually [manually extract some of the params](https://github.com/dimagi/formplayer/blob/master/src/main/java/application/FormplayerAuthFilter.java#L134)  to do the initial auth before starting the process. Using the same parsing would be fine for logging as well.

ctsims:
@knguyen142 One other option would be to tag response data directly into a custom place in the response object (Like a "logging info accumulator" or something) and seeing if Spring will let us bubble that back through the request. That would be more breadcrumb-like and allow us more flexibility, but no idea if Spring has a clear way to attach a metadata object to the response lifecycle.

knguyen142:
@ctsims did you get a chance to see the latest commit I did yesterday https://github.com/dimagi/formplayer/pull/736/commits/6aa7ec70105722640d95bc95aa0b9c0aac764d2a. It switches the logging from an aspect to a filter, which according to my googling of spring filters is a reasonable place to put logging logic. I have all the data I need there except `restore_as` which is in a bean?

knguyen142:
> Yeah, getting access to the internal state components is a bit tricky, since they are turned into beans for most of the java layer.

Are you referring to the `restore_as`. I tried to find what this value actually means, in all my test runs locally it is null. Is it vital to log, @dannyroberts? 

So while handling the servlet requests where we have access to the `restore_as`, we would annotate the response object with that and extract it in the filter?

knguyen142:
Ready for review @dannyroberts @dannyroberts 

- Enforced the filter ordering and verified it was working by switching the priority and seeing it fail.
- Extract the `restoreAs` from request body
- actually log 
- added date in ISO8601 format 

Sample log output:
```
{"date":"2020-06-17T19:51Z","responseBody":"","requestBody":"","requestUrl":"http://localhost:8080/submit-all"}
{"date":"2020-06-17T19:51Z","responseBody":"{\"shouldRequestLocation\":false,\"shouldWatchLocation\":false,\"notification\":null,\"title\":null,\"clearSession\":false,\"appId\":null,\"appVersion\":null,\"selections\":null,\"instanceXml\":null,\"status\":\"success\",\"submitResponseMessage\":\"Form successfully saved!\\n\\nYou submitted [this form](/a/test-project/reports/form_data/fda7818b-2585-4733-b8dc-442059fb83fd/).\\n\\nClick to export your [form data](/a/test-project/data/export/custom/form/).\",\"errors\":{},\"nextScreen\":null,\"session_id\":null,\"seq_id\":0}","requestBody":"{\"action\":\"submit-all\",\"answers\":{\"0\":\"world\"},\"prevalidated\":true,\"domain\":\"test-project\",\"username\":\"knguyen@dimagi.com\",\"restoreAs\":null,\"session-id\":\"e4382a61-8d02-413e-8ff2-402f56421142\",\"session_id\":\"e4382a61-8d02-413e-8ff2-402f56421142\",\"debuggerEnabled\":true,\"tz_offset_millis\":-14400000}","requestUrl":"http://localhost:8080/submit-all","projectSpace":"test-project","username":"knguyen@dimagi.com"}
```

Corresponding Submit web request and Response:

Screen Shot 2020-06-17 at 3 55 02 PM


Screen Shot 2020-06-17 at 3 54 13 PM

src/main/java/application/RequestResponseLoggingFilter.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        JSONObject logLine = (new JSONObject())
+                .put("requestUrl", url)
+                .put("requestBody", requestBody)
+                .put("projectSpace", domain)
+                .put("username", user)
+                .put("restoreAs", restoreAs)
+                .put("date", this.currentTimeISO8601())
+                .put("responseBody", responseBody);
```
knguyen142, line 71: Note, you'll find that some of the sample log lines do not have an entry for fields like `username` and `restoreAs`. If these are null it doesn't add it. It looks like we can force a [null value](https://stackoverflow.com/questions/44631604/unable-to-put-null-values-in-json-object) if we want. Also, in my local setup there are always 2 calls for each request, the first has empty fields. Is that normal?

ctsims: restoreAs won't be present in every request. I think username is always populated, because we had some bugs for a while with HMAC auth when it wasn't being sent. Do you have examples of empty username logs? (In theory the server should be ready to accept the request and log the lack of username, even if such requests are malformed, though)

The doublesend is your browser is sending HTTP OPTIONS requests to the server because it's performing a CORS request since you don't have a proxy that's making requests to your local HQ and your formplayer instance look like they are the same terminator.

knguyen142: Ahh, then I have not seen an empty username. But that reminds me, I should check what happens when an error is thrown. My intuition is it will return an error response right away and we won't log the request/response, is that behavior acceptable? 

Also see that we have filter tests, I'll add some for this.

ctsims: It's fine for anything that does not result in data access to not be logged in this layer. It will be sufficient for us to have the raw request logs from the gateway there, since it'll provide the same context formplayer had (someone tried to make a request, we didn't let them).

src/main/java/application/RequestResponseLoggingFilter.java
```diff
@@ -0,0 +1,91 @@
...
+            FormplayerHttpRequest formplayerHttpRequest = (FormplayerHttpRequest) request;
+            domain = formplayerHttpRequest.getDomain();
+            url = new String(formplayerHttpRequest.getRequestURL());
+
+            HqUserDetailsBean userDetailsBean = formplayerHttpRequest.getUserDetails();
+            user = (userDetailsBean != null) ? formplayerHttpRequest.getUserDetails().getUsername() : null;
+        }
+
+        JSONObject logLine = (new JSONObject())
+                .put("requestUrl", url)
```
knguyen142, line 71: Should we have a `type: formplayer` to distinguish from future logs from other sources?

src/main/java/application/RequestResponseLoggingFilter.java
```diff
@@ -0,0 +1,91 @@
...
+                .put("responseBody", responseBody);
+        log.info(logLine);
+
+        // prep the response for the actual outbound connection
+        responseWrapper.copyBodyToResponse();
+    }
+
+    private String currentTimeISO8601() {
+        TimeZone tz = TimeZone.getTimeZone("UTC");
+        DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'"); // Quoted "Z" to indicate UTC, no timezone offset
```
dannyroberts, line 86: can we get second and milli/microseconds here as well?

ctsims:
One quick question - this log feed will be distinct from the current formplayer-spring logs that are tracking the per-request forensics, right?

knguyen142:
> One quick question - this log feed will be distinct from the current formplayer-spring logs that are tracking the per-request forensics, right?

Yes, when we configure the logging configs (sample in the PR description) the logs from this class will dump into a separate file which will be sent to S3.

knguyen142:
@dannyroberts per our conversation on exception handling, https://github.com/dimagi/formplayer/pull/736/commits/83a3e3e13501c6232db654daa4d97f62f544a5db should handle exceptions thrown in the actual logging logic without affecting the request. 

It is split int 3 phases:
1. doBefore: wrapped around a try catch and logs as much info about the request as possible before exception thrown
2. doFilter: calls the next filter in the chain, verified that it already handles exceptions as you assumed.
3. doAfter: wrapped around a try catch and logs as much info about the response. Always flushes whatever is in the log object out and always preps the response for the next filter to not interfere with the request.

src/main/java/application/RequestResponseLoggingFilter.java
dannyroberts: We offlined about this but for the record, what we decided is that these logs would be ending up in their own isolated location dedicated entirely to formplayer logs and not mixed with logs from anywhere else, so no need to specify it on each row

knguyen142:
@dannyroberts Fixed so sensitive data is off by default https://github.com/dimagi/formplayer/pull/736/commits/68655663500a0da3c0306eda8b23320f2d2958d8.

PR #738, opened 2020-06-19
Improved Locking Pattern for Archivable Files

ctsims:
Core PR: https://dimagi-dev.atlassian.net/browse/COV-8

A few things in here:

1) Previously neither the Formplayer process nor the db archive script actually had the ability to clean up stale archive file locks. Archive file locks are now expunged after 5 minutes (this is a long time given the expected lifetime of such a lock)
2) When locks didn't go away on their own, formplayer threads would just hang in an endless loop. It looks like someone didn't understand how thread interrupts work and may have been assuming that calling thread interrupt did something to exit the loop it was in, but it didn't.
3) Fixed the lock ordering to always create a lock before performing the rest of the operations, to prevent race conditions where locks were created after checks
4) [Most Relevant to the Ticket] Ensures that the db file is `touch`ed each time it is loaded (and before a lock can be applied from the archive). This effectively prevents the situation where a db record is loaded at the start of a request but then the DB is archived before the request is ended (which was previously possible to replicate), causing the db file to be written incorrectly unless the request itself is longer than the archiver check window

### DIFF
--- a/src/main/java/exceptions/SqlArchiveLockException.java
+++ b/src/main/java/exceptions/SqlArchiveLockException.java
@@ -1,0 +1,13 @@
+package exceptions;
+
+import java.io.IOException;
+
+/**
+ * Signals that there was a fault attempting to safely acquire or manage
+ * the a lock around a db archive file, which prevented the success of an I/O operation
+ */
+public class SqlArchiveLockException extends IOException {
+    public SqlArchiveLockException(String message) {
+        super(message);
+    }
+}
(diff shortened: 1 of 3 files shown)


ctsims:
Hi @millerdev . I totally agree that this would be a great area to improve test coverage. Unfortunately the existing framework didn't have good coverage, since it involves a confluence of timing, file system, and implicit scopes, but that's true for a lot of parts of Formplayer that need significantly improved mockability

I have been under-investing in building a better framework for being able to test request-level behaviors, I'll make that my next area of focus for formplayer, since a lot of the weaknesses we're finding are fundamentally in handling the request lifecycle

PR #740, opened 2020-06-23
Add a package namespace for formplayer

ctsims:
Formplayer, somewhat ridiculously, has never had an actual default package namespace, which interferes with some autowiring in Spring. 

This PR is a no-op that JUST deals with adding a root package namespace and the associated fallout: changing redis keys for data that was stored as serialized beans with new paths, updating the config file to point to the right package for flyway. 

This has currently been tested locally and has been reviewed line-by-line to make sure that it's not doing anything unexpected. I have confirmed that incomplete form sessions from before the change still load without issue, as do user db's. The only affected models will be wiping the history of commands from the XPath Debugger.


src/main/java/org/commcare/formplayer/annotations/AppInstall.java
```diff
@@ -1,4 +1,4 @@
-package annotations;
+package org.commcare.formplayer.annotations;
```
dannyroberts, line 1: This is going to be the most annoying nitpick in the world, but shouldn't this be `org.commcarehq.formplayer`? I always thought you were supposed to use a namespace for which you own the DNS entry. We don't own commcare.org (right?) but we do own commcarehq.org

ctsims: `org.commcare` is the root namespace for [commcare core](https://github.com/dimagi/commcare-core/tree/master/src/main/java/org/commcare) and for [commcare android](https://github.com/dimagi/commcare-android/tree/master/app/src/org/commcare) (since they were all in use before there was a www.commcarehq.org) so it made the most sense to me to maintain that consistency. It's occasionally relevant for filtering code in profiling contexts or in packaging contexts, so it would add a meaningful effort to use both `org.commcare` and `org.commcarehq`

Owning the DNS entry is more of a "It just feels weird not to" best practice. If we were starting fresh today I'd definitely use commcarehq, but with the historical technical weight it's not really worth it.

dannyroberts: Ah okay gotcha that all makes sense

PR #742, opened 2020-06-24
Fix submit error reporting and single out too-many-requests

dannyroberts:
Easiest to review with w=1

As far as I can tell the `submitResponse.getStatusCode().is2xxSuccessful` clause was never getting run because `submitService.submitForm` with the default error handler will throw an error instead of returning if it's a 4xx or 5xx response; I guess if we got a 3xx response it would have hit that clause.

Now we make it never raise and always return, letting us handle the status code manually.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/SubmitService.java
+++ b/src/main/java/org/commcare/formplayer/services/SubmitService.java
@@ -5,6 +5,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.client.ClientHttpResponse;
 import org.springframework.web.client.DefaultResponseErrorHandler;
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/formplayer/services/SubmitService.java
```diff
@@ -34,6 +35,12 @@
         submitTimer.start();
         try {
             RestTemplate restTemplate = new RestTemplate();
+            restTemplate.setErrorHandler(new DefaultResponseErrorHandler() {
+                @Override
+                protected boolean hasError(HttpStatus statusCode) {
```
ctsims, line 40: Were you intending this to only apply to 429s?

I think that would be 

```
if (statusCode == HttpStatus.TOO_MANY_REQUESTS) {
    return false;
} 

return super.hasError(statusCode);
```

dannyroberts: I think the code as I wrote it here is what I want to do. I kind of put two changes into one commit in retrospect so it's a little confusing. But even aside from the 429 thing, the `if(!submitResponse.getStatusCode().is2xxSuccessful())` bit above wasn't actually catching errors. So this change is to make non-2xx responses get handled by that if clause instead of raising an exception, across the board.

(The second change is then to add sublogic in that handling for the 429 above.)

ctsims: Got it, that makes sense. Thanks for confirming

PR #743, opened 2020-06-24
Fix Spring Boot Wiring for Testing and Consistency 

ctsims:
Apologies for another big PR, but making structural changes has required fixing a lot of latent mistakes in the code that need joint cleanup for things to work after the changes.

Outline (Changes go in order)

1. Moves `Application` from the Nested Package to the Package root

This is a huge change because it makes Spring Boot's AutoConfiguration **actually work.** Previously all of the classes that were tagged as `@Component` and `@Service` etc, weren't actually being wired that way at all.

2. Cleans up Spring object configurations

Now that AutoConfig actually does something, there were a ton of errors in how beans were declared. I've tried hard to not change the semantics of anything - Everything that was a per-request Bean before stayed that way, although a lot of them are now self-declared rather than being defined by the `WebAppContext`

I also removed a lot of Annotations that didn't seem to be in the right place (like declaring @EnableWebMvc on the root Application), but was a bit conservative, so there are probably more annotations that can be cleared.

3. Shift away from a declared RestTemplateBuilder bean 

Now that the app is a properly configured `@SpringBootApplicaiton,` there is a latent autowired `RestTemplateBuilder` bean that conflicts, so the normal pattern is to declare your downstream `RestTemplate` as the bean. 

This change required shuffling how we use the RestTemplate to make sure that it is stateless, such as changing `HqUserDetailsService` to handle 404's in exception rather than a custom catch. 

4. Transition to OkHttp for internal rest requests

Apache has historically had issues with keeping connections open and letting them go stale, especially with CORS requests that have an extra OPTIONS request upfront. We fixed this previously by forcing request to use new restTemplates, but this is deeply discouraged. We've transitioned to OkHTTP on mobile and have been liking it, so this moves Formplayer over to the same core libs.

5. Update the User Detail Service test to use Spring Autowiring

Finally, updates a test that was previously causing us to break the auto-wiring pattern globally in order to work around Spring not setting the test up correctly.

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/WebAppContext.java
+++ b/src/main/java/org/commcare/formplayer/application/WebAppContext.java
@@ -1,17 +1,41 @@
 package org.commcare.formplayer.application;
 
-import org.commcare.formplayer.aspects.*;
 import com.timgroup.statsd.NonBlockingStatsDClient;
 import com.timgroup.statsd.StatsDClient;
+
+import org.commcare.formplayer.aspects.AppInstallAspect;
+import org.commcare.formplayer.aspects.ConfigureStorageFromSessionAspect;
+import org.commcare.formplayer.aspects.LockAspect;
+import org.commcare.formplayer.aspects.LoggingAspect;
+import org.commcare.formplayer.aspects.MetricsAspect;
+import org.commcare.formplayer.aspects.SetBrowserValuesAspect;
+import org.commcare.formplayer.aspects.UserRestoreAspect;
 import org.commcare.formplayer.engine.FormplayerArchiveFileRoot;
-import org.commcare.formplayer.installers.FormplayerInstallerFactory;
-import io.sentry.SentryClientFactory;
-import io.sentry.dsn.InvalidDsnException;
+import org.commcare.formplayer.objects.FormVolatilityRecord;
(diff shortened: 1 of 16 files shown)


src/main/java/org/commcare/formplayer/application/WebAppContext.java
```diff
@@ -1,24 +1,50 @@
...
+import org.springframework.context.annotation.Scope;
+import org.springframework.context.annotation.ScopedProxyMode;
 import org.springframework.context.support.PropertySourcesPlaceholderConfigurer;
 import org.springframework.data.redis.connection.RedisClusterConfiguration;
 import org.springframework.data.redis.connection.RedisStandaloneConfiguration;
 import org.springframework.data.redis.connection.jedis.JedisConnectionFactory;
 import org.springframework.data.redis.core.RedisTemplate;
 import org.springframework.data.redis.core.StringRedisTemplate;
 import org.springframework.data.redis.serializer.GenericJackson2JsonRedisSerializer;
+import org.springframework.http.client.OkHttp3ClientHttpRequestFactory;
```
shubham1g5, line 46: wow,didn't expect spring boot to have a OkHttp client by default.

PR #746, opened 2020-07-07
Fix 4-byte Emoji encoding in Forms

ctsims:
Ticket:
https://dimagi-dev.atlassian.net/browse/SAAS-11089

Bug: Forms displaying incorrect / butchered emoji due to a bug in the kxml parser with UTF-8 chars > 3 bytes long.

CommCare Core Cross-Post:
https://github.com/dimagi/commcare-core/pull/921


dannyroberts:
Is this jar file wanted? https://github.com/dimagi/formplayer/pull/746/files#diff-939b6e418f64fd02789d829114a8c322. I missed this in my first review, but it says kxml 2.3.0 whereas the commit message says 2.4.1

ctsims:
@dannyroberts Removing that lib is expected. We're moving from a packaged binary (`2.3.0`) to an online package (`2.4.1`) so I removed the binary.

dannyroberts:
Got it, the file was removed not added. I'm going to dubiously blame that on my colorblindness

PR #753, opened 2020-08-12
Remove "user" tag from datadog metrics

dannyroberts:
This tag takes on too many values to be useful or cost-effective.


dannyroberts:
Tested this on staging just to make sure I didn't do anything dumb

PR #756, opened 2020-08-27
Add log info for session purge timings

ctsims:
This appears to be a very time consuming part of the restart process. Adding log events to confirm.

### DIFF
--- a/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
+++ b/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
@@ -57,7 +57,7 @@ public class PostgresFormSessionRepo implements FormSessionRepo {
             log.info("Beginning state form session purge");
             long start = System.currentTimeMillis();
             int deletedRows = this.jdbcTemplate.update(deleteQuery);
-            long elapsed = (System.currentTimeMillis() - start) / 1000;
+            long elapsed = System.currentTimeMillis() - start;
             log.info(String.format("Purged %d stale form sessions in %d ms", deletedRows, elapsed));
         } catch (Exception e) {
             // Don't crash for this. Not fatal and prevents start-up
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -54,7 +54,11 @@ public void purgeFormSessions() {
             String deleteQuery = replaceTableName(
                     "delete from %s where custom_safe_cast(dateopened, '2011-01-01'::timestamp) < NOW() - INTERVAL '7 days';"
             );
-            this.jdbcTemplate.execute(deleteQuery);
+            log.info("Beginning state form session purge");
+            long start = System.currentTimeMillis();
+            int deletedRows = this.jdbcTemplate.update(deleteQuery);
```
dannyroberts, line 59: Does `this.jdbcTemplate.update` mean execute and return, or is there some other difference?

src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -54,7 +54,11 @@ public void purgeFormSessions() {
...
+            long elapsed = (System.currentTimeMillis() - start) / 1000;
+            log.info(String.format("Purged %d stale form sessions in %d ms", deletedRows, elapsed));
```
dannyroberts, line 61: looks like elapsed is in seconds (Millis / 1000), so if I'm reading it right either that line should get rid of the `/ 1000` or this line should have `s` instead of `ms`, right?

src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
ctsims: `update` means "Execute the statement and expect an integer response," in jdbc, more or less. There's no specific wrappers for delete methods.

src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
ctsims: oh, whoops, I meant to get rid of the divide, thanks for catching.

ctsims:
@dannyroberts Note that the config file will need to set
```
logging.level.org.commcare.formplayer.repo=INFO
```

For this to be logged

PR #757, opened 2020-08-27
Add Flag for Startup Cleanup of Form Sessions

ctsims:
The current behavior clears out all stale (more than 7 day old) form sessions on startup, which for large production environments has gotten untenable (Taking up to 45 minutes with a broad worker pool and hundreds of thousands of sessions). 

This sets that behavior to be dependent on the app config setting

```
commcare.formplayer.purge.trigger=startup
```

which can be used in smaller environments or on-demand without requiring new deployment to be halted until complete by default.

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -6,3 +6,5 @@ spring.datasource.driver-class-name=${datasource.formplayer.driverClassName}
 spring.datasource.url=${datasource.formplayer.url}
 spring.datasource.username=${datasource.formplayer.username}
 spring.datasource.password=${datasource.formplayer.password}
+
+commcare.formplayer.purge.trigger=never
--- a/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
+++ b/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
@@ -6,6 +6,7 @@ import org.commcare.formplayer.objects.SerializableFormSession;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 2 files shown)


src/main/resources/application.properties
```diff
@@ -5,4 +5,6 @@ spring.flyway.table=schema_version
 spring.datasource.driver-class-name=${datasource.formplayer.driverClassName}
 spring.datasource.url=${datasource.formplayer.url}
 spring.datasource.username=${datasource.formplayer.username}
-spring.datasource.password=${datasource.formplayer.password}
\ No newline at end of file
+spring.datasource.password=${datasource.formplayer.password}
+
+commcare.formplayer.purge.trigger=never
```
dannyroberts, line 10: Does this file define the defaults for when an environment doesn't have a property defined in its own `application.properties` file? If so I think I should add this to commcare-cloud and update each environment with the right setting for it?

ctsims: This defines the defaults, yes. You won't need to add any other settings on cc-cloud just yet, since we aren't clear if this will be a permanent thing, or something you trigger manually. The app will run fine without any external config file changes.

snopoke: Shouldn't the default be 'startup'?

ctsims: @snopoke This is a reasonable open question that I'm happy to chat through.

This is changing the default behavior of formplayer, but it's not obvious to me that people are relying on the old behavior, since we aren't cleaning up things properly with it anyway (menu sessions never get cleared). The old behavior was problematic in a few ways, and doesn't affect runtime of the system as far as I can tell. On prod there are millions of lingering menusessions with no measurable effects, so I doubt any external teams are going to get into trouble until we can find a better cleanup strategy.

My intuition is that it's better to communicate that the default is changing and that we don't recommend changing it back until a better option is available.

PR #758, opened 2020-09-02
change form language

calellowitz:
These are a couple changes required for https://github.com/dimagi/commcare-hq/pull/28463

I am not totally sure if there are other steps I need to take (first formplayer PR), but let me know

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/FormController.java
+++ b/src/main/java/org/commcare/formplayer/application/FormController.java
@@ -122,7 +122,9 @@ public class FormController extends AbstractBaseController{
         FormSession formEntrySession = new FormSession(serializableFormSession, restoreFactory, formSendCalloutHandler, storageFactory);
         formEntrySession.changeLocale(changeLocaleBean.getLocale());
         updateSession(formEntrySession, serializableFormSession);
-        return formEntrySession.getCurrentJSON();
+        FormEntryResponseBean responseBean = formEntrySession.getCurrentJSON();
+        responseBean.setTitle(formEntrySession.getTitle());
+        return responseBean;
     }
 
     @ApiOperation(value = "Answer the question at the given index")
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -122,7 +122,9 @@ public FormEntryResponseBean changeLocale(@RequestBody ChangeLocaleRequestBean c
         FormSession formEntrySession = new FormSession(serializableFormSession, restoreFactory, formSendCalloutHandler, storageFactory);
         formEntrySession.changeLocale(changeLocaleBean.getLocale());
         updateSession(formEntrySession, serializableFormSession);
-        return formEntrySession.getCurrentJSON();
+        FormEntryResponseBean responseBean = formEntrySession.getCurrentJSON();
+        responseBean.setTitle(formEntrySession.getTitle());
```
ctsims, line 126: Suggestion:

In the other responses from the session, this title setting occurs directly in the Bean Generation method 

Example:
https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/session/FormSession.java#L580

I confirmed that `getCurrentJSON` is actually only called by the `changeLocale()` method anyway, so it's fine to nest it in there without concern about side effects on other processes.

ctsims (approved):
looks good, one suggestion on code / package structure

src/main/java/org/commcare/formplayer/application/FormController.java
calellowitz: I did it here because that is what the `answerQuestion` method below does it in the same place. It might be a difference between the form entry ones and the other session ones. Is the other way preferable? I can change this instance

ctsims: Ah, that makes sense, I missed that. I think generally speaking the more response preparation that can be taken care of directly in the method that returns the bean, the better. The Bean seems "Incomplete" to me without it,  so stylistically I'd put it in the `getCurrentJSON` method, but it's a very weak preference and I wouldn't block on it, especially if it's already incompletely managed elsewhere.

PR #760, opened 2020-09-08
Added appearance attribute to json representation of groups

orangejenny:
Needed for https://github.com/dimagi/commcare-hq/pull/28484

### DIFF
--- a/src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
+++ b/src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
@@ -35,7 +35,6 @@ public class PromptToJson {
         parseCaption(prompt, questionJson);
         questionJson.put("help", jsonNullIfNull(prompt.getHelpText()));
         questionJson.put("binding", jsonNullIfNull(prompt.getQuestion().getBind().getReference().toString()));
-        questionJson.put("style", jsonNullIfNull(parseStyle(prompt)));
         questionJson.put("datatype", jsonNullIfNull(parseDataType(prompt)));
         questionJson.put("control", jsonNullIfNull(prompt.getControlType()));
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
```diff
@@ -201,8 +202,10 @@ private static JSONArray parseSelect(FormEntryPrompt prompt) {
 
     //TODO WSP: What the fuck is drew doing XFormPlayer parse_style_info
     // https://github.com/dimagi/touchforms/blob/master/touchforms/backend/xformplayer.py#L400
-    private static JSONObject parseStyle(FormEntryPrompt prompt) {
+    private static JSONObject parseStyle(FormEntryModel model) {
+        FormEntryCaption prompt = model.getCaptionPrompt();
```
ctsims, line 206: This is a bit of a bigger change than it seems, since it makes this method signature dependent on a mutable state in `FormEntryModel` which has an internal walker, rather than an immutable object.

I would push this call upwards, into the `parseQuestionType` method, and generalize this method's signature to receive a `FormEntryCaption` (which is a supertype of a `FormEntryPrompt`) but has the same stateless semantics.

ctsims (changes requested):
Proposed a small change based on my read, but let me know if there's something I'm missing which would lean towards the existing approach instead.

src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
orangejenny: Sounds good, updated in e94281b0421e975bc38f5bc53ac271e9269ece6d - let me know if that's not what you meant.

ctsims (approved):
@orangejenny that's exactly what I was thinking. Thanks!

PR #761, opened 2020-09-10
Kev/scheduled tasks

knguyen142:
https://dimagi-dev.atlassian.net/browse/SAAS-11257

## Summary
Allows purge db query to be run on a schedule in isolation on one node in a cluster. 
Uses the `@Scheduled` annotation and uses shedlock to handle distributed locking to ensure only one runs at a time.

## Changes
- adds shedlock dependency
- refactors the `PostgresFormSessionRepo` to make purge callable 
- adds a scheduled task class to call purge and report metrics to statsd
- Follows [jdbctemplate](FormSessionRepo) instructiions
    - adds a db migration to create a shedlock table 
    - creates a jdbctemplate lockprovider and calls `usingDbTime()` to use the postgres server db time instead of the formplayer server time to ensure synchronization 
- Adds configurations for a [cron expression](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/scheduling/support/CronSequenceGenerator.html), lockAtMostFor and lockAtLeastFor to control the task schedule
    - `0 0 0 * * *` runs on midnight every day

## Testing
Tested locally
With configuration:
```
commcare.formplayer.scheduledTasks.purge.cron=0/1 * * * * *  # Runs every second
commcare.formplayer.scheduledTasks.purge.lockAtMostFor=6s 
commcare.formplayer.scheduledTasks.purge.lockAtLeastFor=6s
```
Which means it is scheduled for every second but there is a lock for exactly 6 seconds. 
output:
```
2020-09-10 11:32:22.16 [scheduling-1] INFO  o.c.f.services.ScheduledTasks - Starting purge scheduled task.
2020-09-10 11:32:22.22 [scheduling-1] INFO  o.c.f.r.impl.PostgresFormSessionRepo - Beginning state form session purge
2020-09-10 11:32:22.31 [scheduling-1] INFO  o.c.f.r.impl.PostgresFormSessionRepo - Purged 0 stale form sessions in 9 ms
2020-09-10 11:32:28.12 [scheduling-1] INFO  o.c.f.services.ScheduledTasks - Starting purge scheduled task.
2020-09-10 11:32:28.14 [scheduling-1] INFO  o.c.f.r.impl.PostgresFormSessionRepo - Beginning state form session purge
2020-09-10 11:32:28.18 [scheduling-1] INFO  o.c.f.r.impl.PostgresFormSessionRepo - Purged 0 stale form sessions in 4 ms
```
Although it is scheduled for every second, it only runs every 6s because of the lock. 

DB entry:
Screen Shot 2020-09-10 at 11 34 51 AM

TODO: Run on staging to verify statsd.

### DIFF
--- a/src/main/java/org/commcare/formplayer/db/migration/V18__init_shedlock.java
+++ b/src/main/java/org/commcare/formplayer/db/migration/V18__init_shedlock.java
@@ -1,0 +1,18 @@
+package org.commcare.formplayer.db.migration;
+
+import java.util.Arrays;
+
+public class V18__init_shedlock extends BaseFormplayerMigration {
+
+    @Override
+    public Iterable<String> getSqlStatements() {
+        return Arrays.asList("CREATE TABLE shedlock (\n" +
+                "    name VARCHAR(64) NOT NULL,\n" +
+                "    lock_until TIMESTAMP NOT NULL,\n" +
+                "    locked_at TIMESTAMP NOT NULL,\n" +
+                "    locked_by VARCHAR(255) NOT NULL,\n" +
+                "    PRIMARY KEY (name)\n" +
+                ")"
+        );
+    }
+}
--- a/src/main/java/org/commcare/formplayer/db/migration/V18__init_shedlock.java
+++ b/src/main/java/org/commcare/formplayer/db/migration/V18__init_shedlock.java
@@ -2,6 +2,10 @@ package org.commcare.formplayer.db.migration;
 
 import java.util.Arrays;
 
+/**
+ * Creates a table to use with shedlock for JdbcTemplate
+ * https://github.com/lukas-krecan/ShedLock/blob/shedlock-parent-4.14.0/README.md#jdbctemplate
+ */
 public class V18__init_shedlock extends BaseFormplayerMigration {
 
     @Override
--- a/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
+++ b/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
@@ -38,18 +38,33 @@ public class PostgresFormSessionRepo implements FormSessionRepo {
     @Autowired
     private JdbcTemplate jdbcTemplate;
 
-    @Value("${commcare.formplayer.purge.trigger}")
-    private String purgeConfig;
-
-    @PostConstruct
-    public void purge() {
-        if ("startup".equals(purgeConfig)) {
-            this.purgeFormSessions();
-        } else {
-            log.debug(String.format("Skipping stale session purge due to current trigger " +
-                    "config '%s'. To enable, set " +
-                    "commcare.formplayer.purge.trigger=startup", purgeConfig));
+    public int purge() {
+        // Modeled on https://stackoverflow.com/a/6730401/2820312
+        int deletedRows = 0;
+        try {
+            String createFuncQuery = "create or replace function custom_safe_cast(text,anyelement) \n" +
+                    "returns anyelement \n" +
+                    "language plpgsql as $$ \n" +
+                    "begin \n" +
+                    "    $0 := $1; \n" +
+                    "    return $0; \n" +
+                    "    exception when others then \n" +
+                    "        return $2; \n" +
+                    "end; $$;";
+            this.jdbcTemplate.execute(createFuncQuery);
+            String deleteQuery = replaceTableName(
+                    "delete from %s where custom_safe_cast(dateopened, '2011-01-01'::timestamp) < NOW() - INTERVAL '7 days';"
+            );
+            log.info("Beginning state form session purge");
+            long start = System.currentTimeMillis();
+            deletedRows = this.jdbcTemplate.update(deleteQuery);
+            long elapsed = System.currentTimeMillis() - start;
+            log.info(String.format("Purged %d stale form sessions in %d ms", deletedRows, elapsed));
+        } catch (Exception e) {
+            // Don't crash for this. Not fatal and prevents start-up
+            log.error("Exception purge form sessions", e);
         }
+        return deletedRows;
     }
 
     @Override
@@ -303,35 +318,6 @@ public class PostgresFormSessionRepo implements FormSessionRepo {
 
     public String replaceTableName(String query){
         return String.format(query, Constants.POSTGRES_SESSION_TABLE_NAME);
-    }
-
-    public int purgeFormSessions() {
-        // Modeled on https://stackoverflow.com/a/6730401/2820312
-        int deletedRows = 0;
-        try {
-            String createFuncQuery = "create or replace function custom_safe_cast(text,anyelement) \n" +
-                    "returns anyelement \n" +
-                    "language plpgsql as $$ \n" +
-                    "begin \n" +
-                    "    $0 := $1; \n" +
-                    "    return $0; \n" +
-                    "    exception when others then \n" +
-                    "        return $2; \n" +
-                    "end; $$;";
-            this.jdbcTemplate.execute(createFuncQuery);
-            String deleteQuery = replaceTableName(
-                    "delete from %s where custom_safe_cast(dateopened, '2011-01-01'::timestamp) < NOW() - INTERVAL '7 days';"
-            );
-            log.info("Beginning state form session purge");
-            long start = System.currentTimeMillis();
-            deletedRows = this.jdbcTemplate.update(deleteQuery);
-            long elapsed = System.currentTimeMillis() - start;
-            log.info(String.format("Purged %d stale form sessions in %d ms", deletedRows, elapsed));
-        } catch (Exception e) {
-            // Don't crash for this. Not fatal and prevents start-up
-            log.error("Exception purge form sessions", e);
-        }
-        return deletedRows;
     }
 
 }
--- a/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
+++ b/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
@@ -42,36 +42,13 @@ public class PostgresFormSessionRepo implements FormSessionRepo {
     private String purgeConfig;
 
     @PostConstruct
-    public void purgeFormSessions() {
-        // Modeled on https://stackoverflow.com/a/6730401/2820312
-        try {
-            String createFuncQuery = "create or replace function custom_safe_cast(text,anyelement) \n" +
-                    "returns anyelement \n" +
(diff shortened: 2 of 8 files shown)


src/main/resources/application.properties
```diff
@@ -7,4 +7,7 @@ spring.datasource.url=${datasource.formplayer.url}
 spring.datasource.username=${datasource.formplayer.username}
 spring.datasource.password=${datasource.formplayer.password}
 
-commcare.formplayer.purge.trigger=never
\ No newline at end of file
+commcare.formplayer.purge.trigger=never
+commcare.formplayer.scheduledTasks.purge.cron=0 0 0 * * *
```
knguyen142, line 11: I'm not familiar with the `resources/application.properties` but do we have to insert this in ansible as well?

ctsims: Hi Kevin,

The settings in `resources/application.properties` are the default config for the app, so as long as they are conservative and function without and external dependencies they don't need to be set in ansible.

It may be the case that we would want to run a different set of parameters for production than we would for a 'standard' install of formplayer (significantly more or significantly less conservative), in which case we'd want to add them to ansible.

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final SimpleDateFormat dateFormat = new SimpleDateFormat("HH:mm:ss");
+
+    @Autowired
+    protected FormSessionRepo formSessionRepo;
+
+    @Autowired
+    private StatsDClient datadogStatsDClient;
+
+    // the default "-" corresponds to Scheduled.CRON_DISABLED
+    @Scheduled(cron= "${commcare.formplayer.scheduledTasks.purge.cron:-}")
```
ctsims, line 33: If you are providing a default here, you won't also need to replicate in `application.properties` as far as I'm aware (and it might be confusing for someone who changes these defaults and misses that the app is also overriding them in the default config)

src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -328,4 +305,33 @@ public String replaceTableName(String query){
...
+                    "        return $2; \n" +
+                    "end; $$;";
+            this.jdbcTemplate.execute(createFuncQuery);
+            String deleteQuery = replaceTableName(
+                    "delete from %s where custom_safe_cast(dateopened, '2011-01-01'::timestamp) < NOW() - INTERVAL '7 days';"
+            );
+            log.info("Beginning state form session purge");
+            long start = System.currentTimeMillis();
+            deletedRows = this.jdbcTemplate.update(deleteQuery);
+            long elapsed = System.currentTimeMillis() - start;
```
ctsims, line 328: Suggestion: This timing might be valuable to report back to DataDog with the event information.

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
```diff
@@ -0,0 +1,48 @@
...
+            lockAtMostFor = "${commcare.formplayer.scheduledTasks.purge.lockAtMostFor:5h}",
+            lockAtLeastFor = "${commcare.formplayer.scheduledTasks.purge.lockAtLeastFor:20m}")
+    public void purge() {
+        log.info("Starting purge scheduled task.");
+        int deletedRows = formSessionRepo.purgeFormSessions();
+        datadogStatsDClient.count(
+                "%s.%s".format(Constants.SCHEDULED_TASKS_PURGE, "deletedRows"),
+                deletedRows
+        );
+        datadogStatsDClient.increment(
```
ctsims, line 44: Danny would have a better idea than myself about the datadog reporting event structure, but the bucketing of tracking this way seems slightly odd to me, since the number of times run will effectively be a constant. It does seem useful to know if the process isn't succeeding, but I don't think this would actually capture that since failures are effectively swallowed in the `purgeFormSessions` call

src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -42,36 +42,13 @@
...
-            } else {
-                log.debug(String.format("Skipping stale session purge due to current trigger " +
-                        "config '%s'. To enable, set " +
-                        "commcare.formplayer.purge.trigger=startup", purgeConfig));
-            }
-        } catch (Exception e) {
-            // Don't crash for this. Not fatal and prevents start-up
-            log.error("Exception purge form sessions", e);
+    public void purge() {
+        if ("startup".equals(purgeConfig)) {
```
ctsims, line 46: I don't think we need to retain this once we replace it with the scheduled runs. This was just a stopgap to leave us an option to trigger the process if needed, so we can remove it with this PR.

src/main/java/org/commcare/formplayer/db/migration/V18__init_shedlock.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Arrays;
+
+public class V18__init_shedlock extends BaseFormplayerMigration {
+
+    @Override
+    public Iterable<String> getSqlStatements() {
+        return Arrays.asList("CREATE TABLE shedlock (\n" +
+                "    name VARCHAR(64) NOT NULL,\n" +
+                "    lock_until TIMESTAMP NOT NULL,\n" +
```
ctsims, line 11: This table schema is an implicit dependency from the new `import net.javacrumbs.shedlock.core.LockProvider` library, right? 

Can we provide some future proofing documentation to outline where the structure came from? It makes me a bit uneasy for someone reviewing this later to not have breadcrumbs to indicate the source of truth for the table schema.

ctsims (changes requested):
Hey Kevin,

This looks great, thanks for the quick turnaround. 

Added a few points / questions which are mostly just style and you can take or leave. I think the "double default" values for the new properties should get changed, though, since right now as a single unit-jar package the inline defaults will never be used, and would mislead someone who changed them.

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
knguyen142: This was mostly to ensure that things are run at most once daily. With the `deletedRows` counter, we might have more than 1 tasks scheduled/reporting at the same time and we wouldn't know how many it came from.

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
knguyen142: Ahh, I didn't know that `resources/application.properties` were the defaults that are included jar. I'll remove those defaults because having the defaults be in code seem much safer, and the default is that this will not run (cron expression `-`).

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
ctsims: Ah! That makes sense as a metric that we want to be keeping track of, thanks for clarifying.

src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -58,21 +55,16 @@ public void purgeFormSessions() {
...
-                        "commcare.formplayer.purge.trigger=startup", purgeConfig));
-            }
+            log.info("Beginning state form session purge");
+            long start = System.currentTimeMillis();
+            deletedRows = this.jdbcTemplate.update(deleteQuery);
+            long elapsed = System.currentTimeMillis() - start;
+            log.info(String.format("Purged %d stale form sessions in %d ms", deletedRows, elapsed));
         } catch (Exception e) {
             // Don't crash for this. Not fatal and prevents start-up
             log.error("Exception purge form sessions", e);
```
ctsims, line 65: One last quick point of feedback - This code is no longer on a critical path that would justify swallowing this exception without doing something with it, so it would be good to make sure it ends up in sentry.

ctsims (approved):

dannyroberts (approved):
Looks good to me too

knguyen142:
@ctsims more logging on elapsed time and sentry logging done. I deployed to staging but was not able to see the datadog logs because we haven't hooked them up to staging?

In any case, modified the `application.properties` on one of the formplayer staging machines to run every midnight and confirmed working:

```
2020-09-12 00:00:00.160 [scheduling-1] INFO  o.c.f.services.ScheduledTasks - Starting purge scheduled task.
2020-09-12 00:00:00.186 [scheduling-1] INFO  o.c.f.r.impl.PostgresFormSessionRepo - Beginning state form session purge
2020-09-12 00:00:00.236 [scheduling-1] INFO  o.c.f.r.impl.PostgresFormSessionRepo - Purged 156 stale form sessions in 49 ms
...
2020-09-13 00:00:00.110 [scheduling-1] INFO  o.c.f.services.ScheduledTasks - Starting purge scheduled task.
2020-09-13 00:00:00.122 [scheduling-1] INFO  o.c.f.r.impl.PostgresFormSessionRepo - Beginning state form session purge
2020-09-13 00:00:00.132 [scheduling-1] INFO  o.c.f.r.impl.PostgresFormSessionRepo - Purged 2 stale form sessions in 9 ms
```

Will create the ansible changes to application.properties.

ctsims:
@knguyen142 One final thing to close out:

Before my previous PR, which temporarily set the default to never, we were purging these sessions on start, so maintainers would expect these sessions to autopurge. To make sure external clusters get back to the default they were on a few weeks ago, maintainers will either need to update their own localsettings or we should reset the default settings to be doing the purge on some basis.

The expectation we generally have is that we make and broadcast a [changelog entry](https://github.com/dimagi/commcare-cloud/blob/502ab88597c1ebebe0bf886775f08c6054b22528/docs/commcare-cloud/deploy.md#local-settings-deploy) anytime someone would need to make a specific change or run commands. My guess is that these cleanup jobs don't take a ton of resources and that running them at midnight server-time would be a very reasonable default (rather than making a changelog) but either way works.

knguyen142:
Got it, I was following the `trigger.purge=never` default, but I can move the [daily midnight cron expression](https://github.com/dimagi/commcare-cloud/pull/4179/files) as a default in code?

ctsims:
Got it, that's my fault for not communicating clearly enough that the `never` default wasn't actually the "default" state, and was more of a stopgap. The 'default a month ago was "purge on startup," so moving the midnight daily cron into default sounds like the right way forward (and bonus that it prevents the need for a localsettings rollout).

knguyen142:
No worries https://github.com/dimagi/formplayer/compare/kev/purge-task-default-schedule?expand=1

PR #765, opened 2020-09-16
Add Source ID to Request Logs

ctsims:
Adds the source IP address redundantly to this log channel for more convenient in-place aggregation.

Note: For production (or any environments behind a load balance or IP carrying proxy), a new directive will need to be added to deploy scripts before this address is processed correctly.


ctsims:
CommCare Cloud Settings companion PR: https://github.com/dimagi/commcare-cloud/pull/4188

orangejenny (approved):

PR #767, opened 2020-09-25
skip fixtures

calellowitz:
This adds a flag to restore requests sent on form completion to ask HQ to avoid sending down fixtures. There will be linked HQ PR that implements that for domains that enable a feature flag.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -327,16 +327,6 @@ public class RestoreFactory {
         }
     }
 
-    public boolean skipFixturesAfterSubmit() {
-        try {
-            return storageFactory.getPropertyManager().skipFixturesAfterSubmit();
-        } catch (RuntimeException e) {
-            // In cases where we don't have access to the PropertyManager, such sync-db, this call
-            // throws a RuntimeException
-            return false;
-        }
-    }
-
     public boolean useAggressiveSyncTiming() {
         try {
             return storageFactory.getPropertyManager().isSyncAfterFormEnabled() &&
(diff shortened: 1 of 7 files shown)


ctsims:
Hey @calellowitz - What's the feature flagging strategy on this? I'm a bit concerned about having the HTTP directive in all of the request urls and then having it be ignored in some domains. Generally we've followed the pattern of putting those kinds of directives into the app definition custom properties. That also makes it possible for HQ to avoid needing to define a Feature Flag for behavior, the behavior can just be dictated by the URL params directly. 

There are a few examples of that around, like

https://github.com/dimagi/formplayer/blob/faabb7f200cffcf60b9211c28b52c461f10ea319/src/main/java/org/commcare/formplayer/application/FormController.java#L236

and

https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/RestoreFactory.java#L318

My guess is that if we expect this to be a temporary state, we should just add another property fetcher like in the RestoreFactory example that pulls `cc-minimal-post-submit-sync` or something, and only adds the url param if both the Property and Boolean are true.

calellowitz:
@ctsims I offlined with Nick, and have added a flag `cc-skip-fixtures-after-submit`. Let me know if this looks better to you. This should also mean that the HQ code doesn't need any gating

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -622,8 +632,8 @@ private HttpHeaders getHmacHeaders(String requestPath) {
         if( asUsername != null) {
             builder.append("&as=").append(asUsername).append("@").append(domain).append(".commcarehq.org");
         }
-        if (skipFixtures) {
-            builder.append("&skip_fixtures=True");
+        if (skipFixtures && skipFixturesAfterSubmit()) {
```
ctsims, line 635: I think for conciseness / consistency it would be good to move this check upstream to the original request in the Form Controller

ctsims (approved):
This looks good to me. One suggestion on code locality / simplicity.

ctsims (approved):
Solid

PR #768, opened 2020-10-13
Kev/point core to kev timezone

knguyen142:
https://dimagi-dev.atlassian.net/browse/SAAS-11203

If the browser passes up the timezone information, the browser timezone provider uses that timezone ID string as the timezone so the shared commcare core code that does date functions can have tz context.

## Flow
[Browser](https://github.com/dimagi/commcare-hq/pull/28647) sends up IANA tz (e.g. "America/New_York"). ---> 
[Formplayer](https://github.com/dimagi/formplayer/pull/768) uses the tz from the browser as a info for its timezone provider --> 
[the shared commcare-core ](https://github.com/dimagi/commcare-core/pull/935)code that does date maniplations uses the timezone provider.

## Note
This change should not be coupled to the commcare-hq PR nor the commcare-core PR, because without the tz from the browser it should fallback to the tz offset. 
If we rely on timezones, we need to update the java code or periodically run the timezone updater tool https://www.oracle.com/java/technologies/javase/tzupdater-readme.html.

## Testing
The timezone IDs should be in the IANA format. There is code to log any discrepancy between the timezone ID passed from the browser and the offset passed in by the browser, which sends a sentry exception so we can get an overview of any discrepancies. 

## Final Result
(deployed on staging)
Screen Shot 2020-10-08 at 7 16 45 PM
Note: -5 from EST when current offset is -4 in EDT.

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/AuthenticatedRequestBean.java
+++ b/src/main/java/org/commcare/formplayer/beans/AuthenticatedRequestBean.java
@@ -21,6 +21,7 @@ public class AuthenticatedRequestBean {
     private String restoreAsCaseId;
 
     private int timezoneOffsetMillis = -1;
+    private String timezoneFromBrowser = null;
 
     public String getUsername() {
         return username;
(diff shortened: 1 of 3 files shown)


ctsims (changes requested):
I think the submodule update should be pointing to https://github.com/dimagi/commcare-core/pull/935/commits/8418827d2a0fc8fb9dc11b27914dc714fc9b199d , which will be the new formplayer head after https://github.com/dimagi/commcare-core/pull/935 is pulled. 

Looks good to me otherwise, thanks for going the extra mile on this.

knguyen142:
Awesome, I merged https://github.com/dimagi/commcare-core/pull/935 and pointed the submodule to formplayer head https://github.com/dimagi/commcare-core/commits/formplayer.

ctsims (approved):

PR #771, opened 2020-10-20
use minimal entity screen for case lists

calellowitz:
This is the formplayer half of https://github.com/dimagi/commcare-core/pull/937
Opening for review, but it cannot be merged until that is merged and the commcare-core reference in the formplayer repo is updated.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -172,6 +172,7 @@ public class MenuSessionRunnerService {
                                                          String searchText,
                                                          int sortIndex) throws Exception {
         BaseResponseBean nextResponse;
+        boolean needsDetail;
         // If we have no selections, we're are the root screen.
         if (selections == null) {
             return getNextMenu(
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -185,15 +186,19 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
         NotificationMessage notificationMessage = null;
         for (int i = 1; i <= selections.length; i++) {
             String selection = selections[i - 1];
-            boolean gotNextScreen = menuSession.handleInput(selection);
+
+            // minimal entity screens are only safe if there will be no further selection
```
ctsims, line 190: I understand why we would need to keep the screen for a case detail, but what accounts for the 'no further selection' limitation?

ctsims (commented):
Left one question for you on the implementation structure.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
calellowitz: To the extent I can tell, app navigation works by feeding selections as input to CLI screens (thats the loop this line is in). So for anything where a case is part of the input (like when entering a form), the `EntityScreen` needs to be able to handle case selection, and at least currently that requires the two components that I skipped in the minimal version.

ctsims: ah, I see. This is about the future requests that are 'playing back' the user choices that we talked about as one of the other potential targets for performance improvement. That does sound right. Thanks for explaining

ctsims (approved):

PR #772, opened 2020-10-21
update submodule reference

calellowitz:


orangejenny (approved):
This is one of the few cases where I think self-merging without review is totally fine. It's nice to have the PR just so tests run - not sure if that applies for here, it does for some submodules.

calellowitz:
It looks we do have a formplayer travis build, so I will wait on that.

PR #775, opened 2020-10-23
Ability to require search for a case list

orangejenny:
See https://github.com/dimagi/commcare-hq/pull/28677

This PR checks case list details for the `requireSearch` flag before fetching entities.

It also adds the `requireSearch` flag and the search text to the response so that web apps can display appropriate messaging.

Depends on https://github.com/dimagi/commcare-core/pull/939

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
@@ -108,7 +108,7 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
         this.requireSearch = requireSearch;
     }
 
-    @JsonSetter(value = "searchText")
+    @JsonGetter(value = "searchText")
     public String getSearchText() {
         return this.searchText;
     }
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
@@ -17,6 +17,7 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
 
     private EntityDetailResponse[] entityDetailList;
     private boolean isPersistentDetail;
+    private boolean requireSearch;
 
     public EntityDetailListResponse() {}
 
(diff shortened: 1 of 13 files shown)


src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
```diff
@@ -73,7 +75,12 @@ public EntityListResponse(EntityScreen nextScreen,
             }
             entities = processEntitiesForCaseDetail(detail, reference, ec, neededDatum);
         } else {
-            Vector<TreeReference> references = nextScreen.getReferences();
+            Vector<TreeReference> references = new Vector<TreeReference>();
+            this.requireSearch = nextScreen.getShortDetail().doesRequireSearch();
+            this.searchText = searchText;
+            if (!this.requireSearch || this.searchText != null && !"".equals(this.searchText)) {
+                references = nextScreen.getReferences();
```
ctsims, line 82: Might be worth talking with @calellowitz  about his recent changes.

I believe this current structure will still end up doing a huge amount of the work of the work, because the screen itself will have already generated the reference set before it's requested here. You likely want to bubble up the logic for where the search requirement comes in into the screen itself (which will let you suppress the majority of the work of filtering the reference set and loading fields from I/O)

ctsims (commented):
This pattern likely doesn't defer a lot of the work intended, left a note to coordinate with Cal about his recent changes.

If you all are fine with still doing the initial reference set expansion, let me know

src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
```diff
@@ -95,4 +97,24 @@ public boolean getPersistentDetail() {
...
+    public boolean getRequireSearch() {
+        return requireSearch;
+    }
+
+    @JsonSetter(value = "requireSearch")
+    public void setRequireSearch(boolean requireSearch) {
+        this.requireSearch = requireSearch;
+    }
+
+    @JsonSetter(value = "searchText")
```
snopoke, line 111:
```suggestion
    @JsonGetter(value = "searchText")
```

src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
```diff
@@ -95,4 +97,24 @@ public boolean getPersistentDetail() {
...
+    public String getSearchText() {
+        return this.searchText;
+    }
+
+    @JsonGetter(value = "searchText")
```
snopoke, line 116:
```suggestion
    @JsonSetter(value = "searchText")
```

snopoke (changes requested):

orangejenny:
Rebased on top of https://github.com/dimagi/formplayer/pull/780

No need to re-review yet.

orangejenny:
Closing in favor of https://github.com/dimagi/formplayer/pull/783

PR #777, opened 2020-10-28
Add a function to formplayer to encrypt a string

solleks:
Jonathan asked me to add a function to formplayer to implement [this specification](https://docs.google.com/document/d/1xpymVJm27Hq4rl1jatb049oVr_Dto27eahaKBwD76Gk/edit?usp=sharing). Here is my initial version, and I made a copy of the specification to [document the implementation](https://docs.google.com/document/d/1OljNDYf-Z9nARKn5ww-ZZf2tWUIknHoEVJ__9Oq7YIo/edit?usp=sharing).

Essentially, it accepts a message string and an encoded secret key, performs AES encryption using the GCM mode and returns the encoded message along with an initialization vector needed to decode it. It is intended to allow other algorithms to be used in the future, but only AES is implemented for now and certain encryption options are hardcoded that the message receiver will need to accomodate.

I'm new to the formplayer and also am not a security expert, so any feedback on better ways to do this would be appreciated.

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/UtilController.java
+++ b/src/main/java/org/commcare/formplayer/application/UtilController.java
@@ -167,4 +167,12 @@ public class UtilController extends AbstractBaseController {
 
         return reporter.generateJSONReport();
     }
+
+    @ApiOperation(value = "Encrypt a string with a given algorithm and key")
+    @RequestMapping(value = Constants.URL_ENCRYPT_STRING, method = RequestMethod.POST)
+    public EncryptStringResponseBean encryptString(@RequestBody EncryptStringRequestBean encryptStringBean) throws Exception {
+        EncryptStringResponseBean responseBean = new EncryptStringResponseBean(encryptStringBean);
+        return responseBean;
+    }
+
 }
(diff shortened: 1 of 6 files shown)


fyniac:
@calellowitz @orangejenny tagging you both for review

ctsims:
Hi @solleks 

Apologies, I think there was a significant miscommunication on this.

What Jon was trying to ask for was a CommCare XPath function like [this](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xpath/expr/XPathRegexFunc.java) which can be utilized in-engine in a form context to create the output.

solleks:
I see. Well, I should be able to reuse most of the code and fit this into an XPath function. I'll get started on that.

ctsims:
@solleks Thanks, and apologies for the miscommunication. If helpful I could make a form that would outline what I'd guess Jon is thinking which you could test against for requirements.

solleks:
Yes, I think having a form would be helpful. Thanks.

ctsims:
@solleks I think this attached form is in-line with what jon was looking for.

No strong feelings about most of the contract, btw. I'm presuming here that it's easiest for the encrypter to generate the IV internally randomly, since there's not a great way in the form to generate 128 random bits and encode them cleanly. Everything else can come in from existing mechanisms. Let me know if this helps or if you have any other questions.

[Encryption example form.txt](https://github.com/dimagi/formplayer/files/5454857/Encryption.example.form.txt)

solleks:
Thanks for the form. I've made a new PR in the commcare-core repo (https://github.com/dimagi/commcare-core/pull/944), which implements encryption as an XPath function.

I'll close this one.

Issue #741, opened 2020-06-23
Temporary Archive File Misalignment after Restart in Clustered mode


ctsims:
In a multi-worker cluster, the fact that the [Archive File Cache](https://github.com/dimagi/formplayer/blob/master/src/main/java/engine/FormplayerArchiveFileRoot.java#L45) isn't keyed at all by worker (and thus by filesystem) means that after a restart, if another worker has added a different key for the archive, the local request will fail with

```
org.javarosa.core.reference.InvalidReferenceException: Error deriving reference with exception jr://archive/6776c4dd73c94fdf908e7d6561591104/default/app_strings.txt.
        at engine.FormplayerArchiveFileRoot.derive(FormplayerArchiveFileRoot.java:51)
        at org.javarosa.core.reference.ReferenceManager.DeriveReference(ReferenceManager.java:134)
        at org.javarosa.core.reference.ReferenceManager.DeriveReference(ReferenceManager.java:88)
        at org.javarosa.core.reference.ReferenceDataSource.getLocalizedText(ReferenceDataSource.java:45)
```

for the first request until the archive is re-fetched. The Archive File Root process also does not ever attempt to successfully re-link any archives which are recovered successfully from the redis cache into the native `guidToFolderMap` cache if it DOES succeed in re-deriving it, whcih seems intentional

PR #779, opened 2020-10-28
corrupted ccz handling

knguyen142:
https://dimagi-dev.atlassian.net/browse/SAAS-11333

If the downloaded ccz is corrupted, we still save it in memory and continually hit the same exception.

This handles it so that if we hit those exceptions, we clear the ccz from memory so subsequent requests will try to redownload. 

## Testing

Tested locally.

1. modified cchq the ccz file that is downloaded to return scratch (in [cchq](https://github.com/dimagi/commcare-hq/blob/153e8953ff28a064c4d84ff69e3e1bd0bb397b1a/corehq/ex-submodules/soil/__init__.py#L113), instead of returning the contents we return a string like `x`). 
2. Try to start webapps and it will continually try to download the ccz
Screen Shot 2020-10-28 at 4 23 23 PM
3. Fix the cchq code so we return the actual contents
4. retry webapps, and it successfully downloads and runs

## Issues
This can download a lot of ccz's if cchq continually returns invalid files
Screen Shot 2020-10-28 at 4 24 02 PM

I don't know if all paths to those exceptions should have the ziparchive file be removed from memory. I also couldn't test the `UnresolvedResourceException` exception path, which is what was the original exception in the issue.

Depends on `https://github.com/dimagi/commcare-core/pull/943`, do not merge until that is in this PR.

### DIFF
--- a/src/main/java/org/commcare/formplayer/engine/FormplayerArchiveFileRoot.java
+++ b/src/main/java/org/commcare/formplayer/engine/FormplayerArchiveFileRoot.java
@@ -47,4 +47,11 @@ public class FormplayerArchiveFileRoot extends ArchiveFileRoot {
             throw new InvalidReferenceException(String.format("Error deriving reference with exception %s.", guidPath), guidPath);
         }
     }
+
+    @Override
+    public String removeArchiveFile(String appId) {
+        String mGUID = super.removeArchiveFile(appId);
+        redisTemplate.delete(String.format("formplayer:archive:%s", mGUID));
+        return mGUID;
+    }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/formplayer/engine/FormplayerArchiveFileRoot.java
```diff
@@ -47,4 +47,11 @@ public Reference derive(String guidPath) throws InvalidReferenceException {
             throw new InvalidReferenceException(String.format("Error deriving reference with exception %s.", guidPath), guidPath);
         }
     }
+
+    @Override
+    public String removeArchiveFile(String appId) {
+        String mGUID = super.removeArchiveFile(appId);
+        redisTemplate.delete(String.format("formplayer:archive:%s", mGUID));
```
knguyen142, line 54: Look into deleting the file on disk.

ctsims:
@knguyen142 One weakness in this caching code that I found a while ago is that it appears to re-use the same redis cache key across all of the servers despite the file being local https://github.com/dimagi/formplayer/issues/741. One downside to this is that removing the key does remove the key for every other server as well, which may or may not be worth worrying about. I'm not actually sure the redis keyed cache does very much (since it kind of can't) v. the in-memory cache.

knguyen142:
I decided to close this because we cannot reproduce the underlying issue, so these changes might end up doing more harm than good

PR #780, opened 2020-11-02
Session cache

calellowitz:
This PR has two components to hopefully speed up menu navigation.
1) Defers initialization of EntityScreens until necessary
2) Updates the menu session with known valid selections (those previously evaluated) without evaluating the input at the screen level.

It tracks valid selections in redis, keying off a combination of username, restore as user, and domain, and clearing the cache on restore, app install, and form submission.

I don't love the changes to the code organization, especially pushing the initialization and auto advance logic out of the centralized methods (`createFreshEntityScreen` and `getNextScreen`) in `MenuSession.java` and into each of the places that need it, but it was the only way I could figure out how to do it.

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/FormController.java
+++ b/src/main/java/org/commcare/formplayer/application/FormController.java
@@ -14,6 +14,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.data.redis.core.RedisTemplate;
+import org.springframework.data.redis.core.SetOperations;
 import org.springframework.data.redis.core.ValueOperations;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
@@ -66,6 +67,7 @@ import org.commcare.formplayer.session.FormSession;
 import org.commcare.formplayer.session.MenuSession;
 import org.commcare.formplayer.util.Constants;
 import org.commcare.formplayer.util.SimpleTimer;
+import org.commcare.formplayer.util.UserUtils;
 import org.springframework.web.client.HttpClientErrorException;
 
 /**
@@ -95,6 +97,12 @@ public class FormController extends AbstractBaseController{
     @Resource(name="redisVolatilityDict")
     private ValueOperations<String, FormVolatilityRecord> volatilityCache;
 
+    @Autowired
+    private RedisTemplate redisSetTemplate;
+
+    @Resource(name = "redisSetTemplate")
+    private SetOperations<String, String> redisSessionCache;
+
     @Value("${commcarehq.host}")
     private String host;
 
@@ -105,7 +113,7 @@ public class FormController extends AbstractBaseController{
     @RequestMapping(value = Constants.URL_NEW_SESSION, method = RequestMethod.POST)
     @UserLock
     @UserRestore
-        public NewFormResponse newFormResponse(@RequestBody NewSessionRequestBean newSessionBean,
+    public NewFormResponse newFormResponse(@RequestBody NewSessionRequestBean newSessionBean,
                                            @CookieValue(name=Constants.POSTGRES_DJANGO_SESSION_ID, required=false) String authToken) throws Exception {
         String postUrl = host + newSessionBean.getPostUrl();
         return newFormResponseFactory.getResponse(newSessionBean, postUrl);
@@ -207,6 +215,8 @@ public class FormController extends AbstractBaseController{
 
                 // Only delete session immediately after successful submit
                 deleteSession(submitRequestBean.getSessionId());
+                String cacheKey = UserUtils.getFullUserDetail(submitRequestBean.getUsername(), submitRequestBean.getRestoreAs(), submitRequestBean.getDomain());
+                redisSessionCache.getOperations().delete(cacheKey);
                 restoreFactory.commit();
 
             }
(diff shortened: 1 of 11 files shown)


calellowitz:
This has now passed QA, so is ready for review @ctsims / @orangejenny

orangejenny:
> It tracks valid selections in redis, keying off a combination of username, restore as user, and domain, and clearing the cache on restore, app install, and form submission.

If this keyed on formplayer session id, could you then only clear on form submission? Not sure if it would be better, just curious what keys & cache lifecycles you considered.

calellowitz:
> > It tracks valid selections in redis, keying off a combination of username, restore as user, and domain, and clearing the cache on restore, app install, and form submission.
> 
> If this keyed on formplayer session id, could you then only clear on form submission? Not sure if it would be better, just curious what keys & cache lifecycles you considered.

Which session id? The only persistent session ID that I know of is strictly for form sessions, but if there is a menu session and it is properly recreated on sync and install, that does sound nicer.

orangejenny:
@calellowitz I'm almost certainly thinking of the form session ID, nevermind.

src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -207,6 +215,8 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
 
                 // Only delete session immediately after successful submit
                 deleteSession(submitRequestBean.getSessionId());
+                String cacheKey = UserUtils.getFullUserDetail(submitRequestBean.getUsername(), submitRequestBean.getRestoreAs(), submitRequestBean.getDomain());
```
ctsims, line 218: I think you would be better off isolating / abstracting these cache management operations in the **RestoreFactory** or in the UserSqlSandbox, which is retrieved from the restore factory

Those dependencies are autowired everywhere that you are currently managing them, and already have access to the key components in a more centralized way, which reduces the risk of drift between how they are established in different contexts.

That would also let you centralize a few of the cache clearing operations, although based on the size of the cache I think it might make more sense to encode DB state in the key rather than clearing it deliberately

src/main/java/org/commcare/formplayer/application/MenuController.java
```diff
@@ -140,6 +150,9 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
             throw new RuntimeException("Could not find case with ID " + detailSelection);
         }
 
+        String cacheKey = UserUtils.getFullUserDetail(sessionNavigationBean.getUsername(), sessionNavigationBean.getRestoreAs(), sessionNavigationBean.getDomain());
+        String cacheValue = String.join("|", selections);
```
ctsims, line 154: These `String.join` calls should be abstracted into a call somewhere that has clearer semantic meaning (IE: getSessionCacheKey(selections)

I wonder if it would be worth giving the "selections" array here its own business logic object (and potentially the other items above, IE: 'commitSelections' 'detailSeleciton' etc) given how many places we seem to be making assumptions about multiple arrays of strings and what their semantic meaning is. 

Depending on where that object comes from in the lifecycle, you might be able to build more robust cache keys around it in the future.

src/main/java/org/commcare/formplayer/services/InstallService.java
```diff
@@ -81,6 +91,7 @@
             engine.initEnvironment();
             installTimer.end();
             installTimer.record();
+            redisSessionCache.getOperations().delete(storageFactory.getUsername());
```
ctsims, line 94: I'm confused, doesn't this cache key require a domain and a potential "as" name, or is his equivalent somehow to that call?

ctsims (approved):
Hey Cal,

This approach looks right to me based on what we discussed, although I'm not super comfortable with the degree to which it's possible to confirm that the cache management is handled uniformly.

One thing is minor cleanups of just unifying logic into places where future code couldn't cause problems (IE: Every instance of `string.join()` being replaced by a "createCacheKey" call)

The other is that I think it might be possible to simplify the cache key management significantly due to the fact that the DB holds all of the relevant state. I think if we unified all of the commit-points for the app (I believe there are only 3) you could do something like cycle a random guid 'db state token' after each commit, and make that token part of the cache key itself. Alternatively you could just clear the cache after each commit, although that also requires you to catch things like "the database was deleted in the background and re-instantiated." 

Overall, though, given that this commit requires the introduction of *some* trigger which needs to be uniquely fired any time the DB state is changed, there's no reason from my perspective that we shouldn't make that a specific hook (probably in the RestoreFactory) and introduce the cache clearing call inside of that hook, rather than force users to specify the cache clearing as the action they want to trigger.

calellowitz:
@ctsims I reorganized the code and made the updates we discussed. Let me know if this is what you had in mind.

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -192,21 +192,33 @@ private String getReferenceToLatest(String appId) {
...
-        return handleInput(input, true);
+        return handleInput(input, true, false);
     }
 
     /**
      * @param input the user step input
      * @return Whether or not we were able to evaluate to a new screen.
      */
-    public boolean handleInput(String input, boolean needsDetail) throws CommCareSessionException {
+    public boolean handleInput(String input, boolean needsDetail, boolean confirmed) throws CommCareSessionException {
```
ctsims, line 202: Can you put the 'confirmed' param in the method doc string above along with its purpose and behavior?

ctsims (approved):
This looks good to me now. Could use doc strings in a few places since the behavior is so specific.

PR #783, opened 2020-11-16
Added auto launching of actions

orangejenny:
https://dimagi-dev.atlassian.net/browse/USH-366

Adds support for actions marked as "auto_launch." When a case list with such an action is viewed, the user will not be shown the case list and will instead proceed as though they clicked on the action.

Not especially proud of this code, especially the flag being passed through five layers. Auto-launching case claim only works if the case list is the final result of the `advanceSessionWithSelections` loop that iterates over the current selection. If the action is auto-launched after the case id has been added to selections, the user gets perpetually kicked back to the case claim screen instead of advancing on to the form.

Depends on https://github.com/dimagi/commcare-core/pull/949

HQ changes, which depend on this PR but can be deployed independently, are in https://github.com/dimagi/commcare-hq/pull/28788

[QA ticket](https://dimagi-dev.atlassian.net/browse/QA-2109)

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -203,7 +203,8 @@ public class MenuSessionRunnerService {
             // minimal entity screens are only safe if there will be no further selection
             // and we do not need the case detail
             needsDetail = detailSelection != null || i != selections.length;
-            boolean gotNextScreen = menuSession.handleInput(selection, needsDetail, confirmed);
+            boolean allowAutoLaunch = i == selections.length;
+            boolean gotNextScreen = menuSession.handleInput(selection, needsDetail, confirmed, allowAutoLaunch);
             if (!gotNextScreen) {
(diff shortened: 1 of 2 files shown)


ctsims (approved):
Can you add method level docstrings for the public methods with the new signature explaining what the difference is?

Also worth making sure noting here collides with @calellowitz 's recent changes.

orangejenny:
Rebased on top of https://github.com/dimagi/formplayer/pull/780 which required moving the action-examining logic into the constructor.

shubham1g5:
It would be good to check if this requires a version bump for [SQLITE_DB_VERSION](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/util/Constants.java) since the underlying suite model has changed. I am not sure if Formplayer reads suite file from a serialized stream or xml.

orangejenny:
@shubham1g5 Thanks for flagging, we discussed in #formplayer on slack and concluded this doesn't require bumping the db version.

orangejenny:
@ctsims I'll add the requested comments in a followup.

PR #787, opened 2020-12-07
negative index

calellowitz:
This implements the formplayer component of https://github.com/dimagi/commcare-core/pull/957


ctsims (approved):
This looks essentially right (relative to the feedback in the other PR). Did you see whether there were tests you could use as a template for the new functions?

shubham1g5:
We are currently holding merge of core's `master` into `formplayer` because of the ramifications caused by bumping the formplayer DB version. Since this PR is anyway doing it, it would be helpful to perform that pending commcare-core merge before we merge this PR so as to limit the no of times we end up bumping the db versions.

ctsims (approved):

PR #791, opened 2020-12-10
Auto launching of actions, round 2

orangejenny:
https://github.com/dimagi/formplayer/pull/783/ "works" by [executing the autolaunch action](https://github.com/dimagi/commcare-core/blob/1c8d2a04c1219d99e9afd923d2556f3fe2c02e7c/src/cli/java/org/commcare/util/screen/EntityScreen.java#L87-L91) inside the `EntityScreen` constructor. Since that action doesn't get a session frame the same way a real user action would, there are downstream problems, notably the inability to actually claim cases.

Instead of figuring out how to make a better fake user action, this PR changes formplayer's case list response to include a flag telling web apps to immediately make an additional request to case claim. This should mimic real behavior better. Most of the logic from https://github.com/dimagi/formplayer/pull/783/ is retained so that if the case list does have an autolaunch action, it'll still skip unnecessary processing.

Depends on https://github.com/dimagi/commcare-core/pull/960

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
@@ -60,6 +60,9 @@ public class EntityListResponse extends MenuBean {
         EntityDatum neededDatum = (EntityDatum) session.getNeededDatum();
         EvaluationContext ec = nextScreen.getEvalContext();
 
+        this.actions = processActions(nextScreen.getSession());
+        this.autolaunch = processAutolaunch(nextScreen.getSession());
+
         // When detailSelection is not null it means we're processing a case detail, not a case list.
         // We will shortcircuit the computation to just get the relevant detailSelection.
         if (detailSelection != null) {
@@ -73,6 +76,9 @@ public class EntityListResponse extends MenuBean {
                 detail = longDetails[0];
             }
             entities = processEntitiesForCaseDetail(detail, reference, ec, neededDatum);
+        } else if (this.autolaunch != null) {
+            // This is a case list that the UI is going to skip, so don't bother processing entities
+            entities = new EntityBean[0];
         } else {
             Vector<TreeReference> references = nextScreen.getReferences();
             List<EntityBean> entityList = processEntitiesForCaseList(detail, references, ec, searchText, neededDatum, sortIndex, isFuzzySearchEnabled);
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
```diff
@@ -278,6 +280,27 @@ private static EntityBean processEntity(Detail detail, TreeReference treeReferen
...
+        for (Action action: actions) {
+            if (action.isAutoLaunching()) {
+                ret = "action " + index;
+            }
+            index = index + 1;
+        }
+        return ret;
+    }
+
+    private static Vector<Action> getActions(SessionWrapper session) {
```
calellowitz, line 296: This is mildly confusing since it shares the name with a normal getter of the `actions` attribute in line 320.

calellowitz: This is called a few times, is it cheap, or worth saving somewhere?

src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
```diff
@@ -278,6 +280,27 @@ private static EntityBean processEntity(Detail detail, TreeReference treeReferen
         return ret;
     }
 
+    private static String processAutolaunch(SessionWrapper session) {
+        Vector<Action> actions = getActions(session);
+        String ret = null;
+        int index = 0;
+        for (Action action: actions) {
```
calellowitz, line 287: is it possible for there to be 2 autolaunch actions? this looks like it would just return the second if thats the case. Is that ok?

calellowitz (approved):
Small questions, but nothing blocking

src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
orangejenny: The XML spec does support multiple autolaunch actions, but they don't really make sense so I think this is fine. In retrospect, it might have made more sense to add an `autolaunch-action="0"` attribute to the detail instead of a flag on actions themselves. I'm not going to worry about that right now, but might think about it before removing the new feature flag.

src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
orangejenny: Pretty sure it's cheap. I'll rename it.

PR #793, opened 2020-12-18
Updated commcare-core ref

stephherbers:
updated ref after this [PR](https://github.com/dimagi/commcare-core/pull/963)


orangejenny:
```
org.commcare.formplayer.tests.CaseClaimTests > testQueryScreen FAILED
    org.springframework.web.util.NestedServletException: Request processing failed; nested exception is org.javarosa.core.util.NoLocalizedTextException: The Localizer could not find a definition for ID: case.search.title in the 'en' locale.
```

My guess is that tests are using apps that don't have access to the new string. `CaseClaimTests` has a reference to [this file](https://github.com/dimagi/formplayer/blob/master/src/test/resources/restores/caseclaim.xml) which is in a test data directory. Jumping up a few levels to `src/test/resources`, there's a `formplayer_translatable_strings.txt` file with a couple of translations. I'd add your new one in there and see if it fixes tests.

stephherbers:
Closing because [this PR](https://github.com/dimagi/commcare-core/pull/963) was reverted. Decided to leave the try/catch in.

PR #799, opened 2020-12-28
Case claim Itemset Support 

shubham1g5:
**Response Contract** 

Enhances `displays` in response to Case Search request as - 
````
"type": "query",
"displays": [{
      "text": "State",
      "id": "state",
      "input": "select1",
      "value": "",
      "itemsetChoices": ["Rajasthan", "Karnataka"]
    }
]}
````

This adds 3 new attributes to the top level `displays` item. 
1. `value` : This is used to convey the default value for the display fields 
2. `input`:  Type of the display field. `null` or empty value indicates a simple text field as now while `select1` is used to indicate a dropdown field
3. `itemset`:Only valid when `input` is `select1`, indicates the choices calculated by Formplayer engine for the display field.   

In order to refresh `itemset` choices as the value of case claim fields changes, HQ will need to fire a request after every change in the value of a case claim field. Formplayer will take into account the new values set by user and respond with the new set of `itemset` choices per user inputs. 

**Request Contract**

Since in this design, there are multiple requests from HQ to formplayer as user is filling the case claim search fields , we need to add a new variable "do_query" to case claim request as an indication of "Search" button click on Case claim screen.   Earlier Formplayer would just assume that any request containg the `query_dictionary` is the result of the search button click which is no longer true with this PR. 

For `select1` fields, HQ should output the index of the selection in `query_dictionary` when responding with the choices filled by user i.e. if a particlar search field defines choices as ` "itemsetChoices": ["Rajasthan", "Karnataka"]` , in response to a selection of option "Karnataka", HQ should only output the index `1` of the choice in `query_dictionary` as the answer. Formplayer would then take this index and get the corresponding lookup value for this index. In case of no selection, the answer can just be `null`. Vice versa to convey the value for a `select1` field in response, Formplayer will pass the index in the passed `itemset` array as the `value`. 

Hq side of this is still pending and created this ahead of time to get feedback on the approach here. 

cross-request: https://github.com/dimagi/commcare-core/pull/967 , https://github.com/dimagi/commcare-core/pull/968

### CODE REVIEW

**Comment by orangejenny** [comment]
Haven't fully absorbed this yet, but can you say more about this?

> In order to refresh itemset choices as the value of case claim fields changes, HQ will need to fire a request after every change in the value of a case claim field. Formplayer will take into account the new values set by user and respond with the new set of itemset choices per user inputs.

Is this asking HQ to fire a formplayer request as the user types more characters into a search field on the search screen?

**Comment by shubham1g5** [comment]
> Is this asking HQ to fire a formplayer request as the user types more characters into a search field on the search screen?

@orangejenny yes. Similar to what we do when we fill up a form in Formplayer. The reason being a search field can define an itemset instance which is dependent on the output of another search field (location heirarchies for example). To support this HQ will need to ask Formplayer to refresh the itemset choices when the value of input field changes. If it's easier we can decide to only do this on changes with dropdown fields assuming that we only support inter-depency of fields for the dropdowns and not for free texts field.

**Comment by shubham1g5** [comment]
Seprated out more time sensitive parts of this PR in https://github.com/dimagi/formplayer/pull/800 and https://github.com/dimagi/formplayer/pull/801

**Comment by orangejenny** [comment]
> If it's easier we can decide to only do this on changes with dropdown fields assuming that we only support inter-depency of fields for the dropdowns and not for free texts field.

Yeah, I think the requested changes in HQ behavior make sense, but I'd likely only have HQ send a request on dropdown changes. I'm a little worried about how performance will go with large lookup tables and/or complex default value expressions.

> The reason being a search field can define an itemset instance which is dependent on the output of another search field (location heirarchies for example).

Thinking about this more, I don't totally follow. I can imagine a location or lookup table hierarchy of dropdowns, but how would the user write an xpath expression that references the value of a different search prompt?

**Review by orangejenny — APPROVED** [review]

**Comment by shubham1g5** [comment]
@orangejenny [here](https://www.commcarehq.org/a/shubhamgoyaltest/apps/view/1956b120ec4e173ed6a5a160b76fb648/module/1df6da7bbaa4f26f350d0a12c8266d9e04fc4fb7/#case-detail-screen-config-tab) is an example config where the `district` defines an itemset instance nodeset as `instance('district')/district_list/district[state_id = $state]`

**Comment by orangejenny** [comment]
@shubham1g5 Cool! I didn't realize variables like `$state` would be available.

**Review by ctsims — APPROVED** [review]

File: src/main/java/org/commcare/formplayer/application/MenuController.java [hunk 1]
```diff
@@ -96,7 +96,8 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
                     sessionNavigationBean.getQueryDictionary(),
                     sessionNavigationBean.getOffset(),
                     sessionNavigationBean.getSearchText(),
-                    sessionNavigationBean.getSortIndex()
+                    sessionNavigationBean.getSortIndex(),
+                    false
```
orangejenny, line 100: All three of the `advanceSessionWithSelections` calls need to use `sessionNavigationBean.isDoQuery()`, otherwise the case search request itself will do the query, but the subsequent `get_details` request won't and so you can't select the case and end up [here](https://github.com/dimagi/formplayer/blob/64ec42d0a7540c0c2e6f997a34bc25924a1af6b4/src/main/java/org/commcare/formplayer/application/MenuController.java#L132). (It's crazy that the search request gets sent multiple times, but that's a separate issue.)

For what it's worth: I ran into this because I checked this branch out to see how it would interact with a bug I'm fixing, which is that case search doesn't work properly in workflows where you select a parent case and then a child case. The root problem with that bug is that formplayer only has one `queryDictionary`, so it can't handle multiple searches. My approach is turning `queryDictionary` into a map where keys are based on `selections` (like `2,action 1`) and the values are hashtables like the current `queryDictionary`. It's a bit convoluted. Similarly, I expect `doQuery` may have multiple values: you may want to do the query for parent selection but not for child selection. So I'm expecting to turn `doQuery` into a map, again keyed by `selections`.

orangejenny (changes requested):
Found a minor issue, commented in `MenuController`

src/main/java/org/commcare/formplayer/application/MenuController.java
shubham1g5: Ahh I didn't realize there is another subsequent search call. Thanks for highlighting. 

> case search doesn't work properly in workflows where you select a parent case and then a child case.

Ahh interesting. Your solution sounds right to me. I would be happy to make those changes if needed.

orangejenny: Thank you, I think I'm almost there, should have a PR up shortly 🙂

orangejenny:
@shubham1g5 I'm working on the HQ side of this and ran into an issue where when executing a search, a null pointer exception would pop up [here](https://github.com/dimagi/commcare-core/blob/03839cf2da87f4d9f86a64e3bc969c65e0a44c73/src/cli/java/org/commcare/util/screen/QueryScreen.java#L138) because `choices` hadn't been initialized yet. Added e5e65b5b0708a6061b13d270ec2ad3f3ed8ff7f8 to handle this.

shubham1g5:
Thanks @orangejenny . Added some tests to compliment the PR and [fixed another issue](https://github.com/dimagi/formplayer/pull/799/commits/8f750df48faac00c7d51fa82430cc4830197fd53) which was causing the query prompts values to be reinitialized to default/empty values after we have set them from `querydictionary` ( was happening due to `getNextMenu` call and in turn the `QueryScreen.init()` call again after we set the answers [in this block](https://github.com/dimagi/formplayer/blob/caseClaim/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L227))

orangejenny (approved):

orangejenny:
Web apps UI support added in https://github.com/dimagi/commcare-hq/pull/28955

PR #802, opened 2020-12-29
Test harnessing for breaking schema changes

ctsims:
New tests which can be run against an in-place snapshot of what the current implicit database schemas are, using the same code paths as the production code.

Adds a gradle target `CreateSnapshotDbs` which will replace and update the snapshotted databases for when the models are changed. That target processes the `sandbox_reference` resources as if they were requested by a real user, and snapshots the resulting sqlite dbs. 

This process should give us the ability to catch unintended changes to db schemas.


src/test/java/org/commcare/formplayer/tests/SnapshotTests.java
```diff
@@ -0,0 +1,110 @@
...
+        enumerate("User Fixtures",sandbox.getUserFixtureStorage());
+    }
+
+    @Test
+    public void testAppSandbox() throws Exception {
+        try {
+            //Note - Currently this is actually expected to blow away the DB a surprisingly large
+            //percentage of the time, but should be expanded in the future to be more robust
+            //when app installs are less fragile
+            doInstall("sandbox_reference/snapshot.json");
```
shubham1g5, line 78: Since this is not relevant today as we wipe out the App DB on any errors, checking if this test doesn't serve a purpose today and is only here to better support any changes in future with how we handle App DB.

shubham1g5 (approved):
love the approach here! Some optional nits, but looks good to go as it is.

src/test/java/org/commcare/formplayer/tests/SnapshotTests.java
```diff
@@ -0,0 +1,110 @@
...
+        UserDB database = getUserDbConnector("snapshot", "snapshot_test", null);
+        if (!database.databaseFileExists()) {
+            Assert.fail("Snapshot UserDB Missing for tests you may need to rebuild the snapshot " +
+                    "with the CreateSnapshotDbs Gradle Task ");
+        }
+        //Try to enumerate records of each type
+        UserSqlSandbox sandbox = new UserSqlSandbox(database);
+        enumerate("Case", sandbox.getCaseStorage());
+        enumerate("Ledger", sandbox.getLedgerStorage());
+        enumerate("User",sandbox.getUserStorage());
```
knguyen142, line 67: Very small nit, but spacing in this block is inconsistent.

knguyen142 (commented):
I've never seen this snapshot approach for models before. Is that because we don't intend to make changes to the schema because they can break our current DBs because lack of migrations/versioning?

src/test/java/org/commcare/formplayer/tests/SnapshotTests.java
ctsims: thanks! Fixed.

src/test/java/org/commcare/formplayer/tests/SnapshotTests.java
```diff
@@ -0,0 +1,110 @@
...
+
+            FileUtils.copyDirectory(new File(GenerateSnapshotDatabases.snapshotDbDirectory)
+                    , destDirectory);
+        } catch (IOException exception) {
+            exception.printStackTrace();
+            Assert.fail("Error moving snapshot data to test directory");
+        }
+    }
+
+    protected boolean remoteDatabaseFoldersAfterTests() {
```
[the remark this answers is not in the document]
ctsims: whoops, thanks!

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -175,11 +175,19 @@ public void setUp() throws Exception {
...
+        new SQLiteProperties().setDataDir(getDatabaseFolderRoot());
         MockTimezoneProvider tzProvider = new MockTimezoneProvider();
         DateUtils.setTimezoneProvider(tzProvider);
     }
 
+    protected String getDatabaseFolderRoot() {
+        return "testdbs/";
+    }
+
+    protected boolean remoteDatabaseFoldersAfterTests() {
```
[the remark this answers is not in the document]
ctsims: Indeed, good catch. Thanks!

src/test/java/org/commcare/formplayer/tests/SnapshotTests.java
ctsims: Hey @shubham1g5 ,

Confirming that right now that this wouldn't really catch "Breaking changes" in the appdb schema, the big thing it would catch today is a change to the App DB wiping process you are describing that results in a broken state, which I think is a realistic regression. The more I dig into the app db install process the more fragile it looks, so I wanted to make sure that starting with an old DB in place works with each release.

ctsims:
> I've never seen this snapshot approach for models before. Is that because we don't intend to make changes to the schema because they can break our current DBs because lack of migrations/versioning?

Hey @knguyen142 . Yeah, these tests just make sure that when we merge breaking db model changes that we make the relevant change to the database schema numbers, since we don't have an effective migration process for the sqlite dbs other than "re-request from the source of truth and rebuild". That rebuild could be automated, in theory, but that would mean that we could end up missing real bugs that are breaking the db, so right now we require someone to manually trigger the schema version jump.

PR #803, opened 2021-01-04
Parent/child support for case claim

orangejenny:
https://dimagi-dev.atlassian.net/browse/QA-2123

This branches off of https://github.com/dimagi/formplayer/pull/799, first new commit is [Added QueryDictionary](https://github.com/dimagi/formplayer/commit/06d2f03c2f4ecd9852696be7bd2d4f5d3364cf59). This also means it depends on https://github.com/dimagi/commcare-core/pull/966

The issue is that formplayer only had one `queryDictionary`, so it couldn't handle multiple searches in the same session. This PR adds a `QueryData` object that allows for multiple `queryDictionary`/`doQuery` values, one for each search command.


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -214,12 +216,17 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
...
 
-            if (nextScreen instanceof FormplayerQueryScreen && queryDictionary != null) {
-                notificationMessage = doQuery(
-                        (FormplayerQueryScreen)nextScreen,
-                        menuSession,
-                        queryDictionary
-                );
+            String queryKey = menuSession.getSessionWrapper().getCommand();
+            if (nextScreen instanceof FormplayerQueryScreen && queryData != null) {
+                if (queryData.getExecute().get(queryKey) != null && queryData.getInputs().get(queryKey) != null) {
```
shubham1g5, line 221: Should we check for `queryData.getExecute().get(queryKey)` being `true` ?

shubham1g5 (commented):
Looks good overall, left one query though separately wondering if we can combine both the `inputs` and `execute` inside a single `query_key` like - 
````
{
	"query_data":{
		"query_key":{
			"inputs": [
				"name": "burt"
			],
			"execute" : true
		}
	}
}
````

It just seems little more cleaner to me but is totally optional for this PR.

orangejenny:
@shubham1g5 I added a couple of commits to update the `QueryData` structure as you described - I think it's much cleaner now, thank you for the suggestion. I also rebased to integrate updates to the case claim branch. It might be easier to review looking at the [diff with caseClaim](https://github.com/dimagi/formplayer/compare/caseClaim...caseClaim+jls/parent-child-case-search).

shubham1g5 (approved):
Thanks, Looks great to me. You might wanna point the core to head of your [master-merge branch](https://github.com/dimagi/commcare-core/tree/jls/formplayer-merge-master) so that  we can get a pass on the travis build.

orangejenny:
Rebased to pick up recent changes to https://github.com/dimagi/formplayer/pull/799/, no need to re-review.

orangejenny:
@shubham1g5 thank you for the work on tests

PR #804, opened 2021-01-08
add flag for skipping full validation

calellowitz:
This allows forms to skip some of the validation that occurs on submit. If the answer in the submission matches the answer formplayer's model, it will no revalidate. This will still catch required questions and changes since the last validation, but will no longer catch the case where a later response invalidates an earlier one. As such, it should be used with caution, but will provide meaningful speed ups when used in that way.

### DIFF
--- a/src/main/java/org/commcare/formplayer/api/json/JsonActionUtils.java
+++ b/src/main/java/org/commcare/formplayer/api/json/JsonActionUtils.java
@@ -97,9 +97,11 @@ public class JsonActionUtils {
                                                   FormEntryModel model, String answer,
                                                   FormEntryPrompt prompt,
                                                   boolean oneQuestionPerScreen,
-                                                  FormIndex navIndex) {
+						  FormIndex navIndex,
+						  boolean skipValidation) {
         JSONObject ret = new JSONObject();
         IAnswerData answerData;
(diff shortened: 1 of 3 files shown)


orangejenny (approved):
The overall

src/main/java/org/commcare/formplayer/api/json/JsonActionUtils.java
```diff
@@ -113,7 +115,12 @@ private static JSONObject questionAnswerToJson(FormEntryController controller,
                 return ret;
             }
         }
-        int result = controller.answerQuestion(prompt.getIndex(), answerData);
+	IAnswerData currentAnswerData = prompt.getAnswerValue();
```
ctsims, line 118: One quick note: I think this might compute more information than intended, and that it might be equally valid to retrieve the value directly from the model (IE: Model->getform->getelement->getvalue with the index)

ctsims (approved):
This looks good to me. 

If there is a way to get a test written around the validation process, it would make further extensions of improving the efficiency of this process much faster, since we'll want to make sure that we have positive examples of success. We could potentially be less conservative with null value answers as well if we can check for the required conditions in the prompt.

PR #805, opened 2021-01-11
url encode loginas

knguyen142:
## Summary

https://dimagi-dev.atlassian.net/browse/SAAS-11579

Login_as users with characters like `+` and `-` cause an error because the django app converts the `+` query param to a ` `.
Formplayer needs to URL encode the `as` query param. Mostly followed https://stackoverflow.com/a/51678215 to use a `UriComponentsBuilder` instead of a normal url string builder, and manually url encoded the `as` param (and only this one for now, since I don't want to cause any potential issues with the other params). Happy to url encode more if we're confident it won't break.

### Automated test coverage

Unfortunately I looked at [RestoreFactoryTest](https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/RestoreFactoryTest.java) and it didn't have the skeleton to easily test this. I can devote time to investigating whether it's worth it to set it up, but I imagine restoring is more easily done in integration tests, unless we mock a lot of things that I'm not too familiar with.

### QA Plan

This was tested locally and works (can successfully login as a mobile user with a `+` when it failed with error message before). Also logs the correct encoded url.

Screen Shot 2021-01-11 at 1 51 56 PM

Observe that I did not url encode other params like `device_id` because I'm not sure the ramifications (e.g. we key on that and a change would break code). Let me know if I should.

TODO: Will deploy to staging and open a ticket.

### Safety story

I had to refactor some code to convert the string builder to a uricomponentsbuilder and if this causes some regressions, the changes are fundamental enough that it should be caught on staging. The actual url encoding of the `as` param should not cause any regressions as far as I'm concerned because I don't think django assumes that or relies on the param being unencoded. I can't think of a case where a user name currently works and where encoding it would break functionality?

### Rollback instructions

- [x ] This PR can be reverted after deploy with no further considerations

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -33,6 +33,7 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Component;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponentsBuilder;
 import org.w3c.dom.Document;
 import org.w3c.dom.NamedNodeMap;
 import org.w3c.dom.Node;
@@ -59,9 +60,13 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import java.net.URLEncoder;
 import java.sql.SQLException;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Factory that determines the correct URL endpoint based on domain, host, and username/asUsername,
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -611,37 +616,45 @@ private HttpHeaders getHmacHeaders(String requestPath) {
...
-
+        builder.queryParam("device_id", getSyncDeviceId());
         if (useLiveQuery) {
-            builder.append("&case_sync=livequery");
+            builder.queryParam("case_sync", "livequery");
         }
-
         if( asUsername != null) {
-            builder.append("&as=").append(asUsername);
+            String unEncodedAsUsername = asUsername;
```
knguyen142, line 639: @snopoke I see you made some changes to this logic just now, can you check my modified changes.

snopoke: LGTM

snopoke (approved):

knguyen142:
Hey @snopoke, this is still going through QA. My testing shows that its good, but we haven't thoroughly tested. Did this need to go out soon?

snopoke:
@knguyen142 oh, sorry, I didn't see any QA label on it. I merged it since I'm making another change to this piece of code. Do you want to revert?

knguyen142:
Reverted and new PR https://github.com/dimagi/formplayer/pull/811

PR #808, opened 2021-01-15
pass requesting user from formplayer back to HQ

snopoke:
This is necessary to allow CommCare to know exactly which user the restore is being done for (and as). It also means that CommCare doesn't need to fake the `restoreAs` param when making the request to Formplayer which will actually make Formplayer create the UserDB in the incorrect location.

Corresponding HQ change: https://github.com/dimagi/commcare-hq/pull/28989/commits/dfba5ea6dcfbb88480b4d2166d4e03ac5a9d476e and https://github.com/dimagi/commcare-hq/pull/28989/commits/158c2f7ae020a3a5ea11b0a145d09fa318cd70c3


snopoke:
Depends on https://github.com/dimagi/formplayer/pull/811

snopoke:
@calellowitz I've tested this locally and on staging, I think it's GTG.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -148,7 +149,7 @@ public void configure(String domain, String caseId, HqAuth auth) {
         this.setHqAuth(auth);
         this.hasRestored = false;
         this.configured = true;
-        sqLiteDB = new UserDB(domain, username, null);
+        sqLiteDB = new UserDB(domain, scrubbed_username, null);
```
snopoke, line 152: note: all other uses of `username` fall into the following buckets:

* logging
* sentry data
* as part of the device ID

calellowitz: Is this going to break everyone's existing user dbs?

calellowitz (commented):
One question, rest looks fine.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
snopoke: no, `scrubbed_username` is the same value as `username` was prior to this change (https://github.com/dimagi/formplayer/pull/808/files#diff-7dff02876ec9dea6b13f04025dfe2b6231044a78026eb4c86ecd4553d398814fL715-R721)

PR #812, opened 2021-01-18
skip json creation on validation

calellowitz:
We were generating the full form json (`tree` attribute that gets passed to the front end) after every answer validation, when we don't need it at all during the validation stage. We never pass that back to the front end regardless of whether or not validation is successful. We do still need that when answering individual questions, which is why this is a flag for now, although I would like to refactor some of this code so that each part can call what they need, rather than having one method with a bunch of flags.

### DIFF
--- a/src/main/java/org/commcare/formplayer/api/json/JsonActionUtils.java
+++ b/src/main/java/org/commcare/formplayer/api/json/JsonActionUtils.java
@@ -83,7 +83,7 @@ public class JsonActionUtils {
         return new JSONObject().put(ApiConstants.QUESTION_EVENT_KEY,
                 PromptToJson.parseQuestionType(model, new JSONObject()));
     }
-
+             
     /**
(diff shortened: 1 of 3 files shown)


ctsims (approved):
Good catch. Do you have a concrete idea of the relative performance improvement from this change? Seems like it might be very significant (maybe embarrassingly so...)

snopoke (approved):
I also find it a bit odd that we return a JSONObject instead of just using the `FormEntryResponseBean` directly.

calellowitz:
> Good catch. Do you have a concrete idea of the relative performance improvement from this change? Seems like it might be very significant (maybe embarrassingly so...)

The team has been conducting a bunch of before and after timings on various forms in the app known to previously have slow validation. I think they mostly saw about 50% improvement, but I suspect it will be much higher in forms with large lookup tables which is what this originally targeted.

PR #814, opened 2021-01-20
Case Claim: Skip to Default Search and Search Again Workflows

shubham1g5:
Spec: https://docs.google.com/document/d/1Xni7QL2ZdFy4gROUBKR1nLK0m9unR3bSJFOqS0enLEI/edit?ts=6001afe1

#### Summary of Changes

* Adds a `default_search` attribute to `query` node. When `true`, Formplayer auto-initiates the case claim search with default values and replies with the Search Response to WebApps
* Allow actions to be part of Search Results Screen [[commit](https://github.com/dimagi/formplayer/commit/ba45861083e010082ee3749522da17ae62633c03)]
* Adds a `redo_last` flag to `action` node. When set, this should make same request as the last action performed in this case the query search. WebApps should check for this flag and retrigger the search request with a flag `force_manual_action` set. This allows Formplayer to force a manual search even when `default_search` is on. 

Will be following up with some tests for the PR once we are confident that this approach works correctly. 

cross-request: https://github.com/dimagi/commcare-core/pull/971


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -330,11 +343,16 @@ private NotificationMessage doSync(FormplayerSyncScreen screen) throws Exception
...
                                         MenuSession menuSession,
-                                        Hashtable<String, String> queryDictionary) throws CommCareSessionException {
+                                        Hashtable<String, String> queryDictionary,
+                                        boolean autoSearch) throws CommCareSessionException {
         log.info("Formplayer doing query with dictionary " + queryDictionary);
         NotificationMessage notificationMessage = null;
-        screen.answerPrompts(queryDictionary);
-        String responseString = queryRequester.makeQueryRequest(screen.getUriString(), restoreFactory.getUserHeaders());
+
+        if (queryDictionary != null) {
```
orangejenny, line 351: I think the null checking may no longer be necessary now that web apps is fixed to not send a null value. But I suppose it's better defensive coding to keep the checks.

orangejenny (approved):

orangejenny:
@shubham1g5 Thoughts on test failures? A number of them reference an unrecognized `locales` field, which doesn't appear in this PR, but this one does look suspicious:
```
org.commcare.formplayer.tests.CaseClaimTests > testAlreadyOwnCase FAILED
    java.lang.NullPointerException
        at org.commcare.formplayer.tests.CaseClaimTests.testAlreadyOwnCase(CaseClaimTests.java:154)
```

PR #819, opened 2021-01-22
remove swagger UI along with dependent config

snopoke:
I don't think we use / care about having the swagger UI available so removing it.


ctsims (approved):

build.gradle
```diff
@@ -102,6 +100,7 @@ dependencies {
     //Unmanaged Dependencies
     testImplementation group: 'com.jayway.jsonpath', name: 'json-path', version: '2.2.0'
     testImplementation group: 'com.jayway.jsonpath', name: 'json-path-assert', version: '2.2.0'
+    testImplementation group: 'com.google.guava', name: 'guava', version: '30.1-jre'
```
dannyroberts, line 103: Curious what this line is about

snopoke: Previously this was a transitive dependency from the swagger library. Removing swagger also removed this JAR which broke the tests since they make use of it.

PR #831, opened 2021-01-27
Move to refreshing one time after all answers have been registered

shubham1g5:
[Issue](https://github.com/dimagi/commcare-core/pull/972/files#r564570490)

This was a regression introduced by https://github.com/dimagi/commcare-core/pull/972/files, So reverting to original approach here and still reserving the [main fix ](https://github.com/dimagi/commcare-core/pull/972/files#diff-6815b1dc78a9b3d06ce7631378492ba2472c305a7b229dd8a159d201e8fd773dR145-R150) from that PR.

@orangejenny  I think this will still need some Web Apps work, Formplayer now send a null value as as `displays[].value` for the dependent itemset  whose value should be cleared, though doesn't seem like Web Apps is reading that and setting it to the search field again. 

cross-request: https://github.com/dimagi/commcare-core/pull/973

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -279,6 +279,7 @@ public class MenuSessionRunnerService {
         if (queryDictionary != null) {
             screen.answerPrompts(queryDictionary);
         }
+        screen.refreshItemSetChoices();
     }
 
 


orangejenny:
@shubham1g5 I don't think this is quite working. There was an HQ problem where web apps wasn't using the value formplayer was sending back, which is being fixed in https://github.com/dimagi/commcare-hq/pull/29070/ ...but locally, even with that HQ fix and the two `caseSearchItemsetFix` branches checked out, I'm seeing the value get cleared even if it's present in the new set of options: https://recordit.co/V66f4KeQzj

shubham1g5:
@orangejenny Are you sure that the the option `:D` has same `value` for both the ratings. I tried reproing it but didn't have much luck.

orangejenny:
@shubham1g5 Check out [this app](https://staging.commcarehq.org/a/bosco/cloudcare/apps/v2/#%7B%22appId%22%3A%22770b0372456b4faa8074fcebbb8743be%22%2C%22steps%22%3A%5B1%2C%22action%200%22%5D%2C%22page%22%3Anull%2C%22search%22%3Anull%2C%22queryData%22%3A%7B%7D%2C%22forceManualAction%22%3Afalse%7D) - in case that link breaks, it's [this module](https://staging.commcarehq.org/a/bosco/apps/view/199d5aa3beea4c9dbb1ea002c087a302/module/722b1db0f3f41f8db3d0ee776e946789ccfac649/) and in case useful, mobile worker `bic` has data.

orangejenny:
just realized my HQ PR isn't on staging, I'll deploy it now

shubham1g5:
@orangejenny thanks, that was helpful and I can repro the issue now, though unfortunately it seems like there is no way to fix this in the current design without re-introducing the issue this PR tries to solve. Formplayer only knows of the index selected by the user for a select question and not the actual choice's value. Therefore after refreshing the itemsets, Formplayer can only check whether an index can still be a valid index in the new sets of choices and set a value corresponding to that index. This will although cause the issue this PR is trying to solve of dependent itemset changing values. The same issue today exists when Web Apps work on dependent itemsets in XForms as well.  

This PR actually results into ignoring the dependent itemset choice index anytime there is a change in the itemset it depends on. This might still be a better behaviour since we won't change value to something else except blank. 

If we wish to reolve this issue, I think we will need to change the design so that WebApps get access to the choices and instead of working with the index, it works with the actual value for that select question.

orangejenny (approved):

orangejenny:
@shubham1g5 Should this be merged? I agree with this:

> This might still be a better behaviour since we won't change value to something else except blank.

PR #832, opened 2021-01-27
Send form name to datadog for granular timings on submit requests

gherceg:
In addition to sending form names to sentry, we also want to send to datadog to allow for filtering there. Some slight changes were made to the `recordTimingCategory` method to make it easy to add more data in the future. I kept the method overloading that was already in place to not break existing calls.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/CategoryTimingHelper.java
+++ b/src/main/java/org/commcare/formplayer/services/CategoryTimingHelper.java
@@ -1,10 +1,5 @@
 package org.commcare.formplayer.services;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.ArrayList;
-import java.util.Map;
-
 import com.timgroup.statsd.StatsDClient;
 
 import org.apache.commons.logging.Log;
@@ -14,6 +9,11 @@ import org.springframework.context.annotation.Scope;
 import org.springframework.context.annotation.ScopedProxyMode;
 import org.springframework.stereotype.Component;
 import org.commcare.formplayer.util.*;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
 
 import javax.servlet.http.HttpServletRequest;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/formplayer/services/CategoryTimingHelper.java
```diff
@@ -64,33 +73,41 @@ public void recordCategoryTiming(Timing timing, String category) {
         recordCategoryTiming(timing, category, null);
     }
     public void recordCategoryTiming(Timing timing, String category, String sentryMessage) {
-        recordCategoryTiming(timing, category, sentryMessage, null);
+        recordCategoryTiming(timing, category, sentryMessage, Collections.emptyMap());
```
gherceg, line 76: On second thought, I think this field should just be nullable.

src/main/java/org/commcare/formplayer/services/CategoryTimingHelper.java
```diff
@@ -66,31 +75,38 @@ public void recordCategoryTiming(Timing timing, String category) {
...
-                "duration:" + timing.getDurationBucket(),
-                "domain:" + domain
-            );
-        } else {
-            datadogStatsDClient.recordExecutionTime(
+        List<String> datadogArgs = new ArrayList<>();
+        datadogArgs.add("category:" + category);
+        datadogArgs.add("request:" + RequestUtils.getRequestEndpoint(request));
+        datadogArgs.add("duration:" + timing.getDurationBucket());
+        if (extras != null) {
```
calellowitz, line 95: Is it possible to just iterate over the map and add those tags rather than check for known keys? Then future updates won't have to update this method as well and we can add arbitrary tags at any point.

calellowitz (approved):
Not blocking, but one suggestion

src/main/java/org/commcare/formplayer/services/CategoryTimingHelper.java
gherceg: Yeah great point.

gherceg: This will be somewhat moot once the DatadogWrapper is used since we will likely just add the tag prior to this method, but I think this functionality is still useful if wanting to easily add a tag for a specific call.

calellowitz (approved):

PR #833, opened 2021-01-28
Add form_name tag to datadog metrics for submit-all requests

gherceg:
This current implementation will become moot with plans to improve how we interact with the datadog client, but if we want to merge for now this is what it looks like.

### DIFF
--- a/src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
+++ b/src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
@@ -62,7 +62,7 @@ public class MetricsAspect {
             AuthenticatedRequestBean bean = (AuthenticatedRequestBean) args[0];
             domain = bean.getDomain();
             // only tag metrics with form_name if one of these requests
-            if (requestPath == 'submit-all' || requestPath == 'answer') {
+            if (requestPath == "submit-all" || requestPath == "answer") {
                 String sessionId = bean.getSessionId();
                 if (sessionId != null) {
                     try {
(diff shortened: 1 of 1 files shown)


calellowitz (approved):
Nothing blocking, as discussed offline, it would be great if we could refactor this code so we don't need to fetch the form session again (or do the request matching), but instead can add tags wherever we need them like with sentry.

gherceg:
Some more bulk testing is being done on staging now to ensure this does not impact performance _too much_,  but based on what has been observed so far we are only going to tag `submit-all` requests. 

Additionally, I left the test metric in place (and refined it) to enable monitoring how much additional time this adds to requests on production. Like @calellowitz said, the ideal solution is to add tags when most convenient to avoid situations like this. The initial solution for this should be in next week.

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
```diff
@@ -41,46 +48,77 @@
...
-                "restore_blocked_time:" + getRestoreBlockedTimeBucket(),
-                "install_blocked_time:" + getInstallBlockedTimeBucket(),
-                "submit_blocked_time:" + getSubmitBlockedTimeBucket()
+                datadogArgs.toArray(new String[datadogArgs.size()])
         );
+
+        if (fetchTimer != null) {
+            datadogArgs.add("task:fetch_form_session");
+            datadogStatsDClient.recordExecutionTime(
+                    "test_timings",
```
gherceg, line 116: @calellowitz does it seem worth adding this tag as a constant? I figure I'll remove this code in the near future and `test_timings` will either never be used again or prove to be somewhat useful.

calellowitz: Rather than name it `test_timings` and add a tag for `fetch_form_session`, it might make more sense to just name the metric `fetch_form_session`, unless we expect to aggregate this with other timings

gherceg: Updated 👍

PR #835, opened 2021-01-29
Increase default lock pool size from 255 to 65k

ctsims:
The default pool available for sandbox ID Hash -> Lock ID is currently very, very small (255) which on a high-concurrency machine could cause significant problems with unintentional locking collisions.

Long term we should probably explore a better way of measuring / reducing the collisions in general due to the high likelihood of overlap (even 65k may not be that much) and due to how uneven Java String hashes are, but this jump should be more than enough short-term unless the hash overlap is extraordinarily non-uniform.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/FormplayerLockRegistry.java
+++ b/src/main/java/org/commcare/formplayer/services/FormplayerLockRegistry.java
@@ -33,7 +33,7 @@ public class FormplayerLockRegistry implements LockRegistry {
     private final Log log = LogFactory.getLog(FormplayerLockRegistry.class);
 
     public FormplayerLockRegistry() {
-        this(0xFF);
+        this(0xFFFF);
     }
 
     public FormplayerLockRegistry(int mask) {


snopoke (approved):

knguyen142:
Thanks @ctsims. As we discussed, wouldn't it be better to have an option to make it configurable? 

Just jotting down notes here, in regards to handling collisions better, I believe we can use maps with a high initial capacity (limit resizing) instead of tables and the [computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#computeIfAbsent-K-java.util.function.Function-) to default to a lock/object. I believe we can use a concurrent map for the `locktable` and a synchronized map for the `lockTableLocks` for the granularity.  

The redisRegistry for example uses a [concurrentHashmap](https://github.com/spring-projects/spring-integration/blob/master/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java#L93), and I imagine we'd need a synchronized map because we need "transactional" operations on a key-entry pair.

PR #838, opened 2021-02-01
Add slow request thresholds for specific requests

gherceg:
Lowering thresholds for `submit-all`, `answer` and `nagivate_menu` requests. See this [JIRA ticket](https://dimagi-dev.atlassian.net/browse/USH-619?focusedCommentId=116560&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-116560) for more details about # of additional requests that will be sent to sentry due to this change.

### DIFF
--- a/src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
+++ b/src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
@@ -164,7 +164,7 @@ public class MetricsAspect {
     }
 
     private String getSentryMessageForSlowRequest(String category) {
-        Map<String, Long> messages = new HashMap<>();
+        Map<String, String> messages = new HashMap<>();
         messages.put(EXTREMELY_SLOW_REQUEST, "This request took a long time");
         messages.put(SLOW_REQUEST, "This request was slow");
         if (messages.containsKey(category)) {
--- a/src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
+++ b/src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
@@ -16,6 +16,10 @@ import org.commcare.formplayer.util.*;
 
 import javax.servlet.http.HttpServletRequest;
 
+import java.util.Random;
+import java.util.Map;
+import java.util.HashMap;
+
 /**
  * This aspect records various metrics for every request.
  */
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
```diff
@@ -173,12 +191,32 @@ private long getSubmitBlockedTime() {
...
+        Map<String, Long> slowRequestThresholds = new HashMap<>();
+        slowRequestThresholds.put("answer", Long.valueOf(5 * 1000));
+        slowRequestThresholds.put("submit-all", Long.valueOf(20 * 1000));
+        slowRequestThresholds.put("navigate_menu", Long.valueOf(20 * 1000));
+        return slowRequestThresholds;
+    }
+
+    private String getSentryMessageForSlowRequest(String category) {
+        Map<String, String> messages = new HashMap<>();
+        messages.put(EXTREMELY_SLOW_REQUEST, "This request took a long time");
```
calellowitz, line 204: These two messages are basically impossible to tell apart if you dont already know which is which. Is it possible to add more meaningful messages (that indicate which is slower for example)

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
```diff
@@ -173,12 +191,32 @@ private long getSubmitBlockedTime() {
         return submitService.getSubmitTimer().durationInSeconds();
     }
 
-    private void sendTimingWarningToSentry(SimpleTimer timer) {
+    private Map<String, Long> getSlowRequestThresholds() {
```
calellowitz, line 194: It probably doesn't matter since this is only called once, but it might be nice if this map were part of the object rather than calculated on demand

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
```diff
@@ -118,9 +125,20 @@ public Object logRequest(ProceedingJoinPoint joinPoint) throws Throwable {
...
-        if (timer.durationInMs() >= 60 * 1000) {
-            sendTimingWarningToSentry(timer);
+        long extremelySlowRequestThreshold = 60 * 1000;
+        Map<String, Long> slowRequestThresholds = getSlowRequestThresholds();
+
+        if (timer.durationInMs() >= extremelySlowRequestThreshold) {
+            sendTimingWarningToSentry(timer, EXTREMELY_SLOW_REQUEST);
+        } else if (slowRequestThresholds.containsKey(requestPath) && timer.durationInMs() >= slowRequestThresholds.get(requestPath)) {
+            // limit slow requests sent to sentry, send 1 for every 100 requests
+            int chanceOfSending = 100;
```
calellowitz, line 135: It might be nice to start this at 1000. We can always increase if necessary, but it will prevent us from hitting any limits

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
```diff
@@ -30,6 +34,9 @@
 @Order(1)
 public class MetricsAspect {
 
+    private static final String EXTREMELY_SLOW_REQUEST = "long_request"; // artifact of prior naming
```
calellowitz, line 37: These strings (and variable names) are hard to tell apart, it would be nice if the difference was more meaningful.

calellowitz (approved):
Nothing blocking but a few comments

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
gherceg: Changed to 1000.

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
gherceg: Moved to constructor.

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
gherceg: To not disrupt current workflows, the goal is to keep the message "This request took a long time". With that in mind, only the new message can be improved. I updated but am open to suggestions.

calellowitz (approved):

PR #840, opened 2021-02-01
Datadog wrapper

gherceg:
Enables the ability to add tags "as you go" to datadog, rather than having to fetch specific information at the end of the request resulting in slower (time) requests. 

This is inspired by the way Sentry's client works, in that you can add a tag to the current context or scope, so that all requests sent within that context contain this tag. In this case, the `FormplayerDatadog` is scoped for each request, so a fresh object should be used for each request, and any information that we would like to add to datadog can be tagged along the way, and sent at the end of the request.


src/main/java/org/commcare/formplayer/application/WebAppContext.java
```diff
@@ -247,6 +248,13 @@ public FormplayerSentry raven() {
         return raven;
     }
 
+    @Bean(destroyMethod="cleanup")
```
calellowitz, line 251: What does this do?

calellowitz (commented):
One question but looks good

src/main/java/org/commcare/formplayer/application/WebAppContext.java
gherceg: Ah definitely don't need this. Didn't refine my copy/paste for the `FormplayerSentry` object, but looks like it just specifies to call that method (`cleanup`) when destroying the object. The Datadog Client file doesn't have any special cleanup, so I don't think the `FormplayerDatadog` class needs any either.

gherceg:
@calellowitz this is ready to be reviewed. Verified that request metrics are still coming in on staging after applying this change.

calellowitz (approved):

snopoke (commented):
I don't see anywhere in this PR we it's being used in the way you describe in the PR description. Are there more changes coming later?

snopoke:
> I don't see anywhere in this PR we it's being used in the way you describe in the PR description. Are there more changes coming later?

https://github.com/dimagi/formplayer/pull/851

gherceg:
@snopoke updated the description to be more exact.

snopoke:
@gherceg have you looked Spring's built in metrics support? https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-metrics

gherceg:
I haven't looked into built in metrics support. We use this for Spring specific metrics already right? Are you specifically interested in using Spring's [Micrometer](https://micrometer.io/) to decouple us from datadog?

snopoke:
@gherceg I wasn't thinking specifically of decoupling us from DD, just avoiding reinventing the wheel. I haven't looked at Micrometer in depth so I don't know if it solves any of these issues at all or makes them easier.

PR #841, opened 2021-02-02
Sk/cache form session

snopoke:
Form sessions are fetched multiple times per request, sometimes as much as 3 or 4 times. These objects are expensive to fetch since they line in a remote DB and are large.

Some places they are fetched outside of the controllers:

* [LockAspect](https://github.com/dimagi/formplayer/blob/0a6850fdbeb61d8c5dedcda4c4c338e350fbd275/src/main/java/org/commcare/formplayer/aspects/LockAspect.java#L110)
* [UserRestoreAspect](https://github.com/dimagi/formplayer/blob/0a6850fdbeb61d8c5dedcda4c4c338e350fbd275/src/main/java/org/commcare/formplayer/aspects/UserRestoreAspect.java#L78)
* [FormplayerAuthFilter](https://github.com/dimagi/formplayer/blob/4bb52e8791a597ecbc41ed36e66b3739439d845f/src/main/java/org/commcare/formplayer/application/FormplayerAuthFilter.java#L187)
* [FormplayerStorageFactory](https://github.com/dimagi/formplayer/blob/162db2cd7bb0e6f09f9091c5ea03d4b1bee1481b/src/main/java/org/commcare/formplayer/services/FormplayerStorageFactory.java#L55)

This PR adds caching support to the FormSessionRepo that will fetch the session from cache if it is available. This is analogous to how we use quickcache in HQ. By default the cache will use a NOOPCache which does nothing making this safe to deploy as is. The actual cache can be configured by local config.

I want to test this out using an in memory cache [Caffeine](https://github.com/ben-manes/caffeine/). I think it is unnecessary to use a cache like redis since there is no need to have the cache be accessible across machines (due to persistent request routing). This also makes the caching quite safe since there will not be race conditions between machines.

In terms of memory usage, the cache is currently configured to hold a max of 500 objects. The DB size of form sessions is 1MB on average so possible memory usage is 0.5GB. The cache is also configured to expire objects 5min after last access.

Note: I'm not sure if there is a better way to accomplish this that doesn't involve caching. I had thought of making a request scoped bean that would hold the session object as described here: http://www.javabyexamples.com/request-scoped-data-with-spring-mvc/. The advantage of the caching approach is that subsequent request still benefit from the cache and don't have to do a DB query to get the session object.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,6 +67,7 @@ dependencies {
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-jdbc'
     implementation group: 'org.springframework.boot', name: 'spring-boot-autoconfigure'
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-data-jpa'
+    implementation group: 'org.springframework.boot', name: 'spring-boot-starter-cache'
     implementation group: 'org.xerial', name: 'sqlite-jdbc'
     implementation group: 'redis.clients', name: 'jedis'
 
@@ -85,6 +86,7 @@ dependencies {
     implementation group: 'net.javacrumbs.shedlock', name: 'shedlock-spring', version: '4.14.0'
     implementation group: 'net.javacrumbs.shedlock', name: 'shedlock-provider-jdbc-template', version: '4.14.0'
     implementation group: 'io.micrometer', name: 'micrometer-registry-statsd', version: '1.3.0'
+    runtimeOnly group: 'com.github.ben-manes.caffeine', name: 'jcache', version: '2.8.8'
 
     implementation fileTree(dir: 'libs', include: '*.jar')
     implementation project(path: ':commcare', configuration: 'api')
--- a/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
+++ b/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
@@ -45,10 +45,10 @@ public class PostgresFormSessionRepo implements FormSessionRepo {
     @Autowired
     private JdbcTemplate jdbcTemplate;
 
-    @Autowired
+    @Autowired(required = false)
     private StatsDClient datadogStatsDClient;
 
-    @Autowired
+    @Autowired(required = false)
     private FormplayerSentry raven;
 
     @CacheEvict(allEntries = true)
@@ -74,10 +74,14 @@ public class PostgresFormSessionRepo implements FormSessionRepo {
             deletedRows = this.jdbcTemplate.update(deleteQuery);
             long elapsed = System.currentTimeMillis() - start;
             log.info(String.format("Purged %d stale form sessions in %d ms", deletedRows, elapsed));
-            datadogStatsDClient.time("PostgresFormSessionRepo.purge.timeInMillis", elapsed);
+            if (datadogStatsDClient != null) {
+                datadogStatsDClient.time("PostgresFormSessionRepo.purge.timeInMillis", elapsed);
+            }
         } catch (Exception e) {
             log.error("Exception purge form sessions", e);
-            raven.sendRavenException(e, Event.Level.ERROR);
+            if (raven != null) {
+                raven.sendRavenException(e, Event.Level.ERROR);
+            }
         }
         return deletedRows;
     }
--- a/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
+++ b/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
@@ -95,12 +95,10 @@ public class PostgresFormSessionRepo implements FormSessionRepo {
     }
 
     @Override
-    @Lock(LockModeType.OPTIMISTIC)
     public <S extends SerializableFormSession> Iterable<S> saveAll(Iterable<S> entities) {
-        for(SerializableFormSession session: entities){
-            save(session);
-        }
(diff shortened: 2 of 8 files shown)


src/main/resources/application.properties
```diff
@@ -23,3 +23,8 @@ management.metrics.web.server.request.autotime.enabled=false
 management.metrics.enable.logback=false
 management.metrics.enable.system=false
 management.metrics.enable.spring.integration=false
+
+# caching (override type in application.properties to enable)
+spring.cache.type=none
```
snopoke, line 28: disables caching unless overridden in local config

src/test/java/org/commcare/formplayer/tests/FormValidatorTests.java
```diff
@@ -1,77 +1,34 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 @RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest(classes=Application.class)
-@AutoConfigureMockMvc
-@WebAppConfiguration
-public class FormValidatorTests {
+@ContextConfiguration(classes = TestContext.class)
+public class FormValidatorTests extends BaseTestClass {
```
snopoke, line 26: changes here are to avoid conflicts in test auto configuration

src/main/resources/application.properties
```diff
@@ -23,3 +23,8 @@ management.metrics.web.server.request.autotime.enabled=false
...
+spring.cache.cache-names=form_session
+spring.cache.caffeine.spec=maximumSize=500,expireAfterAccess=300s
```
ctsims, line 30: Do we know what to anticipate WRT the memory impact of caching these objects? seems easy for this to take ~1gb of heap or so. That feels very worth the hit to me, but wasn't sure if you'd thought it through in more detail.

ctsims (approved):
Awesome improvement. Based on the current metrics, I imagine this will have a huge positive impact with very low downside.

build.gradle
```diff
@@ -85,6 +86,7 @@ dependencies {
     implementation group: 'net.javacrumbs.shedlock', name: 'shedlock-spring', version: '4.14.0'
     implementation group: 'net.javacrumbs.shedlock', name: 'shedlock-provider-jdbc-template', version: '4.14.0'
     implementation group: 'io.micrometer', name: 'micrometer-registry-statsd', version: '1.3.0'
+    runtimeOnly group: 'com.github.ben-manes.caffeine', name: 'jcache', version: '2.8.8'
```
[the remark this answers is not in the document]
snopoke: That's correct. There are a number of supported engines but this is the most modern of the 'in memory' options. The other option is 'ehcache'.

src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java
```diff
@@ -29,24 +31,27 @@
 import java.util.Map;
 import java.util.Optional;
 
+
 /**
  * Postgres implementation for storing form entry sessions
  * Corresponds to the new_formplayer_session table in the formplayer database
  */
 @Repository
+@CacheConfig(cacheNames = {"form_session"})
```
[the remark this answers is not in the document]
snopoke: That's correct.

src/main/resources/application.properties
snopoke: We could adjust this strategy to be based on the actual memory size but that requires some custom config. I figured based on the size of the objects in the DB that 500 would take up about 1GB of memory (on average). This can also be adjusted via local config if necessary.

src/main/resources/application.properties
snopoke: Yes, long term I think we'd enable it by default but at least for the initial rollout I wanted it disabled. The alternative is that we enable it by default now and then roll out config changes to disable it on prod envs until we're happy with it.

I'm not tied to either approach - this seemed easier initially.

shubham1g5:
@snopoke this looks great. I am wondering what are the pros and cons of using an in-memory cache vs redis ? I am looking to cache the case search results as well ( we currently end up doing case search queries multiple times for a single request) and would love any pointers on what caching mechanism to use.

calellowitz:
> @snopoke this looks great. I am wondering what are the pros and cons of using an in-memory cache vs redis ? I am looking to cache the case search results as well ( we currently end up doing case search queries multiple times for a single request) and would love any pointers on what caching mechanism to use.

Also curious what simon has to say about this, but one advantage of redis for the query results is that redis is shared by all machines, and the users are distributed across machines, so while it was not the original intent, it could provide benefits to all users rather than just local ones. This doesn't have that same constraint because form sessions are only relevant to one user and that user is always on the same machine.

snopoke:
> > @snopoke this looks great. I am wondering what are the pros and cons of using an in-memory cache vs redis ? I am looking to cache the case search results as well ( we currently end up doing case search queries multiple times for a single request) and would love any pointers on what caching mechanism to use.
> 
> Also curious what simon has to say about this, but one advantage of redis for the query results is that redis is shared by all machines, and the users are distributed across machines, so while it was not the original intent, it could provide benefits to all users rather than just local ones. This doesn't have that same constraint because form sessions are only relevant to one user and that user is always on the same machine.

I'd definitely go for redis for anything that is going to be used across multiple machines. Obviously we'd need to make sure the cache keys lend themselves that aren't too specific.

PR #842, opened 2021-02-02
Move adding tag to FormController

gherceg:
While it is ideal to place this in the `FormSession` constructor, I think it adds too much clutter, and obfuscates where this tag is added for future devs. This is also in response to foreseeing that we will add another similar class to `FormplayerSentry` that would also have to be passed into the `FormSession`.


src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -151,9 +154,11 @@ public FormEntryResponseBean answerQuestion(@RequestBody AnswerQuestionRequestBe
     public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitRequestBean,
                                          @CookieValue(name=Constants.POSTGRES_DJANGO_SESSION_ID, required=false) String authToken, HttpServletRequest request) throws Exception {
         SerializableFormSession serializableFormSession = formSessionRepo.findOneWrapped(submitRequestBean.getSessionId());
-        FormSession formEntrySession = new FormSession(serializableFormSession, restoreFactory, formSendCalloutHandler, storageFactory, raven);
+        FormSession formEntrySession = new FormSession(serializableFormSession, restoreFactory, formSendCalloutHandler, storageFactory);
         SubmitResponseBean submitResponseBean;
-
+        if (raven != null) {
```
calellowitz, line 159: It should not be possible for raven to be null, right?

calellowitz (approved):
Nothing blocking, but I think the null checks are unnecessary

src/main/java/org/commcare/formplayer/application/FormController.java
gherceg: Yeah it would signal an issue with Autowired, in which case we would probably want to know about it so makes sense to not have the null check.

calellowitz (approved):

PR #843, opened 2021-02-03
Auto Advance with auto launch in a single request

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-622
https://dimagi-dev.atlassian.net/browse/USH-621

Instead of sending the auto launch actions to Web Apps and making web apps re-issue another request with corresponding selection for the auto launch action, this PR auto-advances Formplayer session on getting a Entity Screen with an action marked as auto launch. 

cross-request: https://github.com/dimagi/commcare-core/pull/975


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -249,6 +258,10 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
                     return syncResponse;
                 }
             }
+
+            if (nextScreen == null) {
+                executeAndRebuildSession(menuSession);
```
shubham1g5, line 263: this is necessary to execute any pending stack ops steps caused by a case claim mark/rewind mechanism. Earlier this step would only happen outside of this `selections` loop, though to support parent-child navigation with case claim, we need to resolve pending stack ops in the current `selections` loop itself. For eg, If we have a parent child navigation setup and user selected a parent and then a child menu, the selections would look something like [parent_menu, parent_case_id, child_menu]. Now while replaying the session with auto launch set on the parent case list, on processing `parent_case_id` the session state would look something like this - 

```

COMMAND_ID m1
MARK case_id
COMMAND_ID search_command.m1
QUERY_REQUEST results : http://localhost:8000/a/test-1/phone/search/)
CASE_ID parent_case_id
```
At this state, needed datum will be null (hence the next screen) until we perform the rewind operation.

ctsims:
@shubham1g5 Just to double check my understanding - If an entity screen has an auto-launch action with it, does the action basically "replace" the state of the entity screen, or can the user still navigate "around" the autolaunch? 

IE: Does going 'back' take you to the entity screen, or would you jump back over it to the previous state?

I'm trying to figure out if there's a clear, definitive session which represents each state the user can be in during navigation.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
orangejenny: I think this is the key I was missing when trying to implement this previously.

orangejenny (commented):
I think this is a significant improvement over the previous code. Do you have a sense of how good test coverage is here? (Speaking as the person who added the original autolaunch logic and did not add test coverage.)

shubham1g5:
> Does going 'back' take you to the entity screen, or would you jump back over it to the previous state?

@ctsims it moves back to the previous state rather than the entity screen. 

> I think this is a significant improvement over the previous code. Do you have a sense of how good test coverage is here? (Speaking as the person who added the original autolaunch logic and did not add test coverage.)

@orangejenny I think the part that lacks test coverage today are parent-child workflows with case search setup for both parent and child. 

Made some more commits to fix some issues with this PR.

orangejenny (approved):

orangejenny:
Force pushed because I did some accidental merging. No real changes.

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -127,55 +128,51 @@ private void initializeBreadcrumbs() {
...
             return false;
         }
         try {
             boolean addBreadcrumb = true;
             if (screen instanceof EntityScreen) {
-                if (input.startsWith("action ") || !confirmed) {
-                    EntityScreen entityScreen = (EntityScreen)screen;
-                    if (input.startsWith("action ") && entityScreen.getAutoLaunchAction() != null) {
-                        sessionWrapper.executeStackOperations(entityScreen.getAutoLaunchAction().getStackOperations(), entityScreen.getEvalContext());
-                        addBreadcrumb = false;
```
orangejenny, line 158: Removing this looks like it reintroduced the bug fixed in https://github.com/dimagi/formplayer/pull/826/, could you check it out?

shubham1g5: Thanks for flagging. [Fix here](https://github.com/dimagi/formplayer/pull/843/commits/5d2b69c4c1dd64f0ecbabe1fa91b751c390572c2)

orangejenny: sweet, thank you

orangejenny (approved):

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -154,6 +154,7 @@ public boolean handleInput(String input, boolean needsDetail, boolean confirmed,
             if (screen instanceof EntityScreen) {
                 EntityScreen entityScreen = (EntityScreen)screen;
                 boolean autoLaunch = entityScreen.getAutoLaunchAction() != null;
+                addBreadcrumb = !autoLaunch;
```
orangejenny, line 157: While debugging something else, I realized breadcrumbs are still off. What's interesting is that when I select a case from [this search-first menu](http://localhost:8000/a/bosco/apps/view/f6d2aa801557aaae7b2c1b180600484a/module/96cb622b8288af9d9fe3e39e9481717dba2d80dd/), the case name *does* get added to the breadcrumb, but it then gets removed later by `executeAndRebuildSession`.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -280,13 +282,13 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
...
-    private NotificationMessage handleQueryScreen(Screen nextScreen, MenuSession menuSession,
-                                                  boolean forceManualAction, QueryData queryData)
+    private NotificationMessage handleQueryScreen(Screen nextScreen, MenuSession menuSession, QueryData queryData,
+                                                  boolean forceSearch, boolean disallowSearch)
             throws CommCareSessionException {
         if (nextScreen instanceof FormplayerQueryScreen) {
             FormplayerQueryScreen formplayerQueryScreen = ((FormplayerQueryScreen)nextScreen);
             formplayerQueryScreen.refreshItemSetChoices();
-            boolean autoSearch = formplayerQueryScreen.doDefaultSearch() && !forceManualAction;
+            boolean autoSearch = forceSearch || (formplayerQueryScreen.doDefaultSearch() && !disallowSearch);
```
shubham1g5, line 291: @orangejenny What does `forceSearch` mean to solve for ? Is there an issue currently without adding this.

src/main/java/org/commcare/formplayer/session/MenuSession.java
shubham1g5: Thanks! It was a [silly mistake ](https://github.com/dimagi/formplayer/pull/843/commits/5cc9af228f0ef284710adcdc2c50f3f0d59ea442)on my side.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
orangejenny: This is an issue with the replay behavior. What happens is you have a parent-child workflow where the parent did an ES search (because it was either search-first or skip to ES results), and then later in the session, when the parent case is being re-selected, formplayer needs to know to run the search again. Otherwise it gets "stuck" on the formplayer query screen.

src/main/java/org/commcare/formplayer/session/MenuSession.java
orangejenny: phew, I was worried that one would be tricky 😅

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5: I see, although these changes doesn't seem harmful I am not so sure if these should be necessary mainly because all the new parent child tests passes without these changes.  For Skip to ES Results since the default search is on `autoSearch` should evaluate to `true` always and for the Search-first, the `execute` param take care of moving forward the search. Is it the case that sometimes `execute` is false for the required search while replaying the session ?

orangejenny: That is a good point about the tests. I didn't get to this today but will take a closer look.

orangejenny: Looked at this more, and I do think this needs to be fixed on the formplayer side. What I'd forgotten in the earlier reply is that this fixes the scenario where you're using "Search Again" for the child menu. I don't think we had any tests where `forceManualAction` is true. Added one in ed782542 that fails before this commit and passes afterward.

This code is basically accounting for the fact that `forceManualSearch` is set on the entire request, it isn't specific to the menu. So when `forceManualSearch` is set because you're searching again on the child menu, that was also being applied to the parent and overriding the parent's "skip to ES results" workflow. I took a look at making `forceManualSearch` a part of `queryData` instead, but had less luck with that.

shubham1g5: Ahh I see, this makes much more sense now. Thanks for taking another paas at it.

PR #845, opened 2021-02-04
reuse 'as' parameter when doing a force sync

snopoke:
Since the introduction of the `for` parameter was problematic due to how SMS forms works (https://github.com/dimagi/commcare-hq/pull/29092) we can just reuse the 'as' field and update HQ to record the auth type.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -680,12 +680,12 @@ public class RestoreFactory {
                     unEncodedAsUsername += "@" + domain + ".commcarehq.org";
                 }
                 builderQueryParamEncoded(builder, "as", unEncodedAsUsername);
+            } else if (getHqAuth() == null && username != null) {
+                // HQ requesting to force a sync for a user
+                builderQueryParamEncoded(builder, "as", username);
             }
             if (skipFixtures) {
                 builderQueryParamEncoded(builder, "skip_fixtures", "true");
-            }
-            if (getHqAuth() == null && username != null) {
-                builderQueryParamEncoded(builder, "for", username);
             }
         } catch (UnsupportedEncodingException e) {
             throw new RuntimeException(String.format("Restore Error: " + e.getMessage()));


src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -680,13 +680,13 @@ private void builderQueryParamEncoded(UriComponentsBuilder builder, String name,
                     unEncodedAsUsername += "@" + domain + ".commcarehq.org";
                 }
                 builderQueryParamEncoded(builder, "as", unEncodedAsUsername);
+            } else if (getHqAuth() == null && username != null) {
+                // HQ requesting to force a sync for a user
+                builderQueryParamEncoded(builder, "as", username);
```
millerdev, line 685: The value of the `for` parameter observed in while investigating the SMS forms issue was not a valid couch username (it did not contain `@{domain}.commcarehq.org`. Should this use logic similar to
https://github.com/dimagi/formplayer/blob/c2f08a60cbbf0ea5e76b37b578ed8077fb8def12/src/main/java/org/commcare/formplayer/services/RestoreFactory.java#L678-L681

millerdev (commented):
Does formplayer have automated tests and could/should this change be tested?

I have not worked on formplayer, so have no context in this code; might want to get a review from someone more familiar.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
snopoke: no since `username` is always the full username, it's only `asUsername` that can be just a mobile workers username without the suffix

snopoke:
To your question of testing - this would require an integration test with HQ to be of any use. It could do with some unit tests that test the url encoding etc but we already know that is working.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
millerdev: > `username` is always the full username

Hmm, that doesn't match up with what I observed when I was troubleshooting the issue yesterday. `for` and `as` were different (see *restore* breadcrumb in [this sentry error](https://sentry.io/organizations/dimagi/issues/2177035085/)). Anonymized: `Restoring from URL https://www.commcarehq.org/a/{domain}/phone/restore/?version=2.0&since=...&device_id=...&as={username}%40{domain}.commcarehq.org&for={username}`

snopoke: Sorry, I should have been more clear. This branch of the if block is only ever triggered by the 'prime formplayer restores'  feature which always sends the full username for the 'username' and the abbreviated username for 'as user'. I don't know why SMS forms use the abbreviated username for `username`.

I guess we could do a check here as well to be more defensive.

PR #846, opened 2021-02-04
form session service

snopoke:
The goal of this PR is:

1. Remove custom form session repository and let Spring and Hibernate (ORM) handle this for us
2. Wrap access to the repo in a service that allows us to do caching and any other logic as well as make it clear what data access is happening.

#### Why?
* Reduce code we have to write and maintain - let Spring do things for us
* In the long run this will make testing easier too
* Make the system easier to reason about by doing things they way the docs suggest we should

Can be reviewed by commit.

#### More notes
I had forgotten that entity managers also have caching built in (https://vladmihalcea.com/jpa-hibernate-first-level-cache/) so the caching at the service level does not save us anything in terms of multiple DB queries for the same object in the same request (since we are using [open-session-in-view](https://www.baeldung.com/spring-open-session-in-view)) but it does still give us the benefit of cross-request caching. 

Note that in future I would like to disable `open-in-view` so that we don't have DB transactions open for the entire request.

I'm going to put this on staging for some testing.


src/main/java/org/commcare/formplayer/Application.java
```diff
@@ -28,8 +28,8 @@
 import java.io.IOException;
 
 @SpringBootApplication
-@EnableJpaRepositories(basePackages = {"repo.*", "objects.*"})
-@EntityScan("objects.*")
+@EnableJpaRepositories(basePackages = {"org.commcare.formplayer.repo"})
+@EntityScan("org.commcare.formplayer.objects")
```
snopoke, line 32: These were just wrong and so weren't picking anything up.

src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
```diff
@@ -11,24 +12,60 @@
 @Entity
 @Table(name = "formplayer_sessions")
 public class SerializableFormSession implements Serializable{
+    @Id @GeneratedValue(generator="system-uuid")
+    @GenericGenerator(name="system-uuid", strategy = "uuid")
     private String id;
+
+    @Column(name="instancexml")
```
snopoke, line 19: normally we wouldn't need to annotate all these fields but since we aren't following the configured naming strategy we have to supply the column names manually. I considered writing a custom naming strategy but we aren't even consistent. Also I think I'd rather just rename the columns in the DB later.

src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
```diff
@@ -11,24 +12,60 @@
 @Entity
 @Table(name = "formplayer_sessions")
 public class SerializableFormSession implements Serializable{
+    @Id @GeneratedValue(generator="system-uuid")
+    @GenericGenerator(name="system-uuid", strategy = "uuid")
```
snopoke, line 16: ID creation is now handled by the DB layer. This has the advantage of the DB layer knowing when an object is new vs exists which means it doesn't have to do a DB query to check (as we used to do here: https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/repo/impl/PostgresFormSessionRepo.java#L128)

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -486,13 +486,12 @@ private NewFormResponse generateFormEntrySession(MenuSession menuSession) throws
         FormSession formEntrySession = menuSession.getFormEntrySession(formSendCalloutHandler, storageFactory);
 
         menuSessionRepo.save(new SerializableMenuSession(menuSession));
-        NewFormResponse response = new NewFormResponse(formEntrySession);
+        SerializableFormSession savedSession = formSessionService.saveSession(formEntrySession.serialize());
+        formEntrySession.setSessionId(savedSession.getId());
```
snopoke, line 490: The ID is now assigned on save so set this back in the bean.

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -178,7 +177,6 @@ public FormSession(UserSqlSandbox sandbox,
         this.domain = domain;
         this.postUrl = postUrl;
         this.locale = locale;
-        this.uuid = UUID.randomUUID().toString();
```
snopoke, line 181: handled by the DB layer

src/main/resources/application.properties
```diff
@@ -6,6 +6,7 @@ spring.datasource.driver-class-name=${datasource.formplayer.driverClassName}
 spring.datasource.url=${datasource.formplayer.url}
 spring.datasource.username=${datasource.formplayer.username}
 spring.datasource.password=${datasource.formplayer.password}
+spring.jpa.hibernate.ddl-auto=validate
```
snopoke, line 9: this validates the the Java Entity objects match with the DB schema

snopoke: (note this won't have any effect currently since we have it set to 'null' in local config: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2#L32)

snopoke:
Smoke testing on staging done.

snopoke:
pinging @dimagi/formplayer FYI

calellowitz:
> benefit of cross-request caching.

Do you have a sense where cross-request caching is used for these form sessions? My intuition is that nearly every request that touches a form session modifies it in some way, and so we wouldn't necessarily gain much from that kind of caching.

Relatedly, do you know how this would affect memory usage on the nodes? Our largest operational issue with formplayer right now seems to be memory pressure in the java heap, and I am a little wary of adding such large objects, especially if they are rarely used. I agree the request level caching seems great, if we are hitting it multiple times per request, but I am not totally sure I understand the costs and benefits of this longer term approach.

snopoke:
> Do you have a sense where cross-request caching is used for these form sessions? My intuition is that nearly every request that touches a form session modifies it in some way, and so we wouldn't necessarily gain much from that kind of caching.

Instead of having to hit the DB on every request to load the session it would already be in memory and we can use it without doing the DB hit. When we save it to the DB we update the cache at the same time.

> Relatedly, do you know how this would affect memory usage on the nodes? Our largest operational issue with formplayer right now seems to be memory pressure in the java heap, and I am a little wary of adding such large objects, especially if they are rarely used. I agree the request level caching seems great, if we are hitting it multiple times per request, but I am not totally sure I understand the costs and benefits of this longer term approach.

There is an adjustable limit to how many objects the cache will keep. We also have [metrics](https://app.datadoghq.com/dashboard/dcs-kte-q8e/formplayer-health?from_ts=1612853934835&fullscreen_section=overview&fullscreen_widget=3136571999473932&live=true&to_ts=1612868334835&tpl_var_environment=staging&fullscreen_start_ts=1612854596842&fullscreen_end_ts=1612868996842&fullscreen_paused=false) that we can use to correlate any bad behaviour.

Currently this is configured to store 500 objects. The average row size in the db is 2.3MB and while I'm not sure how that translates in to Java memory we could be looking at a few GB per process.

calellowitz (commented):
Sorry for the slow review, but this looks great.

ctsims (commented):
This looks good to me, it's a big refactor, thanks, Simon!

My reading of this is that it maintains backwards compatibility with the existing serialized data and doesn't need a flyway migraiton, is that accurate?

snopoke:
> My reading of this is that it maintains backwards compatibility with the existing serialized data and doesn't need a flyway migraiton, is that accurate?

Yes, that's correct.

PR #847, opened 2021-02-04
add spring boot devtools

snopoke:
https://docs.spring.io/spring-boot/docs/current/reference/html/using-spring-boot.html#using-boot-devtools

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,6 +50,13 @@ idea {
     }
 }
 
+configurations {
+    developmentOnly
+    runtimeClasspath {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -50,6 +50,13 @@ idea {
     }
 }
 
+configurations {
+    developmentOnly
+    runtimeClasspath {
+        extendsFrom developmentOnly
```
ShivamPokhriyal, line 56: This is pretty nice.

ShivamPokhriyal (approved):

PR #848, opened 2021-02-05
use junit5

snopoke:
@dimagi/formplayer 
Make tests use Junit5

Junit4 has been removed removed from spring 2.4 (we're on 2.2). We will still have to include the dependency since commcare-core requires Junit4

I'm not sure if we can upgrade commcare-core to Junit5 since it's used by commcare-android which also uses Junit4 and can't be upgraded easily to 5 because of some of the test tools it makes use of.

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/WebAppContext.java
+++ b/src/main/java/org/commcare/formplayer/application/WebAppContext.java
@@ -46,6 +46,7 @@ import org.springframework.data.redis.core.StringRedisTemplate;
 import org.springframework.data.redis.serializer.GenericJackson2JsonRedisSerializer;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.web.context.request.RequestContextListener;
 import org.springframework.web.servlet.ViewResolver;
 import org.springframework.web.servlet.config.annotation.DefaultServletHandlerConfigurer;
 import org.springframework.web.servlet.config.annotation.EnableWebMvc;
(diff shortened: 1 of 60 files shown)


src/main/java/org/commcare/formplayer/application/WebAppContext.java
```diff
@@ -323,4 +324,8 @@ public LockProvider lockProvider(DataSource dataSource) {
                         .build()
         );
     }
+
+    @Bean public RequestContextListener requestContextListener(){
```
snopoke, line 328: this is the recommended way to make the request object available.

The only changes to non-test code are related to this - accessing the request from the `RequestContextHolder`

src/main/java/org/commcare/formplayer/util/RequestUtils.java
```diff
@@ -78,4 +81,12 @@ public static String getHmac(String key, String data) throws Exception {
         sha256_HMAC.init(secret_key);
         return Base64.encodeBase64String(sha256_HMAC.doFinal(data.getBytes("UTF-8")));
     }
+
+    public static FormplayerHttpRequest getCurrentRequest() {
+        RequestAttributes attributes = RequestContextHolder.getRequestAttributes();
+        if (RequestContextHolder.getRequestAttributes() != null) {
```
shubham1g5, line 87: nit: can reuse `attributes`

shubham1g5 (approved):
looks great.

src/main/java/org/commcare/formplayer/util/RequestUtils.java
snopoke:
```suggestion
        if (attributes != null) {
```

PR #849, opened 2021-02-08
misc refactoring

snopoke:
The main goal of this refactoring was to use `SerializableFormSession` and `SerializableMenuSession` inside `FormSession` and `MenuSession` instead of deconstructing and reconstructing them.

In the process I found a few small improvements like avoiding serializing the instance XML multiple times in the same request.

Review by commit.
@dimagi/formplayer FYI


src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -385,17 +374,10 @@ public FormSession getFormEntrySession(FormSendCalloutHandler formSendCalloutHan
         FormDef formDef = engine.loadFormByXmlns(formXmlns);
         HashMap<String, String> sessionData = getSessionData();
         String postUrl = sessionWrapper.getPlatform().getPropertyManager().getSingularProperty("PostURL");
-        return new FormSession(sandbox, formDef, username, domain,
-                sessionData, postUrl, locale, uuid,
-                null, oneQuestionPerScreen,
-                asUser, appId, null, formSendCalloutHandler, storageFactory, false, null);
-    }
-
-    public void reloadSession(FormSession formSession) throws Exception {
```
ctsims, line 394: I presume this method was simply never used? I didn't see any other references in this PR.

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -440,64 +394,27 @@ private void deserializeFormDef(String serializedFormDef) throws IOException, De
...
-        serializableFormSession.setAppId(appId);
-        serializableFormSession.setFunctionContext(functionContext);
-        serializableFormSession.setInPromptMode(inPromptMode);
-        serializableFormSession.setRestoreAsCaseId(restoreAsCaseId);
-        return serializableFormSession;
-    }
-
-    public String getDomain() {
-        return domain;
+        session.incrementSequence();
```
ctsims, line 397: This is the only potentially meaningful semantic change I found in the PR. It looks like this changes the overall approach slightly from relying on the user to increment at the right time (specifically when `updateSession` was called), to presuming that any time the session data is serialized we'll increment the counter so that what is written has a new serial count? That sounds right to me, but wanted to make sure I was seeing it correctly.

ctsims (approved):
This is a really helpful refactor, thanks Simon.

src/main/java/org/commcare/formplayer/session/FormSession.java
snopoke: That's correct, currently the sequence ID is only used at the end of the request to send in the response which is after the session is saved. I agree that if this function were called more than once per request we'd double increment.

I could revert this to setting the value to `0` initially and then only incrementing in `updateSession`.

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -395,6 +395,7 @@ private void deserializeFormDef(String serializedFormDef) throws IOException, De
 
     public SerializableFormSession serialize() throws IOException {
         session.incrementSequence();
+        session.setFormXml(serializeFormDef(formDef));
```
snopoke, line 398: @ctsims Removing this was a breaking change - it seems odd that we have to save the `formDef` every time and not just the `instanceXml`

src/main/java/org/commcare/formplayer/session/FormSession.java
ctsims: Cool, and I don't think it's an issue if the number gets incremented more often than necessary, just making sure I see the path through

src/main/java/org/commcare/formplayer/session/FormSession.java
ctsims: Definitely agree with that, it shouldn't be the case that there's any state in `formDef` which changes in between saves. 

I wonder whether it might be the case that there's some piece of initialization which the system is relying incorrectly on persisting (rather than rebuilding) between requests.

Is there a specific test I could run which demonstrates the issue? I could probably provide some insight on why it's breaking.

snopoke: This was one of the tests that broke: [CaseTests.testCaseCreate](https://github.com/dimagi/formplayer/blob/a1e599955b89bea8a95cf16b98ee35c21a91323c/src/test/java/org/commcare/formplayer/tests/CaseTests.java#L32). Submitting the form failed because the case ID in the form was not set.

snopoke: bb02e1bd

calellowitz (approved):

PR #850, opened 2021-02-09
Cache case search response in redis

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-581

Risk Factor: Redis might start to fill up because of this change in which case it should be ok to just revert this PR and redeploy.


src/main/java/org/commcare/formplayer/services/QueryRequester.java
```diff
@@ -2,38 +2,72 @@
...
-        log.info(String.format("Query request to URL %s successful", uri));
         return responseBody;
     }
+
+    private String loadResponseFromCache(String uri) {
+        return caseSearchCache.get(getCacheKey(uri));
+    }
+
+    private void cacheResponse(String uri, String responseBody) {
+        caseSearchCache.set(getCacheKey(uri), responseBody, 2, TimeUnit.MINUTES);
```
calellowitz, line 67: I might start with 5 minutes just because it probably takes a bit to make it through the menus all the way to the form session

calellowitz (approved):
Not worth blocking on, but I think we might benefit from a slightly longer cache time, just because it might take a few minutes to look over the results and select a case then a form.

src/main/java/org/commcare/formplayer/services/QueryRequester.java
shubham1g5: would it be ok to not show user a new case in search as long as 5 mins ? I am not sure how realtime we generally expect  our systems to perform in this regard.

calellowitz: I would not expect any users to be refreshing he same view looking for new results. I think they generally go through the search, find the case they need, then a number of subsequent requests repeat the search because of formplayer structure. Honestly, I think it would be more confusing for a given user session if the results changed, than if they didn't. And five minutes is short enough that by the time they complete their work (fill out the form), it will have long since expired, so the next query will be fresh.

ctsims:
@shubham1g5 / @calellowitz Design question - is there a reason we wouldn't cache this on the HQ side of the request?

That would have the advantage of potential cross-user authentication since HQ has context on user permissions that formplayer doesn't. The only disadvantage of caching on the HQ side is another hop or two for the data, but I think the overall payload isn't huge, and if it were I'd think it might be a better idea to keep a smaller set of per-user data cached locally in-memory. We couldn't cache as many requests that way, but that probably isn't a big deal since the cache churn on these is already quite high, we'd really just want to be preventing them from being repeated in back-to-back kinds of requests, right?

shubham1g5:
> Design question - is there a reason we wouldn't cache this on the HQ side of the request?

We did discuss this point breifly over [here](https://dimagi-dev.atlassian.net/browse/USH-581?focusedCommentId=116517). Would having the potential ability to validate cached data against user permissions allow us to cache search results by url only without user as part of cache key ? If so, HQ cache does seem better to me. @calellowitz  Thoughts ?

src/main/java/org/commcare/formplayer/services/QueryRequester.java
shubham1g5: makes sense. changed to 5 mins [here](https://github.com/dimagi/formplayer/pull/850/commits/8e7c83991316e61f2e7eb74caf72288bfb0e19a6)

shubham1g5:
@calellowitz @ctsims  This has passed QA now, Though I am not clear if we should hold this PR in lieu of dicussion above or roll this out now and possibly revisit the design later sometime in case of any issues.

ctsims:
Hey all.

This is definitely  bit nuanced. I responded in the Jira ticket for further discussion / steps forward.

shubham1g5:
Closing in favour of https://github.com/dimagi/formplayer/pull/911

PR #851, opened 2021-02-09
Use datadog wrapper to add form_name tag

gherceg:
In addition to adding the call to `FormplayerDatadog` to add the `form_name` tag, I consolidated some string constants. Tested on staging.


snopoke (approved):

gherceg:
Thanks @snopoke. Worth noting that there are still numerous places that need to be refactored to use the `FormplayerDatadog` class instead of the `StatsDClient`, this PR only addresses a newly added use case.

PR #861, opened 2021-02-10
Use resolved install reference

snopoke:
Passing an unresolved reference (`null`) to `installService.configureApplication` causes a NullPointerException.

This was introduced by https://github.com/dimagi/formplayer/pull/849


src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -151,14 +151,14 @@ public NotificationMessage updateApp(String updateMode) {
         }
     }
 
-    private void resolveInstallReference(String installReference, String appId, String host){
+    private static String resolveInstallReference(String installReference, String appId, String host, String domain){
```
snopoke, line 154: changed this and `getReferenceToLatest` to pure, static functions rather than mutating state on the class

snopoke (commented):
TODO: write a test for this

ctsims (approved):

calellowitz (approved):

PR #862, opened 2021-02-11
add column to form session table to replace incorrectly typed columns

snopoke:
* `dateOpened` is a date but stored in the DB as a string with a very unhelpful format "Wed Feb 10 17:44:59 SAST 2021"
  * replace it with an auto-generated timestamp column that stores UTC time
* `sequenceId` is an integer but saved in the DB as a string
  * replace it with an auto-incremented `version` column
  * this has the added benefit of providing [optimistic locking](https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#locking-optimistic)

### DIFF
--- a/src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
+++ b/src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
@@ -1,9 +1,12 @@
 package org.commcare.formplayer.objects;
 
 import org.hibernate.annotations.GenericGenerator;
+import org.springframework.data.annotation.CreatedDate;
+import org.springframework.data.jpa.domain.support.AuditingEntityListener;
 
 import javax.persistence.*;
 import java.io.Serializable;
+import java.time.Instant;
 import java.util.Map;
 
 /**
@@ -11,6 +14,7 @@ import java.util.Map;
  */
 @Entity
 @Table(name = "formplayer_sessions")
+@EntityListeners(AuditingEntityListener.class)
 public class SerializableFormSession implements Serializable{
     @Id
     @GeneratedValue( generator = "uuid" )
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
```diff
@@ -45,9 +52,14 @@
     private String menuSessionId;
     private String title;
 
+    // legacy field to be removed once `dateCreated` is fully populated
```
dannyroberts, line 55: This makes sense, just want to make sure you also consider third party hosters. For us we'd be fine deploying it and waiting 8 days or whatever for all old sessions to get deleted, and then deploy code that uses it, but I think we have to expect third parties to probably deploy it once after it's all over so will need some kind of migration from one column to another probably.

dannyroberts (approved):
Nice cleanup. I guess among other things this will let purge code more straightforward (can get rid of `custom_safe_cast`).

src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
snopoke: Yes, I'll make sure to do that.

snopoke:
> I guess among other things this will let purge code more straightforward (can get rid of `custom_safe_cast`).

Yes that's correct

src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
ctsims: @dannyroberts We delete old form sessions after... 7 days? 14? 

I think it would be a good thing for Solutions and SaaS to get synced up on to have a schedule of how often people are expected to deploy, which would let us set very reasonable fixed dates on this. If we expected teams to deploy formplayer at least once a month we could just set a flag for 45 days after this is merged to be able to remove the field.

PR #864, opened 2021-02-11
Record distribution timings

gherceg:
Datadog's [distribution](https://docs.datadoghq.com/developers/metrics/types/?tab=distribution) metric would allow us to choose between various percentiles in datadog rather than relying on config file changes. This PR adds a distribution type metric for formplayer request timings. For more information on the datadog client, see [here](https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/StatsDClient.java#L508-L524).

Note this also upgrades the [java datadog client](https://github.com/DataDog/java-dogstatsd-client/blob/master/CHANGELOG.md) to 2.11.0.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,7 +79,7 @@ dependencies {
     implementation group: 'redis.clients', name: 'jedis'
 
     //Unmanaged Dependencies
-    implementation group: 'com.datadoghq', name: 'java-dogstatsd-client', version: '2.3'
+    implementation group: 'com.datadoghq', name: 'java-dogstatsd-client', version: '2.11.0'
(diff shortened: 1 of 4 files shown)


calellowitz (approved):
does this prepend `formplayer.metrics` like the other ones do?

gherceg:
> does this prepend `formplayer.metrics` like the other ones do?

Yeah that is [setup on the client](https://github.com/dimagi/formplayer/blob/7e9f766ad10967e3f7d0adfcd73a876679a659ba/src/main/java/org/commcare/formplayer/application/WebAppContext.java#L148).

snopoke:
FYI this requires upgrading the DD agent on the machines as well.

PR #865, opened 2021-02-12
remove the ability to pass install reference into formplayer

snopoke:
As mentioned in https://github.com/dimagi/commcare-hq/pull/29161 being able to pass in an install reference is a security risk (see https://docs.google.com/document/d/13mAWSnp6Y1l-O90mJglxiA7eBUSSMoBt5t3jyViwGfU/edit for justification).

Besides being a security risk this 'feature' is never used outside of tests (as best I can tell - only one route in web apps made use of it and that route has not been used in the past month according to proxy log data).

The test code did use this feature fairly extensively to install local CCZ files. I was able to refactor the code and mock out all the places where that happens.

Note that there is probably more that can be done to make this truly impossible outside of tests like enforcing that all references that reach this point are URLs that point to the configured CommCare service and are not allowed to reference local files:

https://github.com/dimagi/formplayer/blob/0a6850fdbeb61d8c5dedcda4c4c338e350fbd275/src/main/java/org/commcare/formplayer/engine/FormplayerConfigEngine.java#L81


src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -75,9 +77,6 @@
     public MenuSession(SerializableMenuSession session, InstallService installService,
                        RestoreFactory restoreFactory, String host) throws Exception {
         this.session = session;
-        session.setInstallReference(
-                resolveInstallReference(session.getInstallReference(), session.getAppId(), host, session.getDomain())
-        );
```
snopoke, line 80: I removed this since it seems unnecessary to resolve the reference of an existing session. The session is created with an already resolved reference and unless the app ID or domain have changed (which they don't) it will resolve to the same thing anyway.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -652,24 +655,43 @@ EvaluateXPathResponseBean evaluateMenuXPath(String menuSessionId, String xpath)
...
+     * @param requestPath path to JSON file with InstallRequestBean content as well as 'installReference'
+     */
     protected CommandListResponseBean doInstall(String requestPath) throws Exception {
-        InstallRequestBean installRequestBean = mapper.readValue
-                (FileUtils.getFile(this.getClass(), requestPath), InstallRequestBean.class);
-        return generateMockQuery(ControllerType.MENU,
-                RequestType.POST,
-                Constants.URL_INSTALL,
-                installRequestBean,
-                CommandListResponseBean.class);
```
snopoke, line 672: this used to call the `installRequest` endpoint which has been removed.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -690,6 +712,33 @@ CommandListResponseBean doUpdate(String requestPath) throws Exception {
         FORM, MENU, UTIL, DEBUGGER,
     }
 
+    private <T> T mockInstallReference(CheckedSupplier<T> supplier, String installReference) throws Exception {
```
snopoke, line 715: This isn't ideal but a better solution will require making more changes to the production code, particularly the `FormplayerConfigEngine`. I'll make a note to tackle that separately.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -769,4 +818,20 @@ public int getTimezoneOffsetMillis() {
...
+    /**
+     * Utility method to extract the 'installReference' field from test JSON files
+     * and map the remaining JSON fields to the given bean type.
+     *
+     * @param requestPath classpath relative path to JSON file
+     * @param clazz Bean class map JSON to
+     * @return Pair of 'installReference' and deserialized JSON bean
+     * @throws IOException
+     */
+    private <T> Pair<String, T> getInstallReferenceAndBean(String requestPath, Class<T> clazz) throws IOException {
```
snopoke, line 830: The alternative to this is to remove the 'installReference' from the test JSON files and pass it in as a separate parameter. This seemed OK for now but it would probably be better to do that in the long run so that we don't mix request bean data with other data.

src/main/java/org/commcare/formplayer/session/MenuSession.java
ctsims: I think menu session serialization is rare anyway. It's only really used for end of form nav, so it seems right that even if there was an update to the app build ID that it would be wrong to follow it, we wouldn't want to update the app in between finishing a form and navigating to the next session state.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
ctsims: Agreed, and this honestly isn't the hackiest part of tests, really.

ctsims (approved):

dannyroberts (approved):

PR #866, opened 2021-02-12
Refactor formplayer datadog wrapper to support detailed tagging

gherceg:
In an effort to reduce the number of unique tag values, this change gives the formplayer datadog wrapper the ability to only add detailed tags (e.g., `form_name`) if the current domain is eligible.

To make a domain eligible for detailed tagging, the domain can be added to the `detailed_tagging.domains` property in application.properties. 

To add a tag to the detailed tag list, the name can be added to the `detailed_tagging.tag_name` property in application.properties.


gherceg:
fyi @dannyroberts. This is the "domain whitelist" change you mentioned on slack.

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
```diff
@@ -81,6 +81,7 @@ public Object logRequest(ProceedingJoinPoint joinPoint) throws Throwable {
         if (args != null && args.length > 0 && args[0] instanceof AuthenticatedRequestBean) {
             AuthenticatedRequestBean bean = (AuthenticatedRequestBean) args[0];
             domain = bean.getDomain();
+            datadog.setDomain(domain);
```
dannyroberts, line 84: Is the `datadog` instance scoped to a particular request? If it's more like a singleton then this is basically setting a global variable. I haven't traced back how this `datadog` is instantiated; just wanted to make sure you considered that

src/main/java/org/commcare/formplayer/util/FormplayerDatadog.java
```diff
@@ -25,55 +34,116 @@ public String formatted() {
...
         HashSet<String> transientKeys = new HashSet<String>();
         
         for (Tag tag : transientTags) {
-            formattedTags.add(tag.formatted());
-            // keep track of transient tag names
-            transientKeys.add(tag.name);
+            if (shouldAddTag(tag.name)) {
+                formattedTags.add(tag.formatted());
+                // keep track of transient tag names
+                transientKeys.add(tag.name);
```
dannyroberts, line 141: I'm not sure which is better: not setting the tag or setting the tag to a predetermined null value. Just wanted to make sure you considered that, and here you chose that not setting the tag is better. This might show up in datadog as e.g. `form_name:N/A` or something like that, which is fine. Since it's not that there is no form name, you might choose to emit it as "form_name:_other" or something like that. But maybe that doesn't matter and this is simpler

dannyroberts (commented):
Neat change. I left a few comments

src/main/java/org/commcare/formplayer/aspects/MetricsAspect.java
gherceg: My understanding is the instance is scoped at the request level due to [this annotation](https://github.com/dimagi/formplayer/blob/e5dd0ee936c7b621607b6d49accf50a4005bb473/src/main/java/org/commcare/formplayer/application/WebAppContext.java#L252).

src/main/java/org/commcare/formplayer/util/FormplayerDatadog.java
gherceg: I think setting `form_name` to some specified value in this case is wise in order to differentiate intentionally not setting the form name vs unable to set the form name. I will make that change.

gherceg:
@calellowitz I addressed Danny's feedback about sending a specific tag value instead of nothing to differentiate between null tag values and intentionally empty values. Specifically replacing the `shouldAddTag` with `getTagValueToSend`. Curious to hear feedback on that change as it doesn't feel as clear and concise as it could be.

calellowitz (approved):

gherceg:
Tested on staging ✅

PR #867, opened 2021-02-15
only update changing fields when saving form sessions

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-690

* Setting [updatable=false](https://www.objectdb.com/api/java/jpa/Column/updatable) on the fields tells the DB layer that it shouldn't include those fields in update statements.
  * fields that get updated are: `sequenceId`, `version`, `currentIndex`, `instanceXml`, `initLang` (locale) 
* Refactor the model to more accurately reflect it's usage and use [Lombok](https://projectlombok.org/features/GetterSetter) to auto generate the required getters and setters


snopoke:
Accidentally merged. Reopened here: https://github.com/dimagi/formplayer/pull/871

PR #871, opened 2021-02-17
Revert "Revert "only update changing fields when saving form sessions""

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-690

* Setting [updatable=false](https://www.objectdb.com/api/java/jpa/Column/updatable) on the fields tells the DB layer that it shouldn't include those fields in update statements.
  * fields that get updated are: `sequenceId`, `version`, `currentIndex`, `instanceXml`, `initLang` (locale) 
* Refactor the model to more accurately reflect it's usage and use [Lombok](https://projectlombok.org/features/GetterSetter) to auto generate the required getters and setters

### DIFF
--- a/src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
+++ b/src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
@@ -1,5 +1,7 @@
 package org.commcare.formplayer.objects;
 
+import lombok.Getter;
+import lombok.Setter;
 import org.hibernate.annotations.GenericGenerator;
 import org.springframework.data.annotation.CreatedDate;
 import org.springframework.data.jpa.domain.support.AuditingEntityListener;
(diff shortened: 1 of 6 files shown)


ctsims (approved):

dannyroberts (approved):

src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
```diff
@@ -1,248 +1,155 @@
...
-    private String postUrl;
-
-    @Column(name="menu_session_id")
-    private String menuSessionId;
-    private String title;
-
-    // legacy field to be removed once `dateCreated` is fully populated
-    @Column(name="dateopened")
+    /**
+     * Deprecated: to be removed once `dateCreated` is fully populated
```
shubham1g5, line 30: @snopoke Would we be able to deprecate `dateCreated` after a week from when this PR was deployed (Given we only keep incomplete forms for a week)  ? Just wondering if I should add pagination to [deprecated methods here](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/repo/FormSessionRepo.java#L32) as well for Incomplete forms pagination support.

snopoke: I don't think it's worth adding pagination to the deprecated methods. The plan is to remove these within the next week or so.

shubham1g5: Awesome. Thanks for confirming.

PR #872, opened 2021-02-17
check domain access for session auth

snopoke:
Currently there is no domain access check in either the Formplayer auth or the CommCare session details API.

This PR adds a domain check to the session details service.

PR to add a similar check on the CommCare side is on its way.

https://dimagi-dev.atlassian.net/browse/SAAS-11862


snopoke:
Related HQ PR: https://github.com/dimagi/commcare-hq/pull/29184

snopoke:
I since have discovered that there is a domain check here https://github.com/dimagi/formplayer/blob/0a6850fdbeb61d8c5dedcda4c4c338e350fbd275/src/main/java/org/commcare/formplayer/beans/auth/HqUserDetailsBean.java#L59.

@orangejenny do you think the CommCare check is worth keeping?

snopoke:
Decided it's not worth having the HQ check so closing this: https://github.com/dimagi/commcare-hq/pull/29184#issuecomment-780546169

PR #881, opened 2021-02-19
populate version along with sequence ID

snopoke:
Hibernate doesn't support null values for version columns so either
we need to populate it in the DB or lazily as in this commit.

Going for lazy option to avoid having to udpate large amounts of data
on the DB.


ctsims:
@snopoke So the plan is to run this for a while and then transition over when we deprecate the old path?

ctsims (approved):

snopoke:
> @snopoke So the plan is to run this for a while and then transition over when we deprecate the old path?

Yes that's correct. When we do the transition I'll run a query to update any null values.

PR #882, opened 2021-02-22
Take case list page size from Web Apps

shubham1g5:
Allows for case list page size to be set by Web Apps using `cases_per_page` parameter from Web Apps request.


src/test/java/org/commcare/formplayer/tests/CasePaginationTests.java
```diff
@@ -39,11 +39,11 @@ protected String getMockRestoreFileName() {
     public void testSort() throws Exception {
         EntityListResponse entityListResponse =
                 sessionNavigate("requests/navigators/pagination_navigator.json", EntityListResponse.class);
-        assert entityListResponse.getEntities().length == EntityListResponse.CASE_LENGTH_LIMIT;
+        assert entityListResponse.getEntities().length == 20;
         String previousName = "";
         for (EntityBean entity: entityListResponse.getEntities()) {
             String currentName = (String) entity.getData()[0];
-            assert previousName.compareTo(currentName) < 0;
+            assert previousName.toLowerCase().compareTo(currentName.toLowerCase()) <= 0;
```
shubham1g5, line 46: `EntitySorter` converts texts values [to lowercase](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/cases/entity/EntitySorter.java#L90) before comparing, so doing the same here and taking into account for same text values.

orangejenny: This would be a good in-code comment

orangejenny (approved):

orangejenny:
Given that this has test coverage, would you be all right with merging it? Staging is getting pretty unwieldy lately for formplayer.

shubham1g5:
Hey yeah, I would be good with deploying this directly to prod once Adil confirms the changes works for him and we can test end to end workflow locally.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -101,15 +101,16 @@
     private static final Log log = LogFactory.getLog(MenuSessionRunnerService.class);
 
     public BaseResponseBean getNextMenu(MenuSession menuSession) throws Exception {
-        return getNextMenu(menuSession, null, 0, "", 0, null);
+        return getNextMenu(menuSession, null, 0, "", 0, null, 0);
```
ctsims, line 104: Suggestion: There's a lot of data being passed one arg at a time here, it might make sense moving forward to create a SearchContext Object which just wraps all of these parameters. That will let us set all of these defaults and values in one place,  and prevent ongoing creep / expansion of the args list.

ctsims (changes requested):
Only the implicit max is blocking feedback, everything else is suggestion.

src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
```diff
@@ -91,11 +91,16 @@ public EntityListResponse(EntityScreen nextScreen,
         } else {
             Vector<TreeReference> references = nextScreen.getReferences();
             List<EntityBean> entityList = processEntitiesForCaseList(detail, references, ec, searchText, neededDatum, sortIndex, isFuzzySearchEnabled);
-            if (entityList.size() > CASE_LENGTH_LIMIT && !(detail.getNumEntitiesToDisplayPerRow() > 1)) {
+
+            if(casesPerPage == 0){
```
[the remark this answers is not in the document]
shubham1g5: Added [here](https://github.com/dimagi/formplayer/pull/882/commits/7b21358c2f84e9049d8b2781ff20dfea61ccce3c)

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5: Noted. Will keep in mind for any next work around this.

ctsims (approved):

PR #883, opened 2021-02-22
fix repeat bug caused by side effects in `getFullFormJSON`

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-707


snopoke:
This could do with some tests

orangejenny (approved):

snopoke:
@orangejenny I've added a few commits:
1. unit test
2. combine the code so that there's only one code path for creating creating a `NewFormResponse`

orangejenny:
cool, thank you for adding tests

PR #884, opened 2021-02-22
Added URL encoding to case claim URL

orangejenny:
https://dimagi-dev.atlassian.net/browse/QA-2563

Depends on HQ PR https://github.com/dimagi/commcare-hq/pull/29214

### DIFF
--- a/src/main/java/org/commcare/formplayer/screens/FormplayerSyncScreen.java
+++ b/src/main/java/org/commcare/formplayer/screens/FormplayerSyncScreen.java
@@ -8,7 +8,10 @@ import org.commcare.util.screen.CommCareSessionException;
 import org.commcare.util.screen.SyncScreen;
 import org.springframework.web.util.UriComponentsBuilder;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.util.Hashtable;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Screen to make a sync request to HQ after a case claim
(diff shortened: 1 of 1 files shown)


shubham1g5:
@orangejenny the code [here](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/QueryRequester.java#L27) seems to purposefully decode urls. Any idea why would we do that and if this change still takes effect when we are decoding urls later in the code path.

orangejenny:
@shubham1g5 Thanks, I didn't realize that. I think that means the UTF-8 encoding I added isn't doing anything and it's only the URL encoding that's relevant. Removed the URL encoding and retested.

shubham1g5:
@orangejenny  Makes sense. Also wondering if a better change would be to perform URL encoding on the final url [over here](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/QueryRequester.java#L19) rather than on the individual params.

orangejenny:
@shubham1g5 I like the idea of acting on the final url instead, but for claiming it's [this one](https://github.com/dimagi/formplayer/blob/91a1a443e096523d5ec0d7fb6e1a437bb390fe04/src/main/java/org/commcare/formplayer/services/SyncRequester.java#L14) where the params are already passed as a request string, and parsing them out seems messy.

shubham1g5:
Ahh I totally forgot this change is for case claim rather than search. Makes sense to me.

shubham1g5 (approved):

PR #886, opened 2021-02-23
upgrade sentry and refactor sentry usage

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-678

* upgrade to new sentry SDK
* refactor sentry usage to take advantage of new SDK capabilities
  * configure tags / extra in the places where they are set (config files, filters, aspects)
  * use logging to capture certain events
 
Most likely easiest to review per commit.

I have tested this locally: https://sentry.io/organizations/dimagi/issues/?environment=development&project=142105&query=is%3Aunresolved

There will be a commcare-cloud PR coming soon.


src/main/java/org/commcare/formplayer/application/FormplayerAuthFilter.java
```diff
@@ -75,10 +72,10 @@ protected void doFilterInternal(HttpServletRequest req, HttpServletResponse resp
...
                         ace.getLog()), ace);
 
-                sentryReporter.sendEvent(
-                    sentryReporter.getEventForException(ace)
-                            .withMessage(ace.getLog())
-                            .withTag("uri",req.getRequestURI()));
+                Sentry.withScope(scope -> {
+                    scope.setTag("uri",req.getRequestURI());
+                    Sentry.captureException(ace);
+                });
```
snopoke, line 76: sent Sentry is configured by Spring during startup it is available regardless of whether the application is in a request or not.

src/main/java/org/commcare/formplayer/util/FormplayerSentry.java
```diff
@@ -37,12 +32,6 @@
     private final String USER_SYNC_TOKEN = "sync_token";
     private final String USER_SANDBOX_PATH = "sandbox_path";
 
-    private SentryClient sentryClient;
-
-    @Value("${commcarehq.environment}")
-    private String environment;
```
snopoke, line 43: configured via `sentry.environment` config property

src/main/java/org/commcare/formplayer/util/FormplayerSentry.java
```diff
@@ -51,196 +40,128 @@
...
 
         public BreadcrumbRecorder setData(String... dataPairs) {
-            HashMap<String, String> data = new HashMap<>();
             // ignores last element if odd number of elements
             int length = dataPairs.length / 2;
             for (int i = 0; i < length; i++) {
-                data.put(dataPairs[2 * i], dataPairs[2 * i + 1]);
+                String key = dataPairs[2 * i];
+                String value = dataPairs[2 * i + 1];
+                if (key != null && value != null){
```
snopoke, line 24: nulls are not allowed

src/main/java/org/commcare/formplayer/services/FormSessionService.java
```diff
@@ -78,9 +76,6 @@ public int purge() {
             }
         } catch (Exception e) {
             log.error("Exception purge form sessions", e);
-            if (raven != null) {
-                raven.sendRavenException(e, SentryLevel.ERROR);
```
snopoke: this is captured via the log statement above.

src/main/java/org/commcare/formplayer/util/FormplayerSentry.java
ctsims: Now that we have lombok integrated, can we set @nonnull on the param list to fail faster?

src/main/java/org/commcare/formplayer/util/FormplayerSentryUserProvider.java
```diff
@@ -0,0 +1,24 @@
...
+import io.sentry.protocol.User;
+import io.sentry.spring.SentryUserProvider;
+import org.commcare.formplayer.beans.auth.HqUserDetailsBean;
+import org.springframework.stereotype.Component;
+
+@Component
+class FormplayerSentryUserProvider implements SentryUserProvider {
+    public User provideUser() {
+        User user = new User();
+        FormplayerHttpRequest request = RequestUtils.getCurrentRequest();
```
ctsims: This is a really helpful pattern that I don't think I've seen, is this why the component itself isn't Request scoped? 

I don't understand the Spring component selection process super well, but is there an alternative pattern where a request scoped UserProvider takes precedence over a singleton provider?

src/main/java/org/commcare/formplayer/aspects/UserRestoreAspect.java
```diff
@@ -59,12 +61,21 @@ public void configureRestoreFactory(JoinPoint joinPoint) throws Throwable {
...
         configureRestoreFactory(requestBean, auth);
+        configureSentryScope(restoreFactory);
 
         if (requestBean.isMustRestore()) {
             restoreFactory.performTimedSync();
         }
     }
 
+    private void configureSentryScope(RestoreFactory restoreFactory) {
+        Sentry.configureScope(scope -> {
```
ctsims, line 72: I think I'm still a bit confused about this scope object.

Is Sentry managing this state per thread? Per Request?

src/main/java/org/commcare/formplayer/application/AbstractBaseController.java
```diff
@@ -131,7 +122,7 @@ private ExceptionResponseBean getPrettyExceptionResponse(Exception exception, Fo
     public ExceptionResponseBean handleHttpRequestError(FormplayerHttpRequest req, HttpClientErrorException exception) {
         incrementDatadogCounter(Constants.DATADOG_ERRORS_EXTERNAL_REQUEST, req);
         log.error(String.format("Exception %s making external request %s.", exception, req));
-        raven.sendRavenException(exception, Event.Level.INFO);
+        Sentry.captureException(exception);
```
ctsims, line 125: Is there a reason this doesn't go through FormplayerSentry?

ctsims (approved):
These look good to me overall, but as I mentioned in one of the comments, I'm having a bit of a tough time understanding now what the lifecycle / scope model is for Sentry.

Is Sentry 'scope' request bound or is it a perpendicular concept?

For Breadcrumb reporting, is it always just safe to use the static object and presume it will be attached correctly?

Is the move away from using Sentry through dependency injection purely to avoid the incongruence early on in the filter level before request scoping is available? It feels like it might be easier now to accidentally make a mistake in how things are reported since the scoping seems more opaque, but in practice if static reporting will always "just work," I'm happy to be wrong about that.

No real issues found other than my confusion about scoping, and a few places where it wasn't obvious to me when/why the code interchanges `FormplayerSentry` and `Sentry` statics for exception reporting.

src/main/java/org/commcare/formplayer/util/FormplayerSentryUserProvider.java
snopoke: I don't think Sentry would support more than one user provider. This isn't request scoped since you can report errors outside of the request scope.

src/main/java/org/commcare/formplayer/application/AbstractBaseController.java
snopoke: The only use case for `FormplayerSentry.captureException` is to alter the level that it is captured at. The default SDK always captures exceptions at the ERROR level but we have a number of places where we capture them at INFO or WARNING.

Previously we used the `FormplayerSentry` methods to apply all the available scope to the events as well but since that is now handled elsewhere there's no need to use it.

snopoke:
@ctsims thanks for the review. The new SDK is quite different from what we had previously which was basically just the API client. The new SDK does a lot more to make it easy to capture tags and data within the code without having to build your own state management system.

> These look good to me overall, but as I mentioned in one of the comments, I'm having a bit of a tough time understanding now what the lifecycle / scope model is for Sentry.
> 
> Is Sentry 'scope' request bound or is it a perpendicular concept?

There are some docs on how scopes work here but essentially scopes are pushed and popped off a stack, they inherit all the context from their parents and whenever an event is created it will use the top most scope to get the context.

The stack of scopes is stored in a 'hub' which is thread local.

https://docs.sentry.io/platforms/unity/enriching-events/scopes/

You may notice that we are now using the `sentry-spring-boot-starter` dependency and not the base SDK. This dependency allows sentry to autoconfigure during Spring startup and also tie into the DispatchServlet so that it can create request bound scopes.

> For Breadcrumb reporting, is it always just safe to use the static object and presume it will be attached correctly?

Yes, it is safe to do so. Breadcrumbs are attached to current scope

> Is the move away from using Sentry through dependency injection purely to avoid the incongruence early on in the filter level before request scoping is available? It feels like it might be easier now to accidentally make a mistake in how things are reported since the scoping seems more opaque, but in practice if static reporting will always "just work," I'm happy to be wrong about that.

The move away from dependency injection for this was just because the class ended being completely stateless and so there was no need to for it to be managed by Spring.

You are correct that the scoping is a bit more opaque since it is not all configured in one place. I don't think that will lead to any mistakes since the scope is well managed already and new scopes inherit from their parents. I think it makes a lot more sense to add data to the scope in the places where the data becomes available rather than passing it around to a central location.

> No real issues found other than my confusion about scoping, and a few places where it wasn't obvious to me when/why the code interchanges `FormplayerSentry` and `Sentry` statics for exception reporting.

The only use case for `FormplayerSentry.captureException` is to alter the level that it is captured at. The default SDK always captures exceptions at the ERROR level but we have a number of places where we capture them at INFO or WARNING.

Previously we used the FormplayerSentry methods to apply all the available scope to the events as well but since that is now handled elsewhere there's no need to use it unless you want to alter the reporting level.

src/main/java/org/commcare/formplayer/aspects/UserRestoreAspect.java
snopoke: Responded in main comment

src/main/java/org/commcare/formplayer/application/AbstractBaseController.java
ctsims: Got it , thanks.

src/main/java/org/commcare/formplayer/util/FormplayerSentryUserProvider.java
ctsims: Ah, sorry, I think I framed that poorly, I was assuming that the second provider would be the one that was used outside of the request scope.

IE: You have an `@Bean` autowired: If you are in a request context, the component that gets injected is one which presumes there's a request and grabs the user, while if you are out of a request context it provides a bean that doesn't have that info. Basically using the Dependency Injection scheme as a kind of Polymorphism. 

As far as I've read, the expectation is pretty much just that different beans are really only provided in different runtime contexts (IE: "Are you in a test environment or a prod environment"), but I wasn't sure if it was also used for different lifecycle contexts.

ctsims:
@snopoke Great, that all makes sense, and it sounds like this is much more of a change to the underlying library model than a change from our side on how we manage it. Nothing else from me.

src/main/java/org/commcare/formplayer/util/FormplayerSentry.java
snopoke: yea, could do that

src/main/java/org/commcare/formplayer/util/FormplayerSentryUserProvider.java
snopoke: I'm not sure - I haven't dug into that. I think the one disadvantage of doing that is it would create 2 classes instead of 1 for basically doing the same thing. I guess we could put them both in a Kotlin file in which case they source would be together. 

I'll make a note to look into that

src/main/java/org/commcare/formplayer/util/FormplayerSentry.java
snopoke: I'm going to leave this as is since I can't tell if any of the elements will be null at compile time. Most of the data is coming from request beans so we should start with making those APIs clear in terms of what can and can't be null.

PR #891, opened 2021-02-23
add interrupt check at decompress

knguyen142:
https://dimagi-dev.atlassian.net/browse/SAAS-10840

Followed https://github.com/dimagi/commcare-core/pull/893

A few points:
- We also have [InterruptedRuntimeException](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/exceptions/InterruptedRuntimeException.java) which might be better?
- `Thread.interrupted()` clears the interrupted status as opposed to Thread.currentThread().isInterrupted(), I can't think of why clearing it would be a problem for us, but something to think about.

### DIFF
--- a/src/main/java/org/commcare/formplayer/sandbox/ArchivableFile.java
+++ b/src/main/java/org/commcare/formplayer/sandbox/ArchivableFile.java
@@ -10,6 +10,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.util.zip.GZIPInputStream;
 
+import org.javarosa.core.model.condition.RequestAbandonedException;
 import org.commcare.formplayer.exceptions.InterruptedRuntimeException;
 import org.commcare.formplayer.exceptions.SqlArchiveLockException;
 
@@ -71,6 +72,9 @@ public class ArchivableFile extends File {
             int len;
             while ((len = gis.read(buffer)) != -1) {
                 fos.write(buffer, 0, len);
+                if (Thread.interrupted()) {
+                    throw new RequestAbandonedException();
+                }
             }
         }
     }


src/main/java/org/commcare/formplayer/sandbox/ArchivableFile.java
```diff
@@ -71,6 +72,9 @@ private static void decompressGzipFile(File gzipFile, File newFile) throws IOExc
             int len;
             while ((len = gis.read(buffer)) != -1) {
                 fos.write(buffer, 0, len);
+                if (Thread.interrupted()) {
+                    throw new RequestAbandonedException();
```
ctsims, line 76: InterruptedRuntimeException would be a better choice here. This exception specifically is intended to signal cleanup during XPath Query evaluation.,

ctsims (changes requested):

ctsims:
And I don't think it matter whether we use `interrupted` v. `isInterrupted()`. `Thread.interrupted()` is a fine choice if we think the error might be able to be caught and handled cleanly without killing the thread, which is possible in this case at the request level. We should probably be comfortable with the worker thread getting terminated entirely, though, which should be totally fine on Spring Boot. It will churn a worker, but occasionally killing and rebuilding a worker thread isn't a huge deal.

src/main/java/org/commcare/formplayer/sandbox/ArchivableFile.java
knguyen142: I had originally used InterruptedRuntimeException but it requires the InterruptException to [be passed in](https://github.com/dimagi/formplayer/blob/d1e8cba7187ad93248bd4e2440d920b6f0153ffa/src/main/java/org/commcare/formplayer/sandbox/ArchivableFile.java#L111). We can always create a constructor that doesn't require that exception, but I wasn't sure if semantically an InterruptedRuntimeException is tied to an InterruptedException.

Do you also want me to change the 2 SqlStorage references (e.g. https://github.com/dimagi/formplayer/blob/d1e8cba7187ad93248bd4e2440d920b6f0153ffa/src/main/java/org/commcare/formplayer/sandbox/SqlStorage.java#L479) too?

ctsims: Hm, the SqlStorage references throw `RequestAbandonedException` types because they are frequently used in XPath references. Now that we are discussing things, it occurs to me that the `ArchivableFile` errors here are actually part of a `SqlStorage` callstack in the first place, so my initial read was wrong. I now think it makes sense for these to be RequestAbandoned throws. The alternative would probably be throwing an InterruptedException up the stack for the SqlStorage to catch and rethrow, but I don't think that would add much value. 

Thanks for thinking through it.

ctsims (approved):

PR #897, opened 2021-02-24
Adds a wrapper around connection handler for dual writing to sqlite and postgres

ShivamPokhriyal:
https://dimagi-dev.atlassian.net/browse/SAAS-11692

This PR is on top of https://github.com/dimagi/formplayer/pull/813 and is a continuation to the goal of migrating app DB to postgres. 
It'll add a wrapper around SqlStorage so that we can write to both sqlite and postgres and at the same time compare the performance of the two. The wrapper is used for app db only.

### DIFF
--- a/src/main/java/org/commcare/formplayer/postgresutil/PostgresSqlHelper.java
+++ b/src/main/java/org/commcare/formplayer/postgresutil/PostgresSqlHelper.java
@@ -1,6 +1,8 @@
 package org.commcare.formplayer.postgresutil;
 
+import org.apache.commons.lang3.ArrayUtils;
 import org.commcare.formplayer.exceptions.SQLiteRuntimeException;
+import org.commcare.formplayer.services.ConnectionHandler;
 import org.commcare.modern.database.DatabaseHelper;
 import org.commcare.modern.util.Pair;
 import org.javarosa.core.services.storage.Persistable;
@@ -12,8 +14,12 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+
+import static org.commcare.modern.database.TableBuilder.scrubName;
 
 /**
  * @author $|-|!˅@M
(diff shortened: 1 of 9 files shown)


src/main/java/org/commcare/formplayer/postgresutil/PostgresSqlHelper.java
```diff
@@ -50,40 +56,69 @@ public static int setPreparedStatementArgs(PreparedStatement preparedStatement,
...
+    /**
+     * Update entry under id with persistable p
+     *
+     * @param connection  Database Connection
+     * @param tableName   name of table
+     * @param persistable persistable to update with
+     * @param id          sql record to update
+     */
+    public static void updateToTable(Connection connection, String tableName, Persistable persistable, int id, String currentSchema) {
+        upsertToTable(connection, tableName, persistable, id, currentSchema);
```
ShivamPokhriyal, line 119: The write calls to `SqlStorage` updates the [resource id](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/sandbox/SqlStorage.java#L125).
So if we write to `sqlite` first and then write to `postgres`, the [resource id would already be set](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/sandbox/SqlStorage.java#L119) and the `postgres` call to update the resource would fail because the resource was not  present in `postgres`. 

This workaround will use `insert on conflict` idea of `postgres`.

src/main/java/org/commcare/formplayer/sandbox/SqlStorageWrapper.java
```diff
@@ -0,0 +1,325 @@
...
+    private SqlStorage<T> postgresStorage;
+    private DatadogMeterRegistry meterRegistry;
+
+    public SqlStorageWrapper(ConnectionHandler sqliteConnection, ConnectionHandler postgresConnection, Class<T> prototype, String tableName) {
+        sqliteStorage = new SqlStorage<T>(sqliteConnection, prototype, tableName);
+        postgresStorage = new SqlStorage<T>(postgresConnection, prototype, tableName);
+        initMeterRegistry();
+    }
+
+    public void initMeterRegistry() {
```
snopoke, line 39: Did you try autowiring existing MeterRegistry into `FormplayerStorageFactory` and passing that in to the wrapper?

```java
    @Autowired
    private MeterRegistry meterRegistry;

...

    @Override
    public IStorageUtilityIndexed newStorage(String name, Class type) {
        return new SqlStorageWrapper(this.sqLiteDB, this.postgresDB, type, name, meterRegistry);
    }
```

ShivamPokhriyal: I did, but it didn't send anything to datadog and I couldn't find any logs in the console as well.

ShivamPokhriyal: I tried again and found that with autowiring, it was logging the metrics with a prefix of `formplayer.` which I guess is why I missed it in the first place.

src/main/java/org/commcare/formplayer/sandbox/SqlStorageWrapper.java
```diff
@@ -71,12 +47,16 @@ private void compareRunnable(String tag, Runnable sqliteOperation, Runnable post
...
+
+            start = System.currentTimeMillis();
+            postgresOperation.call();
+            long postgresTime = System.currentTimeMillis() - start;
 
-            Timer postgresTimer = meterRegistry.timer("timer.postgres." + tag);
-            postgresTimer.recordCallable(postgresOperation);
-            // TODO compare result with sqlite
+            Timer diffTimer = meterRegistry.timer("storage.timing." + tag);
+            diffTimer.record(postgresTime - sqliteTime, TimeUnit.MILLISECONDS);
```
snopoke, line 59: what made you change the method of capturing the timing?

ShivamPokhriyal: I had a chat with Kevin about it. He mentioned that capturing the difference would be more important so that we can tie each 2 requests.

src/main/java/org/commcare/formplayer/postgresutil/PostgresSqlHelper.java
knguyen142: I see Shivam made changes to this, is this now resolved? Also, if we fail in inserting something to postgres, that means that subsequent reads and potentially writes would fail?

src/main/java/org/commcare/formplayer/sandbox/SqlStorageWrapper.java
```diff
@@ -0,0 +1,305 @@
...
+    private <RESULT> RESULT compareCallable(String tag, Callable<RESULT> sqliteOperation, Callable<RESULT> postgresOperation) {
+        RESULT result = null;
+        try {
+            long start = System.currentTimeMillis();
+            result = sqliteOperation.call();
+            long sqliteTime = System.currentTimeMillis() - start;
+
+            start = System.currentTimeMillis();
+            postgresOperation.call();
+            long postgresTime = System.currentTimeMillis() - start;
```
knguyen142, line 56: Is there anyway to do these 2 asynchronously/simultaneously? 

A potential way we can measure the performance is to see if the overall request is slower. This might be important because even though individual postgres queries might be only subseconds slower than their sqlite counterpart, if we do multiple in the same request it adds up. And in this case, the overall request might be twice as slow if we're doing one request then the other.

knguyen142 (commented):
Overall looks good to me. 

What has the testing been like, locally or on staging? What are the next steps in getting this to staging?

PR #898, opened 2021-02-24
only scan references until found

calellowitz:
This makes two optimizations to help end of form navigation. 

The first is ensuring that all entity screens are initialized without building the reference map (the `false` passed into both `getNextScreen` and `handleInput`). This is the same optimization that normal form nav uses for case list responses. 

The second optimization is for the cases where we really do need the case list references built out, to stop evaluating each item in the case list once we ensure the case is valid (the early exiting for loop). At worst case, this performs the full setup that is currently happening, but in the best case, when cases are early in the list, evaluates fewer references.


src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -60,25 +62,28 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
...
-                        if (entityIds.contains(step.getValue())) {
-                            currentStep = step.getValue();
+                        for (TreeReference ref: entityScreen.getReferences()) {
+                            String id = entityScreen.getReturnValueFromSelection(ref,(EntityDatum) neededDatum, entityScreen.getEvalContext());
+                            if (id.equals(step.getValue())) {
+                                currentStep = step.getValue();
+                                break;
+                            }
                         }
                     }
```
snopoke, line 78: If I'm reading this correctly there should only be one step that matches the `neededDatum` right?
In that case I think this can be simplified (for readability) to:
```java
                Optional<StackFrameStep> matchingStep = steps.stream().
                        filter(step -> step.getId().equals(neededDatum.getDataId()))
                        .findFirst();
                if (matchingStep.isPresent()) {
                    String value = matchingStep.get().getValue();
                    for (TreeReference ref: entityScreen.getReferences()) {
                        String id = entityScreen.getReturnValueFromSelection(ref,(EntityDatum) neededDatum, entityScreen.getEvalContext());
                        if (id.equals(value)) {
                            currentStep = value;
                            break;
                        }
                    }
                }
```

snopoke (commented):
Looks good, one style related commentl.

snopoke (approved):

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
calellowitz: In theory, the app can be set up where two steps match `neededDatum` (`neededDatum.getDataId()` returns `case_id`), in cases where there are multiple case selections (parent/child selection in case lists). There are active QA bugs with that workflow right now (https://dimagi-dev.atlassian.net/browse/QA-2426?focusedCommentId=120648), but it is a possibility.

snopoke: in that case you could do this. It's a bit more complex but it prevents having to iterate over the list of references from the beginning each time and compute the return value. I don't know how much that gains us but if the reference list is long or the cost of computing the return value is not insignificant than it could be a reasonable gain.

```java
                List<String> neededIds = steps.stream().filter(step -> step.getId().equals(neededDatum.getDataId()))
                        .map(step -> step.getValue())
                        .collect(Collectors.toList());
                Iterator<TreeReference> references = entityScreen.getReferences().iterator();
                Set<String> seenIds = new HashSet<String>();
                for (String neededId : neededIds) {
                    if (seenIds.contains(neededId)){
                        currentStep = neededId;
                    } else {
                        while (references.hasNext()) {
                            TreeReference ref = references.next();
                            String refId = entityScreen.getReturnValueFromSelection(ref, (EntityDatum) neededDatum, entityScreen.getEvalContext());
                            seenIds.add(refId);
                            if (refId.equals(neededId)) {
                                currentStep = neededId;
                                break;
                            }
                        }
                    }
                }
```

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -60,25 +62,28 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
...
+                    screen = menuSession.getNextScreen(false);
                     continue;
                 }
                 SessionDatum neededDatum = entityScreen.getSession().getNeededDatum();
-                Set<String> entityIds = entityScreen.getReferenceMap().keySet();
                 for (StackFrameStep step: steps) {
                     if (step.getId().equals(neededDatum.getDataId())) {
-                        if (entityIds.contains(step.getValue())) {
-                            currentStep = step.getValue();
+                        for (TreeReference ref: entityScreen.getReferences()) {
```
ctsims, line 71: If I understand correctly, this is effectively replacing an expensive full evaluation followed by a constant time lookup with a linear lookup that in worst-case takes the same amount of time as the expensive one?

I'm inclined to say it would make sense to add that call directly to the EntityScreen itself (IE: "Confirm ID Exists as valid response" as a function call), since I think it would be valuable for that to be a clear **capability** of the screen rather than a derivative action taken against its components, unless there's a reason not to. That might let the screen itself optimize some aspects, and gets the external callers lets coupled with the internal expectations.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -60,25 +62,28 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
                 EntityScreen entityScreen = (EntityScreen) screen;
                 entityScreen.init(menuSession.getSessionWrapper());
                 if (entityScreen.shouldBeSkipped()) {
-                    screen = menuSession.getNextScreen();
+                    screen = menuSession.getNextScreen(false);
                     continue;
                 }
                 SessionDatum neededDatum = entityScreen.getSession().getNeededDatum();
-                Set<String> entityIds = entityScreen.getReferenceMap().keySet();
```
ctsims, line 67: There is one downside to this change to note, although I don't think it's important enough to change your approach: Calculating the reference map upfront had one positive quality which is that if there were issues with any references on the screen, they would be caught reliably and predictably. With the "minimum effort" approach, we'll non-deterministically hit those problems. 

The only impact I can think of vis-a-vis design here is asking what the inflection point is where this optimization starts to make a difference. IE: I think the current world has positive benefits which are minor (consistency and failing fast), which would be easy to outrule if this optimization is helpful fairly quickly. If it takes really big caselists before this starts to make a big impact, though, we might want to think through how we approach.

ctsims (approved):
Some thoughts on implementation, but the core change seems solid.

Am curious particularly on the performance question.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
calellowitz: Moving it to the EntityScreen makes a lot of sense to me.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
calellowitz: These are fair questions, and I think it depends a lot on usage. The largest gain from this is in cases where end of form nav goes back to the case list itself, because then none of the references are evaluated. It is believable to me that this change is still worth it if we evaluate the full map every time there is a case selection (for validation purposes), but only if there is a selection. That said, I don't know what value we get from erroring when there is a valid selection if other cases in the case list have issues, which would be the main result of that.

As far as timing, I can get some concrete numbers, but it is my understanding that the "first" reference is the longest to calculate by a significant margin, because it evaluates context used by later calculations, and as such the difference between 0 and 1 is much greater than between 1 and 100. I will say that in testing, selecting a case from the first page of the case list caused significant savings, so there are large gains for early cases, but I don't yet know exactly where that inflection point is.

ctsims: Following up on a few of your questions/points:

> The largest gain from this is in cases where end of form nav goes back to the case list itself, because then none of the references are evaluated

This 100% makes sense to me and I think is important to retain regardless of the linear scan. Pushing the check for inclusion in the screen to the screen itself will likely help with that, since it can defer evaluation of ~anything to be deferred to on-demand and cached.

> I don't know what value we get from erroring when there is a valid selection if other cases in the case list have issues, which would be the main result of that.

Currently the erroring wouldn't be related to whether the selection is valid, but rather with whether the malformed entity is before or after the valid selection in an arbitrary ordering.

Since we have an end-user programming system, I think it should be a goal of the platform to both fail-fast and be as consistent as possible in a request. If something fails some % of the time, it's better if that % is 100.

> As far as timing, I can get some concrete numbers, but it is my understanding that the "first" reference is the longest to calculate by a significant margin, because it evaluates context used by later calculations, and as such the difference between 0 and 1 is much greater than between 1 and 100

This should be true in most cases for sure, as the initial reference evaluation unrolling, and the subsequent "bulk load" that fetches and inflates the records are all big bulk operations. I was curious bout the inflection because on some of the bigger county domains we did actually cross the point where it was the subsequent query/transform actions that overtook that runtime, but those were also with multi-column entities rather than just the index value expression, so I wasn't sure what % of the time to expect the linear inflate/transform part of the process to take.

calellowitz:
@ctsims @snopoke followed up with most of the feedback here https://github.com/dimagi/commcare-core/pull/987/files, and changed this code to use the new method in that PR

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
calellowitz: I think I was incorrect here. Whether or not there are multiple steps that match the needed datum doesn't really matter, since what we are checking is if the case id exists, so if there are multiple steps with the same id, the answer would be the same, since the datum is the same. A different datum, such as for parent/child selection, would imply that the second check would be in a different `EntityScreen`, so we would be reevaluating all of this, and would not be able to share results (the while loop wrapping all of this will have advanced).

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -60,25 +62,25 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
...
                 for (StackFrameStep step: steps) {
                     if (step.getId().equals(neededDatum.getDataId())) {
-                        if (entityIds.contains(step.getValue())) {
+                        if (referencesContainStep(step.getValue())) {
```
snopoke, line 71:
```suggestion
                        if (entityScreen.referencesContainStep(step.getValue())) {
```

snopoke (changes requested):

PR #906, opened 2021-02-26
custom JPA repository implementation to avoid 'select on update'

snopoke:
This adds a custom JPA repository implementation that uses ``Session.update`` instead of
``Session.merge`` when saving. This avoids the additional 'select' that
happens when saving detached entities.

With caching enabled Formplayer caches entities between requests to avoid having to re-fetch them
from the DB on successive requests. Since the caching results in detached
entities, the advantage of caching is lost since ``Session.merge`` re-fetch the entity from the DB on update.

It is safe to skip the "select before update" as long as the Formplayer
request routing is consistent. If we remove the consistent routing we would need
to use a global cache (instead of per machine).

Hibernate Docs:
https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/Session.html#merge-java.lang.Object-
https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/Session.html#update-java.lang.Object-

Other resources:
https://www.baeldung.com/hibernate-save-persist-update-merge-saveorupdate
https://thorben-janssen.com/persist-save-merge-saveorupdate-whats-difference-one-use/

### DIFF
--- a/src/main/java/org/commcare/formplayer/Application.java
+++ b/src/main/java/org/commcare/formplayer/Application.java
@@ -3,32 +3,30 @@ package org.commcare.formplayer;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.commcare.formplayer.application.RequestResponseLoggingFilter;
+import org.commcare.formplayer.repo.FormplayerBaseJpaRepoImpl;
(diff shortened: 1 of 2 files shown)


dannyroberts (approved):

dannyroberts:
nice find

PR #908, opened 2021-03-02
add translations attr in form response

knguyen142:
https://dimagi-dev.atlassian.net/browse/SAAS-11642 
Corresonding cchq PR: https://github.com/dimagi/commcare-hq/pull/29279

A client has asked that we support translations for `repeat.dialog.add.new` which correspond to the `Add new repeat` button.

## Design

I believe we handle App UI Translations by having formplayer translate the text and put it in a specific section in the response json. This is a CC UI translations, which we have very little of. Since these are statically defined key names, instead of following the same approach of inserting the translation into a specific portion of the json response (which gets pretty hairy since translations can be nested in there), I created a top level `translations` attribute to the base response class where certain endpoints can append translations that they think would be needed for the webapp and the frontend can access it.

For this translation it is added for from any new form response.

Requests that need it:
Screen Shot 2021-03-02 at 5 01 13 PM

Requests that don't:
Screen Shot 2021-03-02 at 5 01 46 PM

## Testing

Tested locally, will deploy to staging soon. 

Screen Shot 2021-03-02 at 5 41 24 PM

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
@@ -4,6 +4,8 @@ import org.commcare.formplayer.beans.NotificationMessage;
 import org.commcare.modern.session.SessionWrapper;
 import org.commcare.formplayer.util.SessionUtils;
 
+import java.util.HashMap;
+
 /**
  * Base class for responses being sent to the front end. Params are:
  * title - self explanatory
@@ -20,6 +22,7 @@ public class BaseResponseBean extends LocationRelevantResponseBean {
     private String appId;
     private String appVersion;
     private String[] selections;
+    private HashMap<String, String> translations;
 
     public BaseResponseBean() {}
 
@@ -97,5 +100,20 @@ public class BaseResponseBean extends LocationRelevantResponseBean {
 
     public void setSelections(String[] selections) {
         this.selections = selections;
+    }
+
+    public HashMap<String, String> getTranslations() {
+        return translations;
+    }
+
+    public void setTranslations(HashMap<String, String> translations) {
+        this.translations = translations;
+    }
+
+    public void addToTranslation(String key, String value) {
+        if (this.translations == null) {
+            this.translations = new HashMap<String, String>();
+        }
+        this.translations.put(key, value);
     }
 }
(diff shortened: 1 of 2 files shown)


ctsims:
I'd be interested in chatting through approach at a bit of a higher level here, just because things seem a bit hard-coded in this response, and if we are going to need to push through app level translations to more parts of web apps, I think we might need to make some broader considerations, like:

A) What's the boundary for what's translatable? On mobile, for instance, we can't necessarily translate screens before an app is installed. Are there strings on the Web Apps UI which are outside of the App boundary

B) Is it worth making the translation request a downward read into the app database rather than an upward field provided by the app? It would be quite easy to provide sets of strings on demand without needing to cherry-pick which requests provide translation sets.

C) If the set of translations is small enough, maybe we should provide them upfront or upon request in bulk across the whole app (which may be the intent of sending this in the New Form response), and then caching in the frontend layer to prevent the need for further requests.

knguyen142:
Thanks @ctsims. I'll set up a meeting to discuss with you a more comprehensive design, but do you think this one off is good enough to move forward with to unblock the client? 

I think some of the main reservations I had with different designs was that I wasn't familiar with the different translations levels and where they applied. Also, the request flow is a bit complicated because of edge cases, e.g. user can change their language, a new translation can be added, etc. But happy to talk those through.

orangejenny:
> instead of following the same approach of inserting the translation into a specific portion of the json response

@knguyen142 Is the alternative you're thinking of to add the translation as part of the repeat group question? If so, I think having translations at the form level instead of the question level is reasonable.

> A) What's the boundary for what's translatable? On mobile, for instance, we can't necessarily translate screens before an app is installed. Are there strings on the Web Apps UI which are outside of the App boundary

@ctsims USH has had requests to translate the "Applications" header that appears on the home screen of web apps. The buttons for "Incomplete Forms", etc. are on the same screen. For web apps, we could support pre-app translations, although we'd need to create a new domain-level data model and UI for them, and we'd need a way for users to select a language (currently in app preview you can select a "global" language, while in web apps you can only set a language inside a form). 

> C) If the set of translations is small enough, maybe we should provide them upfront or upon request in bulk across the whole app (which may be the intent of sending this in the New Form response), and then caching in the frontend layer to prevent the need for further requests.

I'm ambivalent about this. The set of UI strings in web apps is pretty small: there are ~150 strings translated by HQ, based on the occurrences of `trans`, `blocktrans`, or` gettext` in web apps code. I think it makes the most sense to have the translation as close as possible to its related content...but I can also see caching & simplicity arguments for having one app-level set of UI translations, plus for some of these translations (error messages, at least) it isn't clear which response they would belong to.

knguyen142:
I talked it over with @dannyroberts and it seems like initializing the webapps session with all the translations might not be as complex as I original thought. Still, it needs some designs. How do we feel about pushing this forward with this implementation to respond to the client and then creating a ticket for the design and implementation of a more generic translation mechanism?

orangejenny (approved):

PR #909, opened 2021-03-03
Sk/docker

snopoke:
These changes allow a much simpler setup for running Formplayer in docker. There are two main scenarios in which you can use Formplayer in docker:

### Running HQ and Formplayer in docker
* formplayer communicates with commcare in the docker container:
   * `COMMCARE_HOST=http://web:8000`

### Running HQ outside of docker and Formpalyer inside docker
* formplayer needs to communicate with HQ on the host machine
  * `COMMCARE_HOST: "http://host.docker.internal:8000"`

In both of the above cases:

* commcare is accessed on localhost:8000 which is different from `COMMCARE_HOST` so we need to allow the alternate origin
  * `COMMCARE_ALTERNATE_ORIGINS: "http://localhost:8000"`
* apps built by commcare will have a submission URL of `http://localhost:8000/....` but formplayer can't access that so we need to rewrite it to use `http://web:8000`:
  * `EXTERNAL_REQUEST_MODE: "replace-host"`

Best reviewed by commit:

* multi-stage docker build with jar layers
* properties cleanup
* allow custom CORS origins
* config via env vars
* URL rewriting support for RestTemplate
* minor logging changes


kaapstorm (approved):

Dockerfile
```diff
@@ -1,24 +1,26 @@
...
-# Environment variable WEB_HOST is used for setting where Django is
-# running; either in another container ("web") or on the Docker host
-# machine ("dockerhost").
-#
-# /formplayer/set_webhost adds the IP address of the host machine to
-# /etc/hosts and sets the hostname "webhost" as an alias to HQ.
+FROM adoptopenjdk:8-jre-hotspot as builder
+WORKDIR application
+ARG JAR_FILE=build/libs/*.jar
+COPY ${JAR_FILE} formplayer.jar
```
kaapstorm, line 4: What is the workflow for updating formplayer?

snopoke: I'm still working on that but it's basically this:
```
./gradlew bootJar
docker build -t dimagi/formplayer .
```

kaapstorm: Sweet!

PR #910, opened 2021-03-03
formplayer docker build

snopoke:
Build docker image and push to dockerhub.

Making the gradle build work on Dockerhub is tricky so I went with this approach.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_script:
 script:
   - ./gradlew copyTestResources
   - ./gradlew test
+  - ./gradlew bootJar
 services:
   - postgresql
   - redis-server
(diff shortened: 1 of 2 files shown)


kaapstorm (commented):
I don't know enough to know whether this is a good approach. Can a bad actor use this to push a bad formplayer image to Docker Hub?

Google did help me find
* [bmuschko's Gradle plugin](https://github.com/bmuschko/gradle-docker-plugin)
* [palantir's Gradle plugin](https://github.com/palantir/gradle-docker)

They seem like they're meant to make this easier, but I didn't look very closely.

snopoke:
> I don't know enough to know whether this is a good approach. Can a bad actor use this to push a bad formplayer image to Docker Hub?
> 
> Google did help me find
> 
> * [bmuschko's Gradle plugin](https://github.com/bmuschko/gradle-docker-plugin)
> * [palantir's Gradle plugin](https://github.com/palantir/gradle-docker)
> 
> They seem like they're meant to make this easier, but I didn't look very closely.

Thanks I, did look at a few options but since it's only a few lines of bash script it didn't seem worth it.

kaapstorm (approved):

PR #911, opened 2021-03-04
In Memory Cache For Case Search

shubham1g5:
Alternative implementation for https://github.com/dimagi/formplayer/pull/850

https://dimagi-dev.atlassian.net/browse/USH-581

cross-request: https://github.com/dimagi/commcare-core/pull/979

Both core and this PR should be reverted together in case of rollbacks


src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -0,0 +1,51 @@
...
+public class CaseSearchHelper {
+
+    @Autowired
+    CacheManager cacheManager;
+
+    @Autowired
+    QueryRequester queryRequester;
+
+    @Cacheable(unless="#result == null", key="#username + #uri")
+    public ExternalDataInstance getSearchDataInstance(String username, FormplayerQueryScreen screen, String uri, HttpHeaders headers) {
```
shubham1g5, line 36: Hey @snopoke, I am a bit stuck while implementing caching for Case Search. This method is marked as `@Cacheable` though somehow it's getting executed everytime for the same cache key aka same username and uri passed. Would you be able to take a quick look and see if I am missing something obvious in this implementation.

snopoke: are you testing locally? you need to enable the cache backend: `spring.cache.type=caffeine` (in `application.config`)

snopoke: I think you should also add `case_search` to the list of cache names: https://github.com/dimagi/formplayer/blob/96caf7cb59a09aeb42ba687c36d5ec7364088d11/src/main/resources/application.properties#L33

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -0,0 +1,51 @@
...
+        String responseString = queryRequester.makeQueryRequest(uri, headers);
+        if (responseString != null) {
+            Pair<ExternalDataInstance, String> dataInstanceWithError = screen.processResponse(
+                    new ByteArrayInputStream(responseString.getBytes(StandardCharsets.UTF_8)));
+            return dataInstanceWithError.first;
+        }
+        return null;
+    }
+
+    @CacheEvict(allEntries = true)
```
snopoke, line 46: we should be able to do this with the cache configuration `spring.cache.caffeine.spec=expireAfterWrite=30m`. We could apply that globally to all caches or we could configure the caches separately (https://www.baeldung.com/spring-multiple-cache-managers)

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -0,0 +1,51 @@
...
+    @Autowired
+    QueryRequester queryRequester;
+
+    @Cacheable(unless="#result == null", key="#username + #uri")
+    public ExternalDataInstance getSearchDataInstance(String username, FormplayerQueryScreen screen, String uri, HttpHeaders headers) {
+        String responseString = queryRequester.makeQueryRequest(uri, headers);
+        if (responseString != null) {
+            Pair<ExternalDataInstance, String> dataInstanceWithError = screen.processResponse(
+                    new ByteArrayInputStream(responseString.getBytes(StandardCharsets.UTF_8)));
+            return dataInstanceWithError.first;
```
ctsims, line 41: One thing about the `ExternalDataInstance`: I *don't* think it's safe, unfortunately, for us to reuse the raw java object, since it's not built for concurrency. 

@snopoke Is there a way to manually tell the system HOW to cache the object? It's fairly easy to marshall/unmarshall into a binary format which acts as a deep copy, but it would be nice to only need to do that when the object is retrieved from the cache

ctsims (commented):
A few comments on the in process work.

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
snopoke: I don't know of a way to do this with the annotations, it might be easier in this case to use the raw cache API:

```java
@Autowired
private CacheManager cacheManager;

...

cache = cacheManager.getCache("case_search")
value = cache.get(key, valueLoader)
...
cache.put(key, serializedValue)
```

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
shubham1g5: that was it! I was missing both of these params. Thanks a lot.

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
shubham1g5: @snopoke Lookds like configuring a cache manager in Java code will override the `.properties` config and will enable the cache by default without having the ability to disable the cache using `.properties` file. Therefore I am going to lean towards applying a global config for all caches to `expireAfterWrite` after 5 mins.

snopoke: I think that's fine for the current use cases. The ability to disable it was only intended for initial rollout so it would also be fine to have them permanently configured. Or if we really want we can take an approach like this: https://medium.com/@d.lopez.j/configuring-multiple-ttl-caches-in-spring-boot-dinamically-75f4aa6809f3

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
shubham1g5: @ctsims I am having trouble serializing `ExternalDataInstance` correctly, It seems to skip the [root node](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java#L19) during [serialization](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java#L99) today as root type `AbstractTreeElement` is an interface here. I am not sure what's the best way forward to serialize this. 

Also I did survey ExternalDataInstance and it's containing objects and they all are completely immutable.

ctsims: Ah, I see what's going on. I mistook what the serialization of `ExternalDataInstance` actually accomplished.

It looks like `ExternalDataInstance`'s serialization is used to store **a placeholder** for a data instance that is connected up at runtime and that when we serialize concrete DOM's (like an inline instance in a form) we end up parsing them as [FormInstance](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xform/parse/XFormParser.java#L541) objects.

Options moving forward:

*One:* You could instead switch to serializing the [TreeEleement Root](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/session/RemoteQuerySessionManager.java#L131) itself, since that's how External Data Instance is actually created. `TreeElement`s should be reliably portable. 

*Two:* Given the added complexity of the serialized deep copy, I think it sounds fine to start with the assumption that we'll cache the in-memory version. It does look like the underlying structure is coded with a `TreeElement` DOM which is mutable and has [at least one](https://github.com/dimagi/commcare-core/blob/b47f46d332abd402ba7833dd6617f03306272bf9/src/main/java/org/javarosa/core/model/instance/TreeElement.java#L81) synchronicity concern, but since the DOM is used immutably and doesn't change after init I think it won't be an issue.

build.gradle
```diff
@@ -96,7 +95,8 @@ dependencies {
     implementation group: 'org.simpleframework', name: 'simple-xml', version: '2.7.1'
     implementation group: 'net.javacrumbs.shedlock', name: 'shedlock-spring', version: '4.21.0'
     implementation group: 'net.javacrumbs.shedlock', name: 'shedlock-provider-jdbc-template', version: '4.21.0'
-    runtimeOnly group: 'com.github.ben-manes.caffeine', name: 'jcache', version: '2.8.8'
+    implementation group: 'com.github.ben-manes.caffeine', name: 'jcache', version: '2.8.8'
```
snopoke, line 98: why was it necessary to change this?

src/main/java/org/commcare/formplayer/application/MenuController.java
```diff
@@ -31,6 +32,8 @@
 public class MenuController extends AbstractBaseController {
 
     @Autowired
+    private CaseSearchHelper caseSearchHelper;
```
snopoke, line 35: why not just make the autowired bean in the AbstractBaseController `protected`?

src/main/java/org/commcare/formplayer/application/WebAppContext.java
```diff
@@ -244,6 +245,10 @@ public ArchiveFileRoot formplayerArchiveFileRoot() {
     }
 
     @Bean
+    public CaseSearchHelper caseSearchHelper() {
+        return new CaseSearchHelper();
+    }
```
snopoke, line 250: you could just tag the class with `@Component` and let the auto-discovery pick it up

src/main/java/org/commcare/formplayer/services/QueryRequester.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
+import org.springframework.lang.Nullable;
+import org.springframework.web.client.RestTemplate;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+
+public class QueryRequester {
```
snopoke, line 15: this class can be removed

build.gradle
shubham1g5: reverted this (was experimenting with multiple cache managers before and is not required anymore)

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -0,0 +1,69 @@
...
+    private WebClient webClient;
+
+
+    public ExternalDataInstance getSearchDataInstance(FormplayerQueryScreen screen, URI uri) {
+        Cache cache = cacheManager.getCache("case_search");
+        String cacheKey = getCacheKey(uri);
+        TreeElement cachedRoot = cache.get(cacheKey, TreeElement.class);
+        if (cachedRoot != null) {
+            // Deep copy to avoid concurrency issues
+            TreeElement copyOfRoot = SerializationUtil.deserialize(SerializationUtil.serialize(cachedRoot), TreeElement.class);
```
ctsims, line 40: Nice! Glad to see you found a way through, here.

src/main/java/org/commcare/formplayer/util/SerializationUtil.java
```diff
@@ -23,11 +25,25 @@
...
             throw logAndWrap(e, type, "Totally non-sensical IO Exception");
         } catch (DeserializationException e) {
-            throw logAndWrap(e, type,"CommCare ran into an issue deserializing data");
+            throw logAndWrap(e, type, "CommCare ran into an issue deserializing data");
         }
         return t;
     }
 
+
+    public static byte[] serialize(Externalizable t) {
```
ctsims, line 34: I think the ExtUtil class has an equivalent method?

src/main/resources/application.properties
```diff
@@ -27,8 +27,8 @@ management.metrics.enable.spring.integration=false
 
 # caching (override type in application.properties to enable)
 spring.cache.type=none
-spring.cache.cache-names=form_session
-spring.cache.caffeine.spec=maximumSize=500,expireAfterAccess=300s
+spring.cache.cache-names=form_session,case_search
+spring.cache.caffeine.spec=maximumSize=500,expireAfterAccess=300s,expireAfterWrite=300s
```
ctsims, line 31: @snopoke Is it ok / good practice for us to use the same timings and such for all of these, or should be be trying to tune per-cache?

ctsims (approved):
This seems like a good approach to me, although I'm curious to get @snopoke 's inputs on the caching elements specifically

src/main/resources/application.properties
snopoke: This is fine unless the tuning requirements are vastly different. We did discuss this briefly and decided that apply the same expiration to the session caching wouldn't be an issue. Now that I look at it again I think it would be better to have them separate to avoid expiring the session cache in the middle of a form session. I don't think it's worth blocking the PR. Once this is merged I'll implement this pattern: https://medium.com/@d.lopez.j/configuring-multiple-ttl-caches-in-spring-boot-dinamically-75f4aa6809f3

PR #914, opened 2021-03-05
rest template cleanup

snopoke:
This started out as making sure all usages of `RestTemplate` were using the globally configured beans but I found a lot more to do along the way. The final state is that all simple usage of RestTemplate is handled by a new bean called `WebClient` instead of having multiple 'services' each for performing a single request. The only usages outside `WebClient` are:

* UserDetailsService
* RestoreFactory

More comments inline.

Also note that the impetus for this was that when running in docker we re-write the host in the rest template so anything not using the global rest template can't talk to HQ.


src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -198,27 +196,28 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
...
+                    submitResponseBean.setStatus("error");
+                    NotificationMessage notification = new NotificationMessage(
+                            String.format("Form submission failed with error response: %s, %s, %s",
+                                    e.getMessage(), e.getResponseBodyAsString(), e.getResponseHeaders()),
+                            true, NotificationMessage.Tag.submit);
+                    submitResponseBean.setNotification(notification);
+                    logNotification(notification, request);
+                    log.error("Submit response bean: " + submitResponseBean);
                     return submitResponseBean;
                 }
```
snopoke, line 218: This changed to use the normal rest template instead of the one that passed errors through. The only real change is that we're handling exceptions instead of checking response codes.

src/main/java/org/commcare/formplayer/screens/FormplayerSyncScreen.java
```diff
@@ -32,38 +34,27 @@ public void init (SessionWrapper sessionWrapper) throws CommCareSessionException
...
     }
 
-    public String getBuiltQuery() {
-        return builtQuery;
-    }
-
-    public void setBuiltQuery(String builtQuery) {
-        this.builtQuery = builtQuery;
+    public MultiValueMap<String, String> getQueryParams() {
+        return queryParams;
```
snopoke, line 54: let Spring serialize the data into `application/x-www-form-urlencoded` instead of doing it ourselves. We already do this elsewere in the code (https://github.com/dimagi/formplayer/blob/b4901366edc558af0157c29845cac3756cb56af3/src/main/java/org/commcare/formplayer/services/FormattedQuestionsService.java#L40-L44)

See Spring docs: https://docs.spring.io/spring-framework/docs/current/reference/html/integration.html#rest-template-multipart

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -322,20 +321,16 @@ private BaseResponseBean doSyncGetNext(FormplayerSyncScreen nextScreen,
...
+            webClient.post(screen.getUrl(), screen.getQueryParams());
+        } catch (RestClientResponseException e) {
             return new NotificationMessage(
-                    String.format("Case claim failed. Message: %s", responseEntity.getBody()), true, NotificationMessage.Tag.sync);
+                    String.format("Case claim failed. Message: %s", e.getResponseBodyAsString()), true, NotificationMessage.Tag.sync);
+        } catch (RestClientException e) {
+            return new NotificationMessage("Unknown error performing case claim", true, NotificationMessage.Tag.sync);
         }
+        restoreFactory.performTimedSync(false, false);
+        return new NotificationMessage("Case claim successful.", false, NotificationMessage.Tag.sync);
```
snopoke, line 333: I don't think this code really worked before since the RestTemplate would raise exceptions if the request was not successful

src/test/java/org/commcare/formplayer/tests/SubmitServiceTests.java
```diff
@@ -1,80 +0,0 @@
...
-import org.springframework.web.context.WebApplicationContext;
-
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Matchers.anyString;
-import static org.mockito.Mockito.when;
-import static org.springframework.test.web.client.match.MockRestRequestMatchers.requestTo;
-import static org.springframework.test.web.client.response.MockRestResponseCreators.withBadRequest;
-
-@RestClientTest(components = {SubmitService.class, RestTemplateConfig.class})
-public class SubmitServiceTests {
```
snopoke, line 30: This class was only testing that the RestTemplate didn't raise exceptions. Since it will now raise exceptions and those exceptions are now being handled it's of no use.

src/main/java/org/commcare/formplayer/screens/FormplayerSyncScreen.java
```diff
@@ -32,38 +34,27 @@ public void init (SessionWrapper sessionWrapper) throws CommCareSessionException
         Entry commandEntry = sessionWrapper.getPlatform().getEntry(command);
         if (commandEntry instanceof RemoteRequestEntry) {
             PostRequest syncPost = ((RemoteRequestEntry)commandEntry).getPostRequest();
-            setUrl(syncPost.getUrl().toString());
+            url = syncPost.getUrl().toString();
             Hashtable<String, String> params = syncPost.getEvaluatedParams(sessionWrapper.getEvaluationContext());
-            UriComponentsBuilder builder = UriComponentsBuilder.newInstance();
-            // hack because HQ isn't accepting the first query param key properly
-            builder.queryParam("buffer", "buffer");
```
calellowitz, line 39: Does one of your changes make this no longer necessary?

ctsims (approved):
Solid refactor, agree with the design choices.

Flagging @shubham1g5 as well since he's working with the Querying code, and might want to know that the refactor is coming.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -322,20 +321,16 @@ private BaseResponseBean doSyncGetNext(FormplayerSyncScreen nextScreen,
     }
 
     private NotificationMessage doSync(FormplayerSyncScreen screen) throws Exception {
-        ResponseEntity<String> responseEntity = syncRequester.makeSyncRequest(screen.getUrl(),
-                screen.getBuiltQuery(),
-                restoreFactory.getUserHeaders());
-        if (responseEntity == null) {
-            return new NotificationMessage("Session error, expected sync block but didn't get one.", true, NotificationMessage.Tag.sync);
```
calellowitz, line 329: Is the error that caused this message no longer possible?

calellowitz (approved):
Two questions mainly for my own benefit, otherwise looks great

src/main/java/org/commcare/formplayer/screens/FormplayerSyncScreen.java
snopoke: This was never really necessary, it was an artifact of the way this data was being sent (a query string masquerading as form data). Because this is using a query builder the data is rendered as `?buffer=buffer&....`. On CommCare's side the first param then is `?buffer` (note the question mark) which is why this had to be added. Now that we're letting the framework serialize the data it is coming through correctly (without the question mark).

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
snopoke: The response can never be null (there's code in RestTemplate to ensure that). In the case of errors an exception is raised so I don't think this case could ever have been triggered (certainly not under the current code configuration, perhaps previously if we were using the `errorPassthroughRestTemplate` but even then I think a `null` response is not allowed and a `IllegalStateException` would have been raised.

PR #915, opened 2021-03-05
use restTemplate to download ccz

snopoke:
Similar to https://github.com/dimagi/formplayer/pull/914, this makes use of the global rest template so that it can take advantage of the url re-writing in docker (Also it's just much cleaner).

### DIFF
--- a/src/main/java/org/commcare/formplayer/engine/FormplayerConfigEngine.java
+++ b/src/main/java/org/commcare/formplayer/engine/FormplayerConfigEngine.java
@@ -1,23 +1,19 @@
 package org.commcare.formplayer.engine;
 
-import org.commcare.formplayer.exceptions.ApplicationConfigException;
-import org.commcare.formplayer.exceptions.FormattedApplicationConfigException;
-import org.commcare.formplayer.installers.FormplayerInstallerFactory;
-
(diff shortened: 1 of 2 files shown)


calellowitz (approved):
looks good, definitely much cleaner. Can you explain the docker url rewriting. I am not sure I get how that is related

snopoke:
> looks good, definitely much cleaner. Can you explain the docker url rewriting. I am not sure I get how that is related

When we run formplayer in docker it has to communicate with CommCare via one of 2 URLs:
* `http://web:8000` in the case where commcare is running in docker
* `http://host.docker.internal:8000` in the case where commcare is running outside of docker (in the docker container this host is mapped to the docker network gateway which allows the container to talk to the network on the docker host machine).

In both of those cases the URLs that formplayer gets from of the app's `profile.ccpr` are incorrect (most likely `http://localhost:8000/...`) and so when formplayer tries to submit a form or do some other action with these URLs it can't. To work around the issue we register a custom request interceptor with the RestTemplate (only in docker env) which will replace the host / port of any requests with the correct host / port: https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/web/client/RewriteHostRequestInterceptor.java

PR #919, opened 2021-03-08
finalize form session cleanup

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-745

* Migrate any records with null values
* Remove old model fields
* Remove custom service methods and repository queries
* Update tests


src/main/java/org/commcare/formplayer/services/FormSessionService.java
```diff
@@ -47,34 +42,16 @@
     private StatsDClient datadogStatsDClient;
 
     @CacheEvict(allEntries = true)
+    @Transactional
```
ctsims, line 45: Currently the mechanism that calls these is getting killed before it logs, do we have any concerns about making this method call transactional like this? I'm a bit concerned that might result in long runs getting reverted.

ctsims (approved):

src/main/java/org/commcare/formplayer/services/FormSessionService.java
ctsims: That outcome seems fine to me, and I don't have a ton of clarity on exactly what is timing out currently. I know that we aren't receiving the Datadog notifications with the purge timing anymore, but that someone checked the AWS Cloudwatch logs and identified that purges were still completing, and it was presumed that this was due to the process taking too long.

Unfortunately I don't think purging more frequently can help meaningfully. We purge nightly to prevent running it during peak load, and since the record generation timing syncs up with the daily schedule we'd need to run a purge in the middle of the day to be able to "split the difference" between two cycles. I'm hopeful that the new index will help significantly, though!

PR #920, opened 2021-03-08
formplayer locale installer to always save locale data in the application DB

snopoke:
https://dimagi-dev.atlassian.net/browse/SAAS-11840

### DIFF
--- a/src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
+++ b/src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
@@ -70,10 +70,6 @@ public class FormplayerLocaleInstaller extends SimpleInstaller {
             problemList.addElement(new MissingMediaException(r, "Bad metadata, no locale",
                     MissingMediaException.MissingMediaExceptionType.NONE));
         }
-        if (localizedValues == null) {
-            problemList.addElement(new MissingMediaException(r, "Bad metadata, no locale values",
-                    MissingMediaException.MissingMediaExceptionType.NONE));
-        }
         return true;  // return value is never actually used
     }
 
--- a/src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
+++ b/src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
@@ -1,5 +1,6 @@
 package org.commcare.formplayer.installers;
 
+import lombok.NonNull;
 import org.commcare.resources.ResourceInstallContext;
 import org.commcare.resources.model.*;
 import org.commcare.util.CommCarePlatform;
@@ -40,7 +41,7 @@ public class FormplayerLocaleInstaller extends SimpleInstaller {
     public FormplayerLocaleInstaller() {
     }
 
-    public FormplayerLocaleInstaller(String locale) {
+    public FormplayerLocaleInstaller(@NonNull String locale) {
         this.locale = locale;
     }
 
@@ -66,11 +67,7 @@ public class FormplayerLocaleInstaller extends SimpleInstaller {
 
     @Override
     public boolean verifyInstallation(Resource r, Vector<MissingMediaException> problemList, CommCarePlatform platform) {
-        if (locale == null) {
-            problemList.addElement(new MissingMediaException(r, "Bad metadata, no locale",
-                    MissingMediaException.MissingMediaExceptionType.NONE));
-        }
-        return true;  // return value is never actually used
+        return true;
     }
 
     @Override
--- a/src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
+++ b/src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
@@ -1,0 +1,102 @@
+package org.commcare.formplayer.installers;
+
+import org.commcare.resources.ResourceInstallContext;
+import org.commcare.resources.model.*;
+import org.commcare.util.CommCarePlatform;
+import org.javarosa.core.reference.InvalidReferenceException;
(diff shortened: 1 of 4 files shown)


snopoke:
@ctsims is this all that's required?

src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public FormplayerLocaleInstaller(String locale) {
+        super(locale);
+    }
+
+    @Override
+    public boolean install(Resource r, ResourceLocation location, Reference ref, ResourceTable table, CommCarePlatform platform, boolean upgrade, ResourceInstallContext resourceInstallContext) throws UnresolvedResourceException {
+        try {
+            return cache(ref.getStream(), r, table, upgrade);
```
ctsims, line 31: I think we should probably change the name to something that reflects the new expectations more clearly, like "copyToInstaller" to reflect that this is a permanent part of the record now.

ctsims (approved):
Hey Simon,

I do think this looks like it will work. 

It has been a long time since we made any meaningful changes to the installers - Have you tested what the behavior is when the object used from the factory changes? I think old records still continue to work, but am not quite sure.

src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
snopoke: Are you thinking I should change that in the base class or in this class. The other option would be to create a completely new installer that either follow the same pattern as this or is more similar to the suite / profile installers.

src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
```diff
@@ -0,0 +1,102 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean verifyInstallation(Resource r, Vector<MissingMediaException> problemList, CommCarePlatform platform) {
+        if (locale == null) {
+            problemList.addElement(new MissingMediaException(r, "Bad metadata, no locale",
+                    MissingMediaException.MissingMediaExceptionType.NONE));
+        }
+        if (localizedValues == null) {
```
snopoke, line 82: not actually sure if this should be checked since and empty hashmap would end up being null

src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public FormplayerLocaleInstaller(String locale) {
+        this.locale = locale;
+    }
+
+    @Override
+    public boolean initialize(CommCarePlatform platform, boolean isUpgrade) throws IOException, InvalidReferenceException, InvalidStructureException, XmlPullParserException, UnfullfilledRequirementsException {
+        Localization.getGlobalLocalizerAdvanced().addAvailableLocale(locale);
+        Localization.getGlobalLocalizerAdvanced().registerLocaleResource(locale, new TableLocaleSource(localizedValues));
```
snopoke, line 50: probably need to check if this is null and if not pass in an empty hashmap

snopoke (commented):
same approach as before but re-implemented the functionality in a new class instead of overriding the existing class.

src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+            return true;
+        } catch (IOException e) {
+            throw new UnreliableSourceException(r, e.getMessage());
+        }
+    }
+
+    @Override
+    public boolean uninstall(Resource r, CommCarePlatform platform) throws UnresolvedResourceException {
+        if (localizedValues != null) {
```
ctsims, line 69: I don't think it's necessary to do anything with the table, here. Uninstallers are responsible for removing any persistent storage retained outside of the installer record, but they don't need to de-initialize the record or anything.

src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
ctsims: Agreed. `verifyInstallation` can presume that the record is fully installed, and an empty/null table or null locale would only be possible if the record was malformed, so it would be better for those to fail to install, rather than error during the check.

src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
ctsims: Since the installer is committed atomically with the resource record, I don't think there's a circumstance where the table could end up null after installed as long as `parseLocaleInput` passes back a valid table. Probably best to push any validation into the installer and fail fast if anything is off.

ctsims (changes requested):
Approach looks good, one or two minor comments on ambiguous sections

ctsims (approved):

src/main/java/org/commcare/formplayer/installers/FormplayerLocaleInstaller.java
```diff
@@ -0,0 +1,89 @@
...
+            }
+            return true;
+        } catch (IOException e) {
+            throw new UnreliableSourceException(r, e.getMessage());
+        }
+    }
+
+    @Override
+    public boolean verifyInstallation(Resource r, Vector<MissingMediaException> problemList, CommCarePlatform platform) {
+        if (locale == null) {
```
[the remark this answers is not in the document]
snopoke: 6fc8d651

PR #921, opened 2021-03-10
separate autosearch from skip default

calellowitz:
The previous logic mixed whether to do a search with whether include user input and default values in the search. This separates those two and attempts to clear up the naming.
https://dimagi-dev.atlassian.net/jira/servicedesk/projects/SUPPORT/queues/custom/81/SUPPORT-8316

Followup for 7839d7d4e5335762333a19d63a58f4909abed19a, restores the previous `autoSearch` flag that's passed to `doQuery`, while preserving the new logic that's used to determine whether or not to call `doQuery`.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -305,7 +305,7 @@ public class MenuSessionRunnerService {
                         (FormplayerQueryScreen)nextScreen,
                         menuSession,
                         queryData == null ? null : queryData.getInputs(queryKey),
-                        autoSearch
+                        formplayerQueryScreen.doDefaultSearch()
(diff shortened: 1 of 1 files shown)


orangejenny (approved):

snopoke:
would it be possible to make a test for this?

orangejenny:
@snopoke I think we'd want to modify the case claim tests so that [this](https://github.com/dimagi/formplayer/blob/17e5239deafd3de3d10f58d7a17fa33032cd5a12/src/test/java/org/commcare/formplayer/tests/CaseClaimTests.java#L165) generates the URL, even though it still wouldn't actually send the URL.

@shubham1g5 Any thoughts/suggestions?

shubham1g5:
It should be possible to verify the url argument to `webClientMock` using Mockito APIs, I will take a shot at it.

PR #923, opened 2021-03-10
add custom exception handler

snopoke:
This has been bugging me since I started working on formplayer. All generic spring errors now return plain responses instead of attempting to render error messages which it can't find.


snopoke:
Turns out this isn't necessary if we remove the error handling config we have which is broken (PR to follow)

PR #926, opened 2021-03-11
only log our views

snopoke:


src/main/java/org/commcare/formplayer/aspects/LoggingAspect.java
```diff
@@ -23,14 +23,16 @@
 
     private final Log log = LogFactory.getLog(LoggingAspect.class);
 
-    @Around(value = "@annotation(org.springframework.web.bind.annotation.RequestMapping) " +
+    @Around(value = "within(org.commcare.formplayer..*) " +
```
calellowitz, line 26: what is the result of this change?

src/main/java/org/commcare/formplayer/Application.java
```diff
@@ -44,25 +44,6 @@ public static void main(String[] args) {
...
-            final HttpServletRequest httpRequest = (HttpServletRequest) request;
-            final HttpServletResponse httpResponse = (HttpServletResponse) response;
-            log.info("Got request URL: " + httpRequest.getRequestURL() + " , response code: " + httpResponse.getStatus());
-            filterChain.doFilter(request, response);
-        }
-    }
-
-    // Autowire the filter above
-    @Bean
-    public Filter loggingFilter() {
```
calellowitz, line 62: is this not used anywhere?

calellowitz (approved):
I don't totally understand what the effect of this changes but the individual parts seem fine. nothing blocking, but would be great to get a fuller explanation

src/main/java/org/commcare/formplayer/Application.java
snopoke: We have 2 logging filters and one logging aspect:

1. ResponseLoggingFilter (which is the one removed here)
  - A very generic logging filter that just outputs the url and 200 (the status code is always 200 since the logging happens before the request is processed)
2. RequestResponseLoggingFilter (enabled if `sensitiveData.enableLogging` is true)
  - logs the full request and response
3. LoggingAspect (runs around every view method call)
  - logs basic info about the request

I don't know why we have all these different loggers. (1) is clearly of very little use which is why I'm removing it.

src/main/java/org/commcare/formplayer/aspects/LoggingAspect.java
snopoke: This changes how the aspect so that it only matches methods that are in or below the `org.commcare.formplayer` package. Specifically this should not run on any of the built in Spring views like `/error` since the do not have the expected method arguments.

PR #929, opened 2021-03-12
Manage MenuSession via Spring Jpa

sravfeyn:
https://dimagi-dev.atlassian.net/browse/USH-667


sravfeyn:
@snopoke This is mostly ready except for tests.

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -1,30 +1,61 @@
...
+@Entity
+@Table(name="menu_sessions")
+@EntityListeners(AuditingEntityListener.class)
 public class SerializableMenuSession {
 
+    @Id @GeneratedValue(generator="system-uuid")
+    @GenericGenerator(name="system-uuid", strategy = "uuid")
     private String uuid;
+
+    @Column(name="username")
```
snopoke, line 25: these annotations are only needed if you want to change the defaults. The currently configured naming strategy will:
- Replace dots with underscores
- Change camel case to snake case, and
- Lower-case table names

This doesn't always match up with what the table / column names are in the DB. If that's the case then we need to use the annotation to customize it. For columns like this `username` there are no changes from the default so we can leave it off altogether.

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -1,30 +1,61 @@
...
     private String username;
+
+    @Column(name="domain")
     private String domain;
+
+    @Column(name="appId")
     private String appId;
+
+    @Column(name="installReference")
```
snopoke, line 34: change to match what's in the DB:
```suggestion
    @Column(name="installreference")
```

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -1,30 +1,61 @@
...
     private String installReference;
+
+    @Column(name="locale")
     private String locale;
+
+    // Todo: Will this be large, should this be @Lob?
+    @Column(name="commcareSession")
```
snopoke, line 41:
```suggestion
    @Column(name="commcaresession")
```

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -1,30 +1,61 @@
...
+    @Column(name="appId")
     private String appId;
+
+    @Column(name="installReference")
     private String installReference;
+
+    @Column(name="locale")
     private String locale;
+
+    // Todo: Will this be large, should this be @Lob?
```
snopoke, line 40: if it's worked up till now it will probably continue to work. The session is usually very small.

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -1,30 +1,61 @@
...
+    @Column(name="commcareSession")
     private byte[] commcareSession;
+
+    @Column(name="asUser")
```
snopoke, line 44:
```suggestion
    @Column(name="asuser")
```

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -1,30 +1,61 @@
...
     private String asUser;
+
+    @Column(name="oneQuestionPerScreen")
```
snopoke, line 47: interestingly this isn't a persisted field, looking at it's usage I think it can be removed form this class altogether and kept purely in the `MenuSession` class. The only time this might break is if there is end of form navigation that advances to another form in which case I don't think the `oneQuestionPerScreen` value will get passed on. I don't think that's worth worrying about at this point.

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -91,7 +91,6 @@ public MenuSession(String username, String domain, String appId, String locale,
                        boolean oneQuestionPerScreen, String asUser, boolean preview) throws Exception {
         String resolvedInstallReference = resolveInstallReference(appId, host, domain);
         this.session = new SerializableMenuSession(
-                UUID.randomUUID().toString(),
```
snopoke, line 94: with this being auto-generated on save we have to make sure it's not used before the save happens.

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -1,30 +1,61 @@
...
+
+import javax.persistence.*;
 
 /**
  * Created by willpride on 8/1/16.
  */
+
+@Entity
+@Table(name="menu_sessions")
+@EntityListeners(AuditingEntityListener.class)
```
snopoke, line 18: this isn't needed unless you're going to add a creation / modification date. I do think a creation date would be useful and would allow us to clean up this data better.

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -1,30 +1,61 @@
...
 public class SerializableMenuSession {
```
snopoke, line 19: based on how the update was written (https://github.com/dimagi/formplayer/blob/0a6850fdbeb61d8c5dedcda4c4c338e350fbd275/src/main/java/org/commcare/formplayer/repo/impl/PostgresMenuSessionRepo.java#L49-L51) it looks like the only thing that can change / needs to be saved to the DB is the `commcareSession`. 

If that's the case you can annotate all the other fields with `@Column(updatable=false)`. You can also remove all the "setters" for those fields.

Side note: you could delete all the getters and setters and then let [lombok](https://projectlombok.org/features/GetterSetter) generate them for you (annotate the class with `@Getter` and only annotate `commcareSession` with `@Setter`)

snopoke (changes requested):
Looking good. Made a few comments.

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
sravfeyn: Thanks for the very detailed note, Simon!

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
sravfeyn: Can we also simply use `@Transient` instead of moving it somewhere else?  https://docs.oracle.com/javaee/7/api/javax/persistence/Transient.html

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
sravfeyn: > If that's the case you can annotate all the other fields with @Column(updatable=false). You can also remove all the "setters" for those fields.

Just curious, what't the advantage of marking something `updatable=false`?

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
snopoke: yes we could do that though in the long run it would be better to move it out

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
snopoke: > Just curious, what't the advantage of marking something updatable=false?

it won't be included in the SQL update statement when saving is better for network and DB

src/main/java/org/commcare/formplayer/session/MenuSession.java
snopoke: @sravfeyn I haven't verified yet if this is an issue - just need to check where this ID is used outside the object in comparison to where save happens

src/main/java/org/commcare/formplayer/objects/SerializableMenuSession.java
```diff
@@ -10,6 +12,7 @@
 
 @Entity
 @Table(name="menu_sessions")
+@Getter
```
sravfeyn, line 15: 💯

sravfeyn:
Hey @snopoke, the build is [failing](https://travis-ci.com/github/dimagi/formplayer/builds/220693924) with below error. 
```
       org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'menuSessionRepo' defined in org.commcare.formplayer.repo.MenuSessionRepo defined in @EnableJpaRepositories declared on Application: Cannot create inner bean '(inner bean)#7426632d' of type [org.springframework.orm.jpa.SharedEntityManagerCreator] while setting bean property 'entityManager'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name '(inner bean)#7426632d': Cannot resolve reference to bean 'entityManagerFactory' while setting constructor argument; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named 'entityManagerFactory' available
```
I think it's related to the config here, may be we need to set `entityManagerFactoryRef` but I wasn't able to figure out what it needs to be set to. 

(There is a different build failure on my local machine, but I will retry after this is addressed)

snopoke:
> Hey @snopoke, the build is [failing](https://travis-ci.com/github/dimagi/formplayer/builds/220693924) with below error.
> 
> ```
>        org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'menuSessionRepo' defined in org.commcare.formplayer.repo.MenuSessionRepo defined in @EnableJpaRepositories declared on Application: Cannot create inner bean '(inner bean)#7426632d' of type [org.springframework.orm.jpa.SharedEntityManagerCreator] while setting bean property 'entityManager'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name '(inner bean)#7426632d': Cannot resolve reference to bean 'entityManagerFactory' while setting constructor argument; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named 'entityManagerFactory' available
> ```
> 
> I think it's related to the config here, may be we need to set `entityManagerFactoryRef` but I wasn't able to figure out what it needs to be set to.
> 
> (There is a different build failure on my local machine, but I will retry after this is addressed)

resolved with https://github.com/dimagi/formplayer/pull/929/commits/8d2bbcab8dc56a3f7af1acef99e37d753204b26c

sravfeyn:
For QA, do we simply keep it on staging or do we need to trigger any active QA plan? @shubham1g5

sravfeyn:
> resolved with 8d2bbca

Thank you! @snopoke How did you correlate that error to this change?

snopoke:
> > resolved with [8d2bbca](https://github.com/dimagi/formplayer/commit/8d2bbcab8dc56a3f7af1acef99e37d753204b26c)
> 
> Thank you! @snopoke How did you correlate that error to this change?

Just based on the tests that were impacted I did a manual inspection and saw that they weren't configured to use a DB.

snopoke:
> For QA, do we simply keep it on staging or do we need to trigger any active QA plan? @shubham1g5

I would think the formplayer regression test suite would be sufficient. The QA team knows what that is

src/main/java/org/commcare/formplayer/session/MenuSession.java
snopoke: @sravfeyn have you looked into this yet?

sravfeyn: Yeah, I reviewed in various places where `SerializableMenuSession` is being instantiated or `saveSession` is called and I haven't find any places where the ID is being used.

sravfeyn:
I have raised a QA request for this https://dimagi-dev.atlassian.net/browse/QA-2806

src/main/java/org/commcare/formplayer/session/MenuSession.java
snopoke: and is the model always saved before it is serialized for the response?

sravfeyn:
QA passed on this @snopoke

snopoke (approved):

PR #930, opened 2021-03-12
Incomplete forms pagination support

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-699

Adds pagination support to the incomplete forms screen. 
Adds  `page_size` and `page_number` to the request and `total_records` in response.


src/test/java/org/commcare/formplayer/services/FormSessionsPaginationTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import java.util.List;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration
+@DataJpaTest
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@EnableJpaAuditing
+public class FormSessionsPaginationTest {
```
shubham1g5, line 33: @snopoke I am struggling with getting this test to be configured correctly. I am getting an error `java.lang.IllegalStateException: Unable to retrieve @EnableAutoConfiguration base packages` with the current config. I don't necessarily understand all the annotations used here fully and wondering if you would be able to take a quick look and see if I am doing something obviously wrong.

snopoke: What is it you're trying to test? Can you not add tests to the existing `FormSessionServiceTest` class?

shubham1g5: I can though I am trying to avoid mocking `FormSessionRepo` in order to test the pagination behaviour of the repo. Though I think it would be also alright just to test our wiring aka `FormSessionService` and mock the `FormSessionRepo` pagination ?

snopoke: Up till now I've tested the two independently, the repo I've tested with actual DB access and the service has used a mock repo. I don't think the's a lot of value in testing them together but I could be convinced otherwise. 

If you do want to try set up a test with DB access I would try just copying the pattern of the `FormSessionRepoTest` and autowiring in the service. I can give that a go if you don't have any luck.

shubham1g5: I gave it another shot but facing the same `java.lang.IllegalStateException: Unable to retrieve @EnableAutoConfiguration base packages` issue while trying to autowire `FormSessionService` . On top of `FormSessionRepoTest` I added the `@ExtendWith(SpringExtension.class)` and `@ContextConfiguration` to indicate that this class should provide configuration and then added the config below to provide `formSessionService` - 

```
 @ComponentScan(
            basePackageClasses = {FormSessionService.class},
            useDefaultFilters = false,
            includeFilters = @ComponentScan.Filter(type = FilterType.ASSIGNABLE_TYPE, classes = {FormSessionService.class})
    )
    @EnableCaching
    @Configuration
    public static class FormSessionServiceTestConfig { }
```

snopoke: fixed in 2f54124e. You were missing the requirements for the FormSessionRepo

shubham1g5: Ahh I see, thanks a lot for looking at this. I think I was missing `@EnableJpaRepositories` and `@EntityScan` when I tried adding the same.

src/test/java/org/commcare/formplayer/services/FormSessionsPaginationTest.java
```diff
@@ -0,0 +1,75 @@
...
+            formSessionService.saveSession(session);
+        }
+        FormsSessionsRequestBean formsSessionsRequestBean = new FormsSessionsRequestBean();
+        formsSessionsRequestBean.setUsername("momo");
+        formsSessionsRequestBean.setDomain("domain");
+        formsSessionsRequestBean.setRestoreAs("momo");
+        formsSessionsRequestBean.setPageSize(5);
+        formsSessionsRequestBean.setOffset(15);
+        List<FormSessionListView> formSessions = formSessionService.getSessionsForUser("momo", formsSessionsRequestBean);
+        assert formSessions.size() == 3;
```
sravfeyn, line 55: It will be nice to test that the right records are appearing in the right page-number

sravfeyn (approved):

snopoke (approved):

PR #931, opened 2021-03-15
implement spring security filters for auth

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-686

Functionally this is the same as the current setup but it makes use of the Spring Security framework which provides a lot useful features for working with authentication and authorization.

The main entry point is the [WebSecurityConfig](https://github.com/dimagi/formplayer/pull/931/files#diff-f62bd5ffdadb5c9b1c7e43b268f544410a0c1c609d27000154c45efd6d077a4b) class which is responsible for configuring the security policy. 

More details is included in the [DEV_GUIDE](https://github.com/dimagi/formplayer/pull/931/files#diff-fdf44949339fc92e5bcfe6987dccba1046fe5b44bb17249c13654ad7d5ff460a).

Can be reviewed by commit or as a whole.

Requires: https://github.com/dimagi/commcare-cloud/pull/4584 to be rolled out before deploy.


src/main/java/org/commcare/formplayer/application/FormplayerSentryFilter.java
```diff
@@ -15,7 +14,7 @@
 import java.io.IOException;
 
 @Component
-@Order(Ordered.HIGHEST_PRECEDENCE + 2)
+@Order(1)
```
ctsims, line 17: This is set to the same value as the RequestResponseLoggingFilter below, is that intentional?

Even if we no longer need strict ordering, don't we still not want to be enforcing an order on these two filters for consistency?

src/main/java/org/commcare/formplayer/auth/CommCareSessionAuthFilter.java
```diff
@@ -0,0 +1,60 @@
...
+        if (HttpMethod.valueOf(request.getMethod()) == HttpMethod.POST) {
+            try {
+                JSONObject data = RequestUtils.getPostData(request);
+                String username = getUsername(data);
+                String domain = data.getString("domain");
+                if (username != null && domain != null) {
+                    return new UserDomainPreAuthPrincipal(username, domain);
+                }
+            } catch (Exception e){
+                return null;  // TODO: should sentry capture this?
```
ctsims, line 46: It definitely seems like this should be a typed error rather than a null response. I shudder trying to imagine debugging this.

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private HmacAuthFilter getHmacAuthFilter() {
+        return HmacAuthFilter.builder()
+                .hmacKey(formplayerAuthKey)
+                .formSessionService(formSessionService)
+                .build();
+    }
+
+    private void disableDefaults(HttpSecurity http) throws Exception {
+        http.csrf().disable();
```
ctsims, line 54: Is this matching our existing defaults somehow? I didn't see an obvious place where it was being copied from the old code config.

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    protected void configure(HttpSecurity http) throws Exception {
+        disableDefaults(http);
+
+        http
+            .authorizeRequests()
+            .antMatchers("/serverup", "/validate_form", "/favicon.ico")
+            .permitAll();
+        http.authorizeRequests().antMatchers("/**").authenticated();
+        http.addFilterAt(getHmacAuthFilter(), UsernamePasswordAuthenticationFilter.class);
```
ctsims, line 41: I'm assuming this is why the ordering got removed above?

src/main/java/org/commcare/formplayer/util/RequestUtils.java
```diff
@@ -57,18 +62,19 @@ public static String getBody(HttpServletRequest request) throws IOException {
...
+            throw new RuntimeException("Unreadable POST Body for the request: " + request.getRequestURI(), a);
         }
         return data;
     }
 
-    public static String getRequestEndpoint(HttpServletRequest request) {
-        return StringUtils.strip(request.getRequestURI(), "/");
+    public static String getRequestEndpoint() {
+        HttpServletRequest request = RequestUtils.getCurrentRequest();
+        return request == null ? "unknown" : StringUtils.strip(request.getRequestURI(), "/");
```
ctsims, line 77: Is this method just informational / user facing? It seems like this is potentially dangerous to infer a real string from

src/main/java/org/commcare/formplayer/application/FormplayerSentryFilter.java
snopoke: We could chose and arbitrary order, any suggestions on which should go first?

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
snopoke: We didn't have any CSRF before. If we want to enable it we'd need to implement it on this side as well as the UI since there's currently no support for getting or sending CSRF tokens in Web Apps.

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
snopoke: not sure what ordering you mean, the comment above was about the logging filter and the sentry context filter which are both set to "1" which is immediately after the security filter chain.

The ordering of these filters is set to preserve the order of the processing in the `FormplayerAuthFilter`. I don't think the ordering matters but since the HMAC filter can tell if it should be applied by looking for the header it did make sense to put that first.

src/main/java/org/commcare/formplayer/util/RequestUtils.java
snopoke: this value is only used for metric tag values. What do you mean about "infering a real string"?

src/main/java/org/commcare/formplayer/auth/CommCareSessionAuthFilter.java
snopoke: the null response is intentional and an expected return value if the method was unable to extract the necessary information from the request (e.g. if the body wasn't JSON or didn't contain the `username` or `domain` key).

In the case of returning a `null` this filter will not put an authenticated user into the security context and ultimately the request will be rejected (if it is requesting a URL that requires auth) unless there is another filter in the chain that can perform the auth.

I add logging for the error though: df6f00ae

src/main/java/org/commcare/formplayer/application/FormplayerSentryFilter.java
snopoke: decided to put Sentry first: c31bbf10

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Temporary customization of the firewall to allow '//' in the URL.
+     * This should be removed once all 3rd party hosters have been given an opportunity
+     * to update their proxy configuration.
+     */
+    @Bean
+    public HttpFirewall allowDoubleSlashHttpFirewall() {
+        StrictHttpFirewall firewall = new StrictHttpFirewall();
+        firewall.setAllowUrlEncodedDoubleSlash(true);
```
snopoke, line 96: I've added this so that we can decouple this PR from the proxy changes in https://github.com/dimagi/commcare-cloud/pull/4584

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
snopoke: https://dimagi-dev.atlassian.net/browse/USH-827

src/main/java/org/commcare/formplayer/application/FormplayerSentryFilter.java
ctsims: Logging first makes sense to me. thanks!

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
ctsims: Got it, thanks for clarifying!

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
ctsims: Thanks, wasn't sure if these became the explicit ordering, or whether the system was still respecting the Annotations. Small points either way, just wanted to make sure I wasn't missing anything.

src/main/java/org/commcare/formplayer/util/RequestUtils.java
ctsims: All good if it's just metric tags.

If this was, say, used for request routing, I was concerned that there might be a scenario where providing a real value (`unknown`) when something is missing could be a sidechannel into routing if there was a real potential `unknown`  output from the call.

ctsims (approved):

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
snopoke: Yea, it is a bit confusing since the security chain is a set of filters that are run 'as a unit' in the main filter chain. This diagram illustrates it well: https://docs.spring.io/spring-security/site/docs/5.4.5/reference/html5/#servlet-securityfilterchain

PR #934, opened 2021-03-15
remove default logging config

snopoke:
It turns out that logging config will take the lowest level from all the
available configurations (app properties, logging config etc)


ctsims:
@snopoke Can you clarify this a bit? is there another config that isn't one of these files that's overriding them in practice, defeating the point of defining them different within them?

snopoke:
> @snopoke Can you clarify this a bit? is there another config that isn't one of these files that's overriding them in practice, defeating the point of defining them different within them? Is the expected outcome from a cluster in terms of what is logged the same before and after this PR?

There are 3 places we can configure logging:

1. application.properties file inside the JAR
2. application.properties outside the JAR
3. logback config file

Prior to [this change](https://github.com/dimagi/formplayer/pull/909/files#diff-54eeffbae371fcd1398d4ca5e89a1b8118208b7bb2f8ddf55c1aa2f7d98ab136R55-R59) the logging was only configured in 2 and 3 (2 was used locally and 3 was used in production).

My understanding at the time was that the order of precedence would be taken into account when applying the log level configuration but this appears to not be the case and instead the final logging level will be the minimum from any of the configured values. This has resulted in the level of the `org.commcare` (and child) loggers to have a value of `TRACE` in all environments with no way to change it.

This PR removes (1) so that the log levels can be fully controlled in the environment where the application is running.

PR #936, opened 2021-03-16
AutoLaunch cleanup

shubham1g5:
Very minor cleanup that should not result into any changes today with auto launch functionality but can potentially be harmful if remained unaddressed.

### DIFF
--- a/src/main/java/org/commcare/formplayer/session/MenuSession.java
+++ b/src/main/java/org/commcare/formplayer/session/MenuSession.java
@@ -153,8 +153,8 @@ public class MenuSession implements HereFunctionHandlerListener {
             boolean addBreadcrumb = true;
             if (screen instanceof EntityScreen) {
                 EntityScreen entityScreen = (EntityScreen)screen;
-                boolean autoLaunch = entityScreen.getAutoLaunchAction() != null;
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -153,8 +153,8 @@ public boolean handleInput(String input, boolean needsDetail, boolean confirmed,
             boolean addBreadcrumb = true;
             if (screen instanceof EntityScreen) {
                 EntityScreen entityScreen = (EntityScreen)screen;
-                boolean autoLaunch = entityScreen.getAutoLaunchAction() != null;
-                addBreadcrumb = !(autoLaunch && allowAutoLaunch);
+                boolean autoLaunch = entityScreen.getAutoLaunchAction() != null && allowAutoLaunch;
+                addBreadcrumb = !autoLaunch;
                 if (input.startsWith("action ") || (autoLaunch) || !confirmed) {
```
shubham1g5, line 158: Today the only place where `allowAutoLaunch` is false , `confirmed` is also set to `false` resulting in this line evaluating to `true` even when we don't factor in `allowAutoLaunch` in the `autoLaunch` flag.

orangejenny (approved):

PR #944, opened 2021-03-29
add basic timing to answerQuestion

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-826

This adds very basic timings to `answerQuestion` requests via Sentry breadcrumbs.

To get any more granularity will require one of the following:
1. adding sentry to the core library
2. moving code out of core
3. adding a layer of abstraction to allow timing in core without needing Sentry directly

(3) is likely the only real option


src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -130,19 +127,44 @@ public FormEntryResponseBean changeLocale(@RequestBody ChangeLocaleRequestBean c
...
+                Constants.TimingCategories.PROCESS_ANSWER,
+                () -> formEntrySession.answerQuestionToJSON(
+                        answerQuestionBean.getAnswer(), answerQuestionBean.getFormIndex()
+                )
+        );
+
+        FormplayerSentry.timedBreadcrumb(
+                Constants.TimingCategories.GET_SESSION,
+                () -> formSessionService.getSessionById(answerQuestionBean.getSessionId())
+        );
```
gherceg, line 155: Is this duplicate timing necessary for `GET_SESSION` considering the same call is timed a few lines earlier?

gherceg (approved):
One non-blocking comment. Looks good.

src/main/java/org/commcare/formplayer/application/FormController.java
snopoke: hmm, no, not sure why that's there.

snopoke, line 155:
```suggestion
```

PR #945, opened 2021-03-30
prune formplayer metrics to reduce cost + remove unused

snopoke:
* Remove `formplayer.metrics.requests` entirely and update all usages with `formplayer.metrics.timings.count`
* Changes to `formplayer.metrics.timings.*` tags
    * Only add the 'domain' tag to for requests > 1s
    * Drop the following tags altogether: "unblocked_time", "blocked_time", "restore_blocked_time", "install_blocked_time", "submit_blocked_time"        
        * These tags were either unused or can be replaced with the granular timings metrics.


gherceg (approved):
Just need to fix the tests. I probably shouldn't have referenced a constant in tests in hindsight so sorry about that. Thanks for cleaning up 👍

PR #948, opened 2021-03-31
Break out of case purge in case of cycles

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-818

In case of case purge failing due to cycles in case graph, forces a fresh sync in order to try resolve the cycles and In case of no resolution throws an error. 

cross-request: https://github.com/dimagi/commcare-core/pull/985

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/FormController.java
+++ b/src/main/java/org/commcare/formplayer/application/FormController.java
@@ -214,8 +214,10 @@ public class FormController extends AbstractBaseController{
                     ).getPurgeCasesTimer();
                 } catch (InvalidCaseGraphException e) {
                     submitResponseBean.setStatus(Constants.SUBMIT_RESPONSE_CASE_CYCLE_ERROR);
-                    NotificationMessage notification = new NotificationMessage("Form submission failed due to a cyclic case relationship",
-                            true, NotificationMessage.Tag.submit);
+                    NotificationMessage notification = new NotificationMessage(
+                            "Form submission failed due to a cyclic case relationship. Please contact the support desk to help resolve this issue.",
+                            true,
+                            NotificationMessage.Tag.submit);
                     submitResponseBean.setNotification(notification);
                     logNotification(notification, request);
                     log.error("Submission failed with structure exception " + e);
--- a/src/main/java/org/commcare/formplayer/application/FormController.java
+++ b/src/main/java/org/commcare/formplayer/application/FormController.java
@@ -213,7 +213,13 @@ public class FormController extends AbstractBaseController{
                             storageFactory.getPropertyManager().isAutoPurgeEnabled()
                     ).getPurgeCasesTimer();
                 } catch (InvalidCaseGraphException e) {
-                    log.error("Case purge failed during form submission due to " + e.getMessage());
+                    submitResponseBean.setStatus(Constants.SUBMIT_RESPONSE_CASE_CYCLE_ERROR);
+                    NotificationMessage notification = new NotificationMessage("Form submission failed due to a cyclic case relationship",
+                            true, NotificationMessage.Tag.submit);
+                    submitResponseBean.setNotification(notification);
+                    logNotification(notification, request);
+                    log.error("Submission failed with structure exception " + e);
+                    return submitResponseBean;
                 }
 
                 categoryTimingHelper.recordCategoryTiming(purgeCasesTimer, Constants.TimingCategories.PURGE_CASES,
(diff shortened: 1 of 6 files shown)


src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -200,14 +201,20 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
...
+                    purgeCasesTimer = FormRecordProcessorHelper.processXML(
+                            new FormplayerTransactionParserFactory(
+                                    restoreFactory.getSqlSandbox(),
+                                    storageFactory.getPropertyManager().isBulkPerformanceEnabled()
+                            ),
+                            formEntrySession.submitGetXml(),
+                            storageFactory.getPropertyManager().isAutoPurgeEnabled()
+                    ).getPurgeCasesTimer();
+                } catch (InvalidCaseGraphException e) {
+                    log.error("Case purge failed during form submission due to " + e.getMessage());
```
ctsims, line 216: Does the purge occur *after* the form is processed in this case? I feel like if that's the case, we should avoid submitting the form to HQ, and error out of the process as if the form failed validation after rolling back the changes.

This would prevent formplayer from submitting up a form which likely creates a cycle, preventing the need to be able to deal with it asynchronously on HQ's side (It can't catch every case, but might catch most of them).

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -173,49 +176,56 @@ public void configure(AuthenticatedRequestBean authenticatedRequestBean, HqAuth
...
     }
 
     // This function will only wipe user DBs when they have expired, otherwise will incremental sync
     public UserSqlSandbox performTimedSync() throws Exception {
-        return performTimedSync(true, false);
+        return performTimedSync(true, false, false);
     }
-    public UserSqlSandbox performTimedSync(boolean shouldPurge, boolean skipFixtures) throws Exception {
+
+    public UserSqlSandbox performTimedSync(boolean shouldPurge, boolean skipFixtures, boolean responseTo412) throws Exception {
```
ctsims, line 188: 'response' is a verb, so I found this name slightly confusing, can you change it to "isResponseTo412"?

ctsims (changes requested):
A few minor-ish points of feedback, but approach looks good.

src/main/java/org/commcare/formplayer/application/FormController.java
shubham1g5: @ctsims That does seem like a good suggestion.The purge does happen after we have processed the form. Though I am a bit confused how would we rollback the changes caused by form processing locally once the form processing is done ?

ctsims:
One last small point of feedback: Can we make sure a Sentry message gets posted if there is a cycle detected that causes a 412? 

I want to make sure that these don't end up being "silent recoveries" that prevent us from seeing the larger problem.

src/main/java/org/commcare/formplayer/application/FormController.java
ctsims: Since autocommit is disabled, I believe you can execute the following to rollback changes to the sqlite db

```
restoreFactory.getSQLiteDB().getConnection().rollback();
```

that call should really get abstracted *into* the restoreFactory, though.

calellowitz: We could set the status on the response that indicates the form was invalid because it creates a cycle and early exit. And then that would be displayed to the user here

ctsims: Ah, Cal is correct, it looks like as long as this method exits before committing, it'll automatically roll back the transacitons

https://github.com/dimagi/formplayer/pull/948/files/fd4b7d5ab86c073b950e53bdc80eea60b73716f7#diff-d44694a41dfd4c703a35f74e71c28499ca3bb16b86ad10e9353c6737ae39cf91R260

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -218,6 +220,7 @@ public UserSqlSandbox performTimedSync(boolean shouldPurge, boolean skipFixtures
                         extras
                 );
             } catch (InvalidCaseGraphException e) {
+                FormplayerSentry.captureException(e, SentryLevel.INFO);
```
ctsims, line 223: I think this might be a warning level event since we think it should never happen.

ctsims (approved):

ctsims (approved):

PR #949, opened 2021-04-01
record DD metrics for answer timings

snopoke:
[USH-826](https://dimagi-dev.atlassian.net/browse/USH-826)

Publish DD metrics in addition to recording the timings as Sentry breadcrumbs.

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/FormController.java
+++ b/src/main/java/org/commcare/formplayer/application/FormController.java
@@ -6,7 +6,6 @@ import io.sentry.Sentry;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.commcare.cases.util.InvalidCaseGraphException;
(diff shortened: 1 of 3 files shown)


gherceg (approved):

gherceg:
Do you think it is worth keeping the unused `timedBreadcrumb` methods around in `FormplayerSentry` with the exception of `public static void timedBreadcrumb(Timing timing, String category, String sentryMessage)`? (pretty sure they are now unused)
https://github.com/dimagi/formplayer/blob/28d075f7a3a91db2d6c93dab669f91d05c78df29/src/main/java/org/commcare/formplayer/util/FormplayerSentry.java#L86-L104

I can imagine a use case where we only want timings for sentry, but just wanted to bring it up in case you see a reason to remove them.

snopoke:
@gherceg I've removed the sentry methods. We can add them back if necessary but it seems like we'll mostly want to do both DD and sentry.

PR #960, opened 2021-04-29
Auto advance Menu screen with single option

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-636

Auto advances any menu screen with a single option if the property `cc-auto-advance-menu` is enabled. 

It was hard to test this feature without fixes in `caseSearchSkip` due to which I ended up merging `caseSearchSkip` in the current branch

cross-request: https://github.com/dimagi/commcare-core/pull/993

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -33,6 +33,9 @@ public class MenuSessionFactory {
     @Autowired
     private InstallService installService;
 
+    @Autowired
+    protected FormplayerStorageFactory storageFactory;
+
     @Value("${commcarehq.host}")
     private String host;
 
@@ -78,7 +81,7 @@ public class MenuSessionFactory {
             if (currentStep == null) {
                 break;
             } else {
-                menuSession.handleInput(currentStep, false, true, false);
+                menuSession.handleInput(currentStep, false, true, false, storageFactory.getPropertyManager().isAutoAdvanceMenu());
                 menuSession.addSelection(currentStep);
                 screen = menuSession.getNextScreen(false);
             }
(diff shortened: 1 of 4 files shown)


orangejenny (approved):

ctsims:
Is there a specific reason why we went with an implementation here that was app specific for every menu, rather than menu specific? It seems like that's a very, very broad behavior and I'm worried we'll end up special casing the other direction (suppressing the behavior on some menus) which will mean that we've added quite significant complexity to the underlying logic.

orangejenny:
@ctsims It was to reduce complexity/confusion around how this setting would interact with display-only forms and possibly also parent/child menus. Historically we've had buggy and ill-defined behavior around those settings.

orangejenny:
@ctsims [This comment](https://dimagi-dev.atlassian.net/browse/USH-636?focusedCommentId=136674) might be of interest.

ctsims (approved):
@orangejenny 

Got it. The PR itself looks good

I'm still pretty uneasy in general with continuing to special case a lot of these new navigation approaches. Each discrete decision seems like it makes sense within the limitations presented, but to your point, we've found a lot of edge cases in these areas in the past.

I'd recommend reaching out if y'all run across additional limitations or challenges to work around with this pattern to see if there's maybe a higher level place we can be encoding this unit.

orangejenny:
@ctsims That's fair. For this particular feature, I'd love to just make it the default behavior, but it would be a very noticeable change. Do you think that's a reasonable avenue to consider?

ctsims:
@orangejenny I think we can't consider that approach without (among other things) also making that the default approach for mobile, since I think maintaining execution parity by default is a central design constraint for the platform.

That would make it difficult to imagine pursuing that change on any practical timeframe for USH, although if there was consensus between the platform stakeholders that the broader behavior change (always select a menu item if there's only one option) was worth considering, it would make the flag here unobjectionable. It might be worth considering whether USH would be willing to tweak the behavior of this flag if there are changes to the approach that other divisions would want to make for such a pilot / investigation.

PR #963, opened 2021-05-06
pass down templateforms list in getDetails for markdown

knguyen142:
https://dimagi-dev.atlassian.net/browse/SAAS-11685
and QA https://dimagi-dev.atlassian.net/browse/QA-2976?focusedCommentId=141272

Needs to pass down the list of case list property format. In the code this is called `templateForms` so I used that naming, but should we name it to something else that is more understandable for the frontend? 

Tested locally:

Screen Shot 2021-05-06 at 6 19 16 PM

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
@@ -59,7 +59,7 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
                         detailList[i],
                         subContext,
                         titles);
-                EntityDetailResponse response = new EntityDetailResponse(subscreen, titles[i]);
+                EntityDetailResponse response = new EntityDetailResponse(subscreen, titles[i], detailList[i]);
                 accumulator.add(response);
             } else {
                 TreeReference contextualizedNodeset = detailList[i].getNodeset().contextualize(ref);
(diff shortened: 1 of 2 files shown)


ShivamPokhriyal (approved):
Looks good to me.

ctsims (approved):

gherceg (approved):
I find `templateForms` not very intuitive, but I also haven't spent much time in this codebase.
>Needs to pass down the list of case list property format
Something with `formats` in the name seems to make a bit more sense to me (e.g., `templateFormats`, `propertyFormats`, etc) but not blocking.

orangejenny:
I like "format" better than "form", which is already overloaded. So `templateFormats` sounds good.

knguyen142:
Just curious, what does `template` mean in this context and does that formplayer naming need to be conserved to webapps?

orangejenny:
@knguyen142 Template is a term used by the suite that describes the detail field's content. It usually contains some kind of string or calculation, depending on the format. The detail spec is documented [here](https://github.com/dimagi/commcare-core/wiki/Suite20#detail) 

...Looking at those docs, I was mis-remembering how the suite file is configured. The variable should be called `templateForms` after all, since it's literally sending back the value of the `form` attribute of the templates.

knguyen142:
@orangejenny thanks, so you think we should keep it as `templateForms` instead of renaming it to something that is still semantically the same but more frontend-friendly?

orangejenny:
Yeah, I'd keep `templateForms`, the web apps UI doesn't need its own term.

PR #964, opened 2021-05-19
Feature flag checker for Formplayer, Kotlin Support

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-807

cross-request: https://github.com/dimagi/formplayer/pull/964

### DIFF
--- a/src/main/java/org/commcare/formplayer/auth/HmacAuthFilter.java
+++ b/src/main/java/org/commcare/formplayer/auth/HmacAuthFilter.java
@@ -5,6 +5,8 @@ import org.commcare.formplayer.objects.SerializableFormSession;
 import org.commcare.formplayer.services.FormSessionService;
 import org.commcare.formplayer.util.Constants;
 import org.commcare.formplayer.util.RequestUtils;
+import org.commcare.util.JsonUtils;
+import org.json.JSONArray;
 import org.json.JSONObject;
 import org.springframework.core.log.LogMessage;
 import org.springframework.http.HttpMethod;
@@ -106,14 +108,15 @@ public class HmacAuthFilter extends GenericFilterBean {
 
     private HqUserDetailsBean getUserDetails(HttpServletRequest request) {
         JSONObject body = RequestUtils.getPostData(request);
+
         if (body.has("username") && body.has("domain")) {
             return new HqUserDetailsBean(
                     body.getString("domain"),
                     new String[]{body.getString("domain")},
                     body.getString("username"),
                     false,
-                    new String[]{},
-                    new String[]{}
+                    JsonUtils.toArray(body.optJSONArray("enabled_toggles")),
+                    JsonUtils.toArray(body.optJSONArray("enabled_previews"))
             );
         } else if (body.has("session-id")) {
             String sessionId = body.getString("session-id");
@@ -123,8 +126,8 @@ public class HmacAuthFilter extends GenericFilterBean {
                     new String[]{formSession.getDomain()},
                     formSession.getUsername(),
                     false,
-                    new String[]{},
-                    new String[]{}
+                    JsonUtils.toArray(body.optJSONArray("enabled_toggles")),
+                    JsonUtils.toArray(body.optJSONArray("enabled_previews"))
             );
         }
         throw new BadCredentialsException("Unable to extract user credentials from the request");
--- a/src/main/java/org/commcare/formplayer/auth/HmacAuthFilter.java
+++ b/src/main/java/org/commcare/formplayer/auth/HmacAuthFilter.java
@@ -1,7 +1,5 @@
 package org.commcare.formplayer.auth;
 
-import lombok.Builder;
-import lombok.extern.apachecommons.CommonsLog;
 import org.commcare.formplayer.beans.auth.HqUserDetailsBean;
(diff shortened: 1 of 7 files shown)


src/main/java/org/commcare/formplayer/auth/HmacAuthFilter.java
```diff
@@ -108,19 +108,23 @@ private HqUserDetailsBean getUserDetails(HttpServletRequest request) {
         JSONObject body = RequestUtils.getPostData(request);
         if (body.has("username") && body.has("domain")) {
```
shubham1g5, line 109: @snopoke I am not sure what's the right course of action here. Should we be checking for feature flags in the response body and setting them here ? I am not sure what requests end up here and whethery they will have the `enabled_previews` and `enabled_toggles` specified.

snopoke: this path won't currently have any FF set since HQ isn't sending them through with these requests. I suppose it should be sending them so let's add a check and if they are present we can use them otherwise use an empty list.

src/test/java/org/commcare/formplayer/tests/HqUserDetailsTests.java
```diff
@@ -8,22 +9,39 @@
...
+        String[] enabledToggles = new String[]{"toggle_a", "toggle_b"};
+        String[] enabledPreviews = new String[]{"preview_a", "preview_b"};
+        HqUserDetailsBean user = new HqUserDetailsBean("domain", new String[]{"domain"}, "bilbo",
+                false, enabledToggles, enabledPreviews);
+        Assertions.assertTrue(FeatureFlagChecker.INSTANCE.isPreviewEnabled("preview_a", user));
+        Assertions.assertTrue(FeatureFlagChecker.INSTANCE.isPreviewEnabled("preview_b", user));
+        Assertions.assertFalse(FeatureFlagChecker.INSTANCE.isPreviewEnabled("preview_c", user));
+        Assertions.assertTrue(FeatureFlagChecker.INSTANCE.isToggleEnabled("toggle_a", user));
+        Assertions.assertTrue(FeatureFlagChecker.INSTANCE.isToggleEnabled("toggle_b", user));
+        Assertions.assertFalse(FeatureFlagChecker.INSTANCE.isToggleEnabled("toggle_c", user));
```
snopoke, line 45: what's `INSTANCE`? Why not make the methods on that class static?

src/main/java/org/commcare/formplayer/beans/auth/FeatureFlagChecker.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+object FeatureFlagChecker {
+
+    fun isPreviewEnabled(preview: String, userDetailsBean: HqUserDetailsBean): Boolean {
+        return isGranted(HqUserDetailsBean.PREVIEW_PREFIX + preview, userDetailsBean)
+    }
+
+    fun isToggleEnabled(preview: String, userDetailsBean: HqUserDetailsBean): Boolean {
+        return isGranted(HqUserDetailsBean.TOGGLE_PREFIX + preview, userDetailsBean)
+    }
```
snopoke, line 14:
```suggestion
    fun isToggleEnabled(name: String, userDetailsBean: HqUserDetailsBean): Boolean {
        return isGranted(HqUserDetailsBean.TOGGLE_PREFIX + name, userDetailsBean)
    }
```

src/main/java/org/commcare/formplayer/beans/auth/FeatureFlagChecker.kt
```diff
@@ -0,0 +1,24 @@
+package org.commcare.formplayer.beans.auth
+
+/**
+ * Utility class to check for enabled previews and toggles
+ */
+object FeatureFlagChecker {
+
+    fun isPreviewEnabled(preview: String, userDetailsBean: HqUserDetailsBean): Boolean {
```
snopoke, line 8: Instead of passing in the user details here we could get it from the security context. Or use this utility that we already have: https://github.com/dimagi/formplayer/blob/61b49325e7a131e7ef78d63ca92d3ca413a6bbf2/src/main/java/org/commcare/formplayer/util/RequestUtils.java#L99-L106

src/test/java/org/commcare/formplayer/tests/HqUserDetailsTests.java
shubham1g5: thanks for the prompt. It was an oversight from my part.

snopoke (approved):

snopoke:
@shubham1g5 this merge broke the build because it didn't include the submodule update

snopoke:
also the PR build was failing so it shouldn't really have been merged

shubham1g5:
@snopoke I did the submodule update directly on master after the merge. Build status is positive [now](https://jenkins.dimagi.com/job/formplayer) . I was refraining from updating the submodule itself in the PR since if it was to deploy on staging, it would have created unresolvable conflicts, but I guess I could follow this process in future to avoid these failures - 
1. Merge Core PR
2. Update Submodule in the Formplayer
3. Verify Build
4. Merge Formplayer PR

PR #973, opened 2021-05-28
Support for multi-select fields in case search

shubham1g5:
[Spec](https://docs.google.com/document/d/19enjhrAMx0bnpUyFpCR7TgdD4W-4BL21Vwc_elXafoc/edit)
https://dimagi-dev.atlassian.net/browse/USH-1028

cross-request: https://github.com/dimagi/commcare-core/pull/998

We are using `#,#` as a delimiter to parse the multi-select choices in between web apps and Formplayer. That means that if user selects choice A and B on Web Apps, they will get communicated to Formplayer as `A#,#B` and vice versa


src/main/java/org/commcare/formplayer/screens/FormplayerQueryScreen.java
```diff
@@ -15,19 +10,4 @@
     public FormplayerQueryScreen(){
         super(null, null, null);
     }
-
-    /**
-     *
-     * @param skipDefaultPromptValues don't apply the default value expressions for query prompts
-     * @return case search url with search prompt values
-     */
-    public URI getUri(boolean skipDefaultPromptValues) {
```
shubham1g5, line 24: Removed in favour of [QueryScreen.buildUrl](https://github.com/dimagi/commcare-core/blob/master/src/cli/java/org/commcare/util/screen/QueryScreen.java#L94)

src/test/java/org/commcare/formplayer/tests/CaseClaimTests.java
```diff
@@ -121,7 +121,8 @@ public void testQueryScreen() throws Exception {
                 QueryResponseBean.class);
         assert queryResponseBean.getDisplays()[0].getValue().contentEquals("Burt");
         assertArrayEquals(queryResponseBean.getDisplays()[1].getItemsetChoices(), new String[]{"karnataka", "Raj as than"});
-        assertArrayEquals(queryResponseBean.getDisplays()[2].getItemsetChoices(), new String[]{"Bangalore", "Hampi"});
+        assertArrayEquals(queryResponseBean.getDisplays()[2].getItemsetChoices(), new String[]{"Bangalore", "Baran", "Hampi", "Kota"});
```
snopoke, line 124: not quite following this test. How come there are more choices now?

snopoke: Would it be useful to add a line to test the value of `queryResponseBean.getDisplays()[2].value`?

shubham1g5: choices for this field (`district`) filters on the selection for the `state` field. Since with the multiselect, we have changed state selection to add one more state, the district options have increased as well. 

Added tests on `value` . Thanks for the prompt.

src/main/java/org/commcare/formplayer/screens/FormplayerQueryScreen.java
shubham1g5: Had to add it back since QueryScreen.buildURI in core encodes the url params which results in double encoding url params on Foromplayer since `restTemplate`  by default encodes the url in Spring.

shubham1g5:
Bumping this for a review as we anticipate QA to be done soon.

src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
```diff
@@ -43,9 +44,17 @@ public QueryResponseBean(QueryScreen queryScreen, SessionWrapper session) {
...
-            if (queryPromptItem.isSelectOne()) {
-                currentAnswer = choicesAndAnswerIndex.second == -1 ? null : String.valueOf(choicesAndAnswerIndex.second);
+            String[] choiceLabels = null;
+            if (queryPromptItem.isSelect()) {
+                String[] selectedChoices = RemoteQuerySessionManager.extractSelectChoices(currentAnswer);
+                String answerWithChoiceIndices = "";
+                for (int i = 0; i < selectedChoices.length; i++) {
+                    int choiceIndex = ItemSetUtils.getIndexOf(queryPromptItem.getItemsetBinding(), selectedChoices[i]);
+                    answerWithChoiceIndices += choiceIndex == -1 ? "" : String.valueOf(choiceIndex);
+                    answerWithChoiceIndices += i == selectedChoices.length -1 ? "" : RemoteQuerySessionManager.MULTI_SELECT_DELIMITER;
```
snopoke, line 54: might be cleaner to put the indices in an array or list and then use `String.join` with the delimiter to get the final result.

src/main/java/org/commcare/formplayer/screens/FormplayerQueryScreen.java
snopoke: Pity about the duplication - we could create a separate RestTemplate instance that doesn't do the encoding or possibly some other workaround.

src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
```diff
@@ -43,9 +44,17 @@ public QueryResponseBean(QueryScreen queryScreen, SessionWrapper session) {
...
+                }
+                currentAnswer = answerWithChoiceIndices;
```
snopoke, line 56: quite a lot of duplication with `QueryScreen.answerPrompts`, any way to reduce that?

src/test/java/org/commcare/formplayer/tests/CaseClaimTests.java
```diff
@@ -120,9 +123,23 @@ public void testQueryScreen() throws Exception {
...
         assert queryResponseBean.getDisplays()[0].getValue().contentEquals("Burt");
+        assert queryResponseBean.getDisplays()[1].getValue().contentEquals("0");
         assertArrayEquals(queryResponseBean.getDisplays()[1].getItemsetChoices(), new String[]{"karnataka", "Raj as than"});
+
+        // check if we have districts corresponding to karnataka state
         assertArrayEquals(queryResponseBean.getDisplays()[2].getItemsetChoices(), new String[]{"Bangalore", "Hampi"});
 
+
+        // multi-select test
+        inputs.put("district", "0#,#1"); // select 2 districts
```
snopoke, line 134: do you need to call `queryData.setInputs("search_command.m1", inputs);` again after modifying `inputs`?

shubham1g5: Not really, modifying `inputs` modify the memory reference for `queryData`. That's how we are seeing the two district values added to the search url at L185 below.

src/main/java/org/commcare/formplayer/screens/FormplayerQueryScreen.java
shubham1g5: yeah I am not sure if it's worth creating another RestTemplate instance just to avoid duplication here. It feels like we will be introducing complexity to avoid complexity. I am going to leave it as it is in context of this PR and can revisit if it bothers us again.

src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
shubham1g5: thanks, made the change

src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
shubham1g5: I looked at it, but both of these places seem to be doing fundamentally opposite things making it hard to DRY this code out. The variable declarations and loop is same in both of these places though the value conversion is opposite and I don't have good ideas on how to unify this code better.

src/test/java/org/commcare/formplayer/tests/CaseClaimTests.java
snopoke: :+1: just checking since it was called like that above

snopoke (approved):

PR #974, opened 2021-05-31
Return caption information for each choice

gherceg:
[JIRA Ticket](https://dimagi-dev.atlassian.net/browse/SAAS-10816)

As Kevin says in the ticket:
> we need a way to pass down the image strings from formplayer

This PR achieves this by creating a new field on the question response called `choices_captions` (open to other names). The indexes on this field match those of `choices`, and each entry includes image, audio, and video fields matching the same naming scheme used for a single question.


shubham1g5 (changes requested):
@gherceg the idea looks right to me, couple points of feedback first of which is optional -
1. Instead of maintaining 2 separate arrays it would be nicer if we can merge the choice attributes and values into a single JsonArray like - 

````
"choices_v2": [
      {
        "value" : "..",
        "caption_audio": "..",
        "caption_image": "..",
        "caption_video": ".."
      },
      {
        ...
      }
]
````
That way HQ code would not need to match the choice across two arrays. 

2. Can we add test coverage for checking the coversion of choices from form xml to `QuestionBean`  probably as part of `FormEntryTest`.

**Comment by gherceg** [comment]
Thanks for the feedback @shubham1g5 and sorry for the delayed response. We did talk about consolidating choices and captions into a single array because that is a better solution. The hesitation was refactoring the existing `choices` field to return this new consolidated array, whereas the solution you mention gives us a safe path forward for migrating to the new value. 

My only concern is making sure there is a plan in place to migrate in the near future to avoid two versions of choices hanging around for long. I'll make that change and document the intention. I'll also happily add a test coverage. Thanks for enforcing that!

**Comment by gherceg** [comment]
@shubham1g5 @ShivamPokhriyal finally added some tests. I could add some more tests around QuestionBean properties, but before doing so just want feedback on what is there now. Let me know your thoughts. Thanks.

File: src/main/java/org/commcare/formplayer/api/json/PromptToJson.java [hunk 1]
```diff
@@ -246,4 +249,34 @@ private static String parseDataType(FormEntryPrompt prompt) {
...
+        if (prompt.getControlType() == Constants.CONTROL_SELECT_MULTI && choice.getValue().contains(" ")) {
+            throw new ApplicationConfigException(String.format("Select answer options cannot contain spaces. " +
+                    "Question %s with answer %s", prompt, choiceValue));
+        }
+        String imagePath = prompt.getSpecialFormSelectChoiceText(choice, FormEntryCaption.TEXT_FORM_IMAGE);
+        String audioPath = prompt.getSpecialFormSelectChoiceText(choice, FormEntryCaption.TEXT_FORM_AUDIO);
+        String videoPath = prompt.getSpecialFormSelectChoiceText(choice, FormEntryCaption.TEXT_FORM_VIDEO);
+
+        JSONObject choicesAndCaptions = new JSONObject();
+        choicesAndCaptions.put("value", jsonNullIfNull(choiceValue));
```
shubham1g5, line 275: `value` is non-nullable, so you don't need to wrap it inside `jsonNullIfNull`

src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
```diff
@@ -44,6 +44,9 @@ public static void parseQuestion(FormEntryPrompt prompt, JSONObject questionJson
 
         if (prompt.getDataType() == Constants.DATATYPE_CHOICE || prompt.getDataType() == Constants.DATATYPE_CHOICE_LIST) {
             questionJson.put("choices", parseSelect(prompt));
+            // this creates an improved choices object that contains both values and captions for each choice
+            JSONArray choicesWithCaptions = parseChoicesWithCaptions(prompt);
+            questionJson.put("choices_v2", choicesWithCaptions);
```
shubham1g5, line 49: Is there a HQ PR that handles this attribute, If yes, would you be able to link it the PR description.

src/test/java/org/commcare/formplayer/tests/FormEntryTest.java
```diff
@@ -229,4 +230,25 @@ public void testSubmitAnswerValidation_prevalidate() throws Exception {
...
+    public void testQuestionBeanChoices() throws Exception {
+        NewFormResponse newFormResponse = sessionNavigate(new String[]{"0", "0"}, "multiple_choice_test", NewFormResponse.class);
+        QuestionBean[] questions = newFormResponse.getTree();
+        String[] choices = questions[0].getChoices();
+        assertEquals("With attachment", choices[0]);
+        assertEquals("Without attachment", choices[1]);
+    }
+
+    @Test
+    public void testQuestionBeanChoicesV2WithMultimedia() throws Exception {
```
shubham1g5, line 244: thanks for adding these!

shubham1g5 (approved):
left very minor feedback, this looks great to me and is good to go!

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -181,10 +178,11 @@ public void setUp() throws Exception {
         restoreFactoryMock.getSQLiteDB().closeConnection();
         PrototypeUtils.setupThreadLocalPrototypes();
         LocalizerManager.setUseThreadLocalStrategy(true);
+        LocalizerManager.getGlobalLocalizer().addAvailableLocale("en");
+        Localization.setLocale("en");
```
gherceg, line 182: @snopoke I needed to set the current locale to fix the `LocaleTextException` I was facing. When running the full test suite, it gets registered/set [here](registerLanguageReference), but when running individual tests it is null. I didn't use this `register` method because I didn't have a valid uri to pass in. Does this seem like the best approach?

src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
gherceg: Yeah it's currently on a branch locally. Figured I wouldn't PR it until this was set in stone so I'll make that today and link it.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
snopoke: @shubham1g5 is probably the best person to answer that.

shubham1g5: Looks harmless to me. Though if it's only specific to this test probably good to localize this inside the `FormEntryTest` itself. If we need to have it in `BaseTestClass`,  I think setting the locale to `default` might be slightly generalized than `en`.   I didn't get how it's getting set when running the full test suite, the link you mentioned seems to be broken. Is it that the locale set as part of a different test is getting retained in this test ?

gherceg: > Is it that the locale set as part of a different test is getting retained in this test ?
Yes
Sorry about that. Meant to link [here](https://github.com/dimagi/formplayer/blob/4a6317119f716b5490772622fe0e1bee1dc8f272/src/main/java/org/commcare/formplayer/engine/FormplayerConfigEngine.java#L179). I am fairly confident that this issue extends to more than just the `FormEntryTest`, though don't have any data to back that up. Simon mentioned he ran into a similar error when running one commcare-core test, but otherwise hasn't had an issue. With that in mind, I'll move into the `FormEntryTest` for now, but if we run into a `LocaleTextException` in another test, this is likely the solution. What is the best way to document that?

PR #977, opened 2021-06-23
csrf protection

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-11750

Using `CookieCsrfTokenRepository` to manage the csrf token generation. It persists the CSRF token in a cookie named "XSRF-TOKEN" and then reads from the header "X-XSRF-TOKEN"

Corresponding HQ PR: https://github.com/dimagi/commcare-hq/pull/29954

In case of 403s this PR should be reverted first and then the HQ PR can be reverted. 

## **Deploy instructions**:: 
First the HQ PR should be deployed https://github.com/dimagi/commcare-hq/pull/29954. 
Once that is successful, we can deploy this PR.

### DIFF
--- a/src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
+++ b/src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
@@ -18,6 +18,7 @@ import org.springframework.security.config.annotation.web.configuration.WebSecur
 import org.springframework.security.config.http.SessionCreationPolicy;
 import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationProvider;
+import org.springframework.security.web.csrf.CookieCsrfTokenRepository;
 import org.springframework.security.web.firewall.HttpFirewall;
 import org.springframework.security.web.firewall.StrictHttpFirewall;
 
@@ -49,6 +50,7 @@ public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
         http.authorizeRequests().antMatchers("/**").authenticated();
         http.addFilterAt(getHmacAuthFilter(), UsernamePasswordAuthenticationFilter.class);
         http.addFilterAfter(sessionAuthFilter(), HmacAuthFilter.class);
+        http.csrf().csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse());
         http.cors();
     }
 
@@ -60,7 +62,6 @@ public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
     }
 
     private void disableDefaults(HttpSecurity http) throws Exception {
-        http.csrf().disable();
         http.sessionManagement().sessionCreationPolicy(SessionCreationPolicy.STATELESS);
         http.requestCache().disable();  // only needed for login workflow
         http.logout().disable();
--- a/src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
+++ b/src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
@@ -4,6 +4,7 @@ import org.commcare.formplayer.auth.CommCareSessionAuthFilter;
 import org.commcare.formplayer.auth.HmacAuthFilter;
 import org.commcare.formplayer.services.FormSessionService;
 import org.commcare.formplayer.services.HqUserDetailsService;
+import org.commcare.formplayer.util.Constants;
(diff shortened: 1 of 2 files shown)


ShivamPokhriyal:
Created a draft PR to get early feedback. 
The corresponding HQ PR: https://github.com/dimagi/commcare-hq/pull/29954

snopoke (approved):

ctsims:
Nice! This is great to see.

orangejenny:
@ShivamPokhriyal Can you be explicit in the PR description that if this is rolled back, the HQ PR must first be rolled back (I think that's true)?

ShivamPokhriyal:
> @ShivamPokhriyal Can you be explicit in the PR description that if this is rolled back, the HQ PR must first be rolled back (I think that's true)?

I think if we need to rollback, formplayer should be rolled back first, and we can even leave the HQ as it is. 

HQ is only sending the `X-XSRF-TOKEN` after reading it from a cookie, and if it doesn't find anything, I believe it'll send an empty string in that header. So if we rollback this, then formplayer will simply ignore that header and everything will continue to work.

src/main/java/org/commcare/formplayer/configuration/WebSecurityConfig.java
```diff
@@ -49,6 +50,7 @@ protected void configure(HttpSecurity http) throws Exception {
         http.authorizeRequests().antMatchers("/**").authenticated();
         http.addFilterAt(getHmacAuthFilter(), UsernamePasswordAuthenticationFilter.class);
         http.addFilterAfter(sessionAuthFilter(), HmacAuthFilter.class);
+        http.csrf().csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse());
```
snopoke, line 53: I think you should update the filter to ignore requests that have the HMAC header (this will check if the header is present and the value is not `null`). These requests don't come from a browser, they come from HQ directly so CSRF isn't necessary.

```suggestion
        http.csrf()
            .csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse())
            .ignoringRequestMatchers(new RequestHeaderRequestMatcher(Constants.HMAC_HEADER));
```

We should also test to make sure that if the HMAC auth fails we don't fall back to the session auth otherwise that would allow you to provide any value for the HMAC header and bypass csrf.

It would be great if we could get an integration test for this. Some useful docs:
* https://www.baeldung.com/spring-security-integration-tests
* https://www.baeldung.com/spring-security-csrf

src/test/java/org/commcare/formplayer/tests/CsrfIntegrationTest.java
```diff
@@ -0,0 +1,142 @@
...
+import static org.springframework.test.web.client.match.MockRestRequestMatchers.jsonPath;
+import static org.springframework.test.web.client.match.MockRestRequestMatchers.requestTo;
+import static org.springframework.test.web.client.response.MockRestResponseCreators.withSuccess;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+/**
+ * @author $|-|!˅@M
+ */
```
snopoke, line 41: nit: comment not useful

src/test/java/org/commcare/formplayer/tests/CsrfIntegrationTest.java
```diff
@@ -0,0 +1,142 @@
...
+        headers.setContentType(MediaType.APPLICATION_JSON);
+        headers.add("Cookie", Constants.POSTGRES_DJANGO_SESSION_ID + "=" + "derp");
+        headers.add(Constants.HMAC_HEADER, "BHOwo3mPXbtWM91RO0g5HQOt+DtiiQVnCWMFsvjkWVc=");
+        HttpEntity<String> entity = new HttpEntity<>(payload, headers);
+
+        ResponseEntity<String> response = testRestTemplate.exchange(
+                "http://localhost:" + port + "/" + Constants.URL_DELETE_APPLICATION_DBS,
+                HttpMethod.POST, entity, String.class);
+
+        System.err.println(response.getStatusCode());
```
snopoke, line 140: nit: print

snopoke: should this be an assert?

snopoke (approved):

orangejenny:
Merging at @ShivamPokhriyal 's request

PR #982, opened 2021-06-28
Revert "Return caption information for each choice"

orangejenny:
Reverts dimagi/formplayer#974


calellowitz (approved):

shubham1g5:
Was there an issue with this ?

orangejenny:
@shubham1g5 Yeah, the additional data in formplayer responses contributed to performance issues. I wouldn't say this PR is itself a problem, but it was the safest/easiest thing to roll back to relieve some acute issues. There's more info in #web-team and https://dimagi-dev.atlassian.net/browse/SUPPORT-9926

PR #984, opened 2021-07-07
Adds endpoint to Launch a session endpoint

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1120

[Spec](https://docs.google.com/document/d/1w7oRuH9skVhNB1vuhIa8XWbSBoUWNj3f_N6lnU_6lnc/edit#heading=h.qr8tne77ash9)

cross-request: https://github.com/dimagi/commcare-core/pull/1013


snopoke (approved):
It would be nice to be able to see the relevant config in `endpoint.ccz`

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -584,4 +588,32 @@ private NotificationMessage establishVolatility(FormSession session) {
         }
         return null;
     }
+
+    public BaseResponseBean advanceSessionWithEndpoint(MenuSession menuSession, String endpointId, @Nullable HashMap<String, String> endpointArgs)
+            throws Exception {
+        Endpoint endpoint = menuSession.getEndpoint(endpointId);
+        if (endpoint == null) {
+            throw new RuntimeException("No endpoint provided to navigateToEndpoint");
```
ctsims, line 596: NOTE: This will be the presumed path when an app is changed such that an old endpoint no longer exists, right? We probably want to provide a more meaningful semantic error in that case to let people know why the link they followed can't work.

ctsims (changes requested):
Hey Shubham,

Looking like a good start, one small point about the error handling approach.

One other thing - I think the linked spec is a little too "speculative" to be the spec for this actual PR. It's hard for me to tell how this PR will actually fit into the solution structure proposed there, since it's so high level. Are the details about the overlap fleshed out elsewhere?

In particular I'd very much like to understand the plan for serializing the endpoint state into a URL and knowing how that's going to get translated between formplayer and HQ responsibilitites.

shubham1g5:
@ctsims I think this PR mostly tries to follow the MVP design from [here](https://docs.google.com/document/d/1w7oRuH9skVhNB1vuhIa8XWbSBoUWNj3f_N6lnU_6lnc/edit#heading=h.qr8tne77ash9). The idea here is just to expose an endpoint that HQ can utilize and to compliment the [HQ PR](https://github.com/dimagi/commcare-hq/pull/30002) here. 

> In particular I'd very much like to understand the plan for serializing the endpoint state into a URL and knowing how that's going to get translated between formplayer and HQ responsibilitites.

Probably @orangejenny would have better insights on how links are going to be generated though think the MVP right now skips this part of functionality.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5:
```suggestion
            throw new RuntimeException("This link does not exist. Your app may have changed so that the given link is no longer valid");
```
Would this be better ?

orangejenny: New wording sounds much better, I updated the HQ PR to match it.

orangejenny:
@shubham1g5 I added a commit to sync before going to the endpoint, largely because it's likely that HQ will need to claim any cases in the URL, and it's simpler to sync from formplayer than to do it as an additional request from web apps.

@ctsims Let's offline if the spec doesn't explain what you're interested in. The serialized URL contains the canonical app id and the endpoint id and passes the endpoint arguments as URL params.

src/main/java/org/commcare/formplayer/application/MenuController.java
```diff
@@ -162,27 +162,10 @@ public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBe
...
-        EvaluationContext evalContext = sessionWrapper.getEvaluationContext();
-        HashMap<String, String> endpointArgs = new HashMap<>();
-        for (Map.Entry<String, String> arg : sessionNavigationBean.getEndpointArgs().entrySet()) {
-            endpointArgs.put(arg.getKey(), arg.getValue());
-        }
-        Endpoint.populateEndpointArgumentsToEvaluationContext(endpoint, endpointArgs, evalContext);
-
-        sessionWrapper.executeStackOperations(endpoint.getStackOperations(), evalContext);
-        runnerService.rebuildMenuSession(menuSession);
-        sessionNavigationBean.setSelections(menuSession.getSelections());
```
orangejenny: Was it intentional to drop this line? The response now has `selections` as null, so web apps can't create breadcrumbs. Maybe `advanceSessionWithSelections` never sets selections because it assumes that they're already set properly?

shubham1g5: Ahh I removed it because it looked repetitive to me as the `advanceSessionWithSelections` should itself be enough of setting all necessary params. Though it looks like the `setSelections` call in it is under this [condition](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L286).My sense is that we should always be setting `selections` without this if condition specially if it's used in setting the breadcrumbs. Although there seems to be a separate `breadcrumbs` array in response which looks like the entity that should ideally be responsible `breadcrumbs` ? I am quite unsure what do we use `selections` for on Web Apps.

orangejenny: Ah, removing that conditional sounds good to me. Web Apps needs both breadcrumbs and selections: breadcrumbs has the text for display, and then selections has the indexes so we can navigate when someone clicks on one of the breadcrumbs.

shubham1g5: Thanks added https://github.com/dimagi/formplayer/pull/984/commits/00fb686bc3e99e33d53d03012d761adb4ed3fd3e

orangejenny:
Looks like https://github.com/dimagi/commcare-core/pull/1019/ broke this - it seems that endpoints including a case id are now getting stuck on the case list, rather than proceeding into the next menu. I'm not sure if that's also why tests are failing on this PR, but I'll look into it on Monday.

orangejenny:
Fixed the issue caused by https://github.com/dimagi/commcare-core/pull/1019 in 8cab61b126ae2a1eddfedf9ab17594dee8e0f1fa

orangejenny:
@ctsims This has evolved quite a bit since you requested changes back in July. Could you see if your concerns are addressed now?

orangejenny:
@ctsims Pinging for review again - this PR is otherwise ready to go.

orangejenny:
Merging, we'll do any followup in future PRs.

PR #994, opened 2021-08-18
Sends question_id to webapps

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1070

We want to show the `question_id` in the web apps errored questions text whenever question text is missing. 

HQ PR - https://github.com/dimagi/commcare-hq/pull/30259

### DIFF
--- a/src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
+++ b/src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
@@ -35,6 +35,7 @@ public class PromptToJson {
         parseCaption(prompt, questionJson);
         questionJson.put("help", jsonNullIfNull(prompt.getHelpText()));
         questionJson.put("binding", jsonNullIfNull(prompt.getQuestion().getBind().getReference().toString()));
+        questionJson.put("question_id", jsonNullIfNull(prompt.getQuestion().getBind().getReference().getNameLast()));
         questionJson.put("datatype", jsonNullIfNull(parseDataType(prompt)));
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/formplayer/api/json/PromptToJson.java
```diff
@@ -35,6 +35,7 @@ public static void parseQuestion(FormEntryPrompt prompt, JSONObject questionJson
         parseCaption(prompt, questionJson);
         questionJson.put("help", jsonNullIfNull(prompt.getHelpText()));
         questionJson.put("binding", jsonNullIfNull(prompt.getQuestion().getBind().getReference().toString()));
+        questionJson.put("question_id", jsonNullIfNull(prompt.getQuestion().getBind().getReference().getNameLast()));
```
snopoke, line 38: could pull `prompt.getQuestion().getBind().getReference()` out into a variable to avoid repeating it.

Also, any reason not to return the `question_id` in `toString()` if there is no text?

shubham1g5: > Also, any reason not to return the question_id in toString() if there is no text?

I am not sure if we would want to always default to question_id whenever there is no text for example in the primary form entry view today we just show empty space whenever the question view is empty. This seems a valid UX to me as it accurately reflects the app buider design. 

Screenshot 2021-08-18 at 2 28 36 PM

snopoke (approved):

PR #1004, opened 2021-08-31
Updated session caching to apply to all sub-lists of selections.

orangejenny:
https://dimagi-dev.atlassian.net/browse/QA-3382

PR into https://github.com/dimagi/formplayer/pull/984

Depends on https://github.com/dimagi/commcare-core/pull/1036

https://github.com/dimagi/commcare-core/pull/1021 updates `EntityListSubscreen.handleInputAndUpdateHost` so that when a case is selected, the EntityScreen highlights the selected case and then returns false. That false is passed up to `CompoundScreenHost` (the EntityScreen) which then also returns false. That tells the CLI to redraw the EntityScreen - but since the selected case was highlighted, the redraw will cause display case details instead of the case list.

This is largely irrelevant to formplayer, which handles case details differently, except that 1021 also meant that the session was no longer being updated when a case was selected (see [CompoundScreenHost.handleInputAndUpdateSession](https://github.com/dimagi/commcare-core/pull/1036/files#diff-4eacb2ddf14c3ed6080ac1e6a556e269c7820578caee952fd1f56b38b993b20eR36)) which meant smart links wouldn't add their case id to the session and wouldn't be able to proceed beyond the case list.

In https://github.com/dimagi/formplayer/pull/984/commits/8cab61b126ae2a1eddfedf9ab17594dee8e0f1fa I dealt with this by caching the smart link's selections list. That cache is checked in [advanceSessionWithSelections](https://github.com/dimagi/formplayer/blob/0f1667ef4cf711b57496b4561b22c1d21a8a9557/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L230), and ultimately that [this block in MenuSession.handleInput](https://github.com/dimagi/formplayer/blob/0f1667ef4cf711b57496b4561b22c1d21a8a9557/src/main/java/org/commcare/formplayer/session/MenuSession.java#L145-L153) executes the `else` and just updates the session with the smart link's case id.

However, since only the full selections list was added to the cache, that only worked for smart links with a single case id that came last in the list. Parent-child selection links, which make a selections list like `[parent_menu_index, parent_case_id, child_menu_index, child_case_id]` were failing because when `advanceSessionWithSelections ` loops over the selections, it checks the cache for just the selections list up to the current one. On the iteration that handles the parent case index, it's looking for `[parent_menu_index, parent_case_id]` which wasn't in the cache.

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
@@ -5,7 +5,6 @@ import org.commcare.modern.session.SessionWrapper;
 import org.commcare.formplayer.services.RestoreFactory;
 import org.commcare.formplayer.util.SessionUtils;
 
-import java.util.Arrays;
 import java.util.HashMap;
 
 /**
@@ -100,15 +99,8 @@ public class BaseResponseBean extends LocationRelevantResponseBean {
         return selections;
     }
 
-    public void setSelections(String[] selections, RestoreFactory restoreFactory) {
+    public void setSelections(String[] selections) {
         this.selections = selections;
-
-        // When selections are set manually, the session is likely going to be played back automatically.
-        // Cache selections so that playing back the session won't get hung up on case details when it hits
-        // a case id, in MenuSession.handleInput. Any subset of selections that ends in a case id needs to be cached.
-        for (int i = 1; i <= selections.length; i++) {
-            restoreFactory.cacheSessionSelections(Arrays.copyOfRange(selections, 0, i));
-        }
     }
 
     public HashMap<String, String> getTranslations() {
--- a/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
@@ -2,7 +2,6 @@ package org.commcare.formplayer.beans.menus;
 
 import org.commcare.formplayer.beans.NotificationMessage;
 import org.commcare.modern.session.SessionWrapper;
-import org.commcare.formplayer.services.RestoreFactory;
 import org.commcare.formplayer.util.SessionUtils;
 
 import java.util.HashMap;
--- a/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
@@ -2,8 +2,10 @@ package org.commcare.formplayer.beans.menus;
 
 import org.commcare.formplayer.beans.NotificationMessage;
 import org.commcare.modern.session.SessionWrapper;
+import org.commcare.formplayer.services.RestoreFactory;
 import org.commcare.formplayer.util.SessionUtils;
 
+import java.util.Arrays;
 import java.util.HashMap;
 
 /**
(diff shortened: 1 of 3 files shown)


orangejenny:
Pinging for review

snopoke (approved):

shubham1g5 (commented):
It doesn't seem great behaviour to me that we depend on what's cached to determine the application workflow. For eg - this cache might get busted anytime (due to crossing size limit or so) which would result into unexpected issues for users while navigating a CommCare Session. 

Do we know that this dependency on the cached selections is introduced by https://github.com/dimagi/commcare-core/pull/1036 or has it been there from earlier ?

orangejenny:
@shubham1g5 Hmm, that is a good point. I don't think it's likely that the cache will have issues, since the selections get added to the cache right before they're used, but I agree that it's dodgy to use the cache to store data that's critical to the workflow. I'll take another look at this.

https://github.com/dimagi/commcare-core/pull/1021 introduced the issue, since it flipped the return values in `handleInputAndUpdateHost` to fix the case details workflow in the CLI.

orangejenny:
@shubham1g5 I pulled the caching out in https://github.com/dimagi/formplayer/pull/1004/commits/5e9e6251d9437baa6c831f1244e32bd16e9cb7ef and then updated the commcare-core PR in https://github.com/dimagi/commcare-core/pull/1036/commits/35a7d64ae19db014323006665d62746aa923aba7 so that the session is updated regardless of what `handleInputAndUpdateHost` returns. What do you think? It works for the two workflows described in the QA ticket (smart links and EOF nav). The only thing I'm worried about is the possibility of breaking some other workflow, so I will do some more testing and probably request QA to also do some navigation-based testing.

shubham1g5:
@orangejenny agree that does seem a bit risky as that might result in us advancing the session incorrectly.  Looking at the PR description on the [original PR](https://github.com/dimagi/commcare-core/pull/1019), I am wondering if following would fix the issue and don't change the Formplayer behaviour as well - 

- Not change the return values on `handleInputAndUpdateHost` and keep it as it is. This seems inline with the method doc as well - `True if the compound screen is ready and the session should move to the next step`

- Reverse the return values on `handleInputAndUpdateSession` as below. This does feel definitely wrong today as per the method description - `True if the session was updated and the app should proceed to the next phase`. 

```
    public final boolean handleInputAndUpdateSession(CommCareSession session, String input, boolean allowAutoLaunch) throws CommCareSessionException {
        if (getCurrentScreen().handleInputAndUpdateHost(input, this, allowAutoLaunch)) {
            this.updateSession(session);
            return true;
        }
        return false;
    }
```

As `handleInputAndUpdateSession` is not called by the Formplayer, so that should not change the FP behaviour. On the CLI side, preserve the change   `screenIsRedrawing = !s.handleInputAndUpdateSession(mSession, input, false);`  which should result into moving to the next screen everytime there is an update to the session. It's possible I am not thinking straight here as I have not played with CLI myself and unsure what's the exact issue there, though happy to get on a call to discuss this further.

orangejenny:
@shubham1g5 

> Not change the return values on handleInputAndUpdateHost and keep it as it is.

I tried this out today: the commcare-core PR now restores 1019 but adds [this fix commit](https://github.com/dimagi/commcare-core/pull/1036/commits/73cec178461aba87c7a472733453413e67be8037), which reverts the return values of `handleInputAndUpdateHost` to what they were pre-1019. Changing those values fixes formplayer but breaks the CLI, in the sense that the CLI then never displays the case details screen, so I also added an ugly little if statement to the main CLI loop.

> True if the session was updated and the app should proceed to the next phase

Ha, I have been pretty hung up on this comment and whether case list + case details is one "phase" or two. It's counterintuitive to me that case list selection updates the session, since the user hasn't yet verified the selection in case details...but case details are only required in the interactive workflow, not the automated navigation workflows (smart links and EOF nav). Anyway, what do you think of the current state of the code?

orangejenny:
@shubham1g5 Great. This PR just removes the last bit of the caching approach, so it's still worth merging. The open commcare-core PR should now no longer depend on the smart linking formplayer PR, which is good.

PR #1005, opened 2021-09-07
Serialize full xml when saving to a session unless submitting a form

gherceg:
[JIRA Ticket](https://dimagi-dev.atlassian.net/browse/SAAS-12385)

This is the first step in successfully moving to a shared `FormDefinition` table in postgres. Prior to this change, a `FormSession` was dependent on saving the form definition with state, when in reality the saved form definition should be stateless. Shoutout to @ctsims for identifying and making this change. I'll explain my understanding, but Clayton feel free to clarify further if anything I say is inaccurate.

Specifically, when initializing the `XFormSerializingVisitor` class, the `respectRelevance` option tells the serializer if it should ignore/exclude irrelevant questions from the serialized xml. We serialize the form xml prior to saving, and because this option defaults to true, we would exclude irrelevant nodes from the form. 

This manifested itself in a test where a form was created, a couple answers were performed, and finally the form was submitted. If the form definition were serialized and saved prior to the form def initialization, default values (e.g., `case_id`) that were triggered in the initialization would end up missing since those triggerables only fire once. By serializing the full xml, the reinitialization for the default value triggerables should fire properly when being read from an existing form session in a future request.

This change has a large potential blast radius from both a functional and performance perspective, and will require QA and load testing. 

[QA Ticket](https://dimagi-dev.atlassian.net/browse/QA-3617)


snopoke (commented):
Thanks for the detailed explanation.

src/test/java/org/commcare/formplayer/tests/FormDefInitTests.java
```diff
@@ -0,0 +1,40 @@
...
+        sandbox.getConnection().close();
+
+        String sessionId = newSessionResponse.getSessionId();
+
+        answerQuestionGetResult("0", "Tom Brady", sessionId);
+        answerQuestionGetResult("1", "1", sessionId);
+
+        SubmitResponseBean submitResponseBean = submitForm("requests/submit/submit_request_case.json", sessionId);
+
+        assertEquals("success", submitResponseBean.getStatus());
```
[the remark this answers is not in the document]
gherceg: Relevant for both of your comments, this isn't a very robust test. I grabbed it from [the test that initially failed](https://github.com/dimagi/formplayer/blob/a1e599955b89bea8a95cf16b98ee35c21a91323c/src/test/java/org/commcare/formplayer/tests/CaseTests.java#L32).

However this change is worthy of better automated testing so I will look into that.

src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -244,7 +244,7 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
                 String response;
                 try {
                     response = submitService.submitForm(
-                            formEntrySession.getInstanceXml(),
+                            formEntrySession.getInstanceXml(false),
```
gherceg, line 247: @ctsims I noticed this when writing tests. This is the minimum change necessary to make sure we respect relevance when serializing the instance xml for this particular form submission. Just want to confirm this makes sense with you.

ctsims: Correct, this should result in the xml not including nodes which are irrelevant.

dannyroberts:
@gherceg did this PR hit a blocker, or is did it just get lost in other work?

ctsims:
Hey Danny,

QA has been running through performance testing to make sure that this PR won't risk significant impacts on programs with certain uncommon (but important) patterns which might be disproportionally impacted, to figure out if we need to do anything to mitigate them.

dannyroberts:
Thanks @ctsims. @gherceg is it understood why the build currently fails?

gherceg:
@dannyroberts yeah the FormSession constructor was updated in a separate PR, and the tests weren't updated to reflect that. Should be good now.

gherceg:
@dannyroberts yeah tests weren't updated after merging in master, and it had been awhile so things were a bit busted. It is fixed now.

calellowitz (approved):

PR #1006, opened 2021-09-10
Adds Query Datum Support for a Form

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1231

Preserves the Session Frame Inside the `FormSession` to load up any query xml data instances into the Form Session

cross-request: https://github.com/dimagi/commcare-core/pull/1040

PR #1010, opened 2021-09-21
Smart linking in web apps

orangejenny:
https://dimagi-dev.atlassian.net/browse/USH-1292

Opening for review on the general approach.

Depends on https://github.com/dimagi/commcare-core/pull/1046 and should be released together with https://github.com/dimagi/commcare-hq/pull/30441, although using this feature requires making app changes, so the release doesn't need to be perfect / backwards compatible.

This implements smart linking for case search workflows that have a data registry configured. How it works:
1. Case search results now include a `commcare_project` attribute, thanks to https://github.com/dimagi/commcare-hq/pull/30388
1. In formplayer, the `get_details` response when case details is requested adds the `commcare_project` to the response. It's passed in the new `smartLinkParams` attribute on `EntityDetailListResponse`
1. Web apps sees the `smart_link_params`, compares the param to the user's current domain, and if they're different, sends back the params as part of the navigation request - this is the new `smartLinkParams` attribute in the `SessionNavigationBean`. If the domain in the formplayer response is the user's current domain, web apps creates a smart link template, based on the `session_endpoint` URL, the domain, and the app id. The template looks something like `https://www.commcare.org/a/DOMAIN/app/APP_ID/---/`, where the last `---` needs to be replaced by an endpoint id.
1. Formplayer gets the smart link template in the navigation request, determines what the next screen is, and looks up the corresponding endpoint for that screen. This lookup depends on https://github.com/dimagi/commcare-core/pull/1046 adding the command id to the definition of an endpoint.
1. Formplayer adds the endpoint's id and any arguments to the URL and passes it back to web apps, using the new `smartLinkRedirect` attr on `BaseResponseBean`.
   1. Adding in the arguments isn't currently working.
1. Web apps sees the `smartLinkRedirect` attr on the response and does a hard redirect to that URL.

Open issues (also making some comments in the code):
* ~~There probably needs to be some app manager configuration to turn this on. As is, smart links will be generated for any case search config that uses a registry.~~ [USH-1307](https://dimagi-dev.atlassian.net/browse/USH-1307)
* I haven't tested or really thought much about multiple-case workflows.

[QA Ticket](https://dimagi-dev.atlassian.net/browse/QA-3592)

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
@@ -7,6 +7,10 @@ import org.commcare.util.screen.EntityDetailSubscreen;
 import org.commcare.util.screen.EntityScreen;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.xpath.XPathParseTool;
+import org.javarosa.xpath.expr.XPathExpression;
+import org.javarosa.xpath.XPathNodeset;
+import org.javarosa.xpath.parser.XPathSyntaxException;
 
 import java.util.ArrayList;
 
@@ -17,6 +21,7 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
 
     private EntityDetailResponse[] entityDetailList;
     private boolean isPersistentDetail;
+    private String[] smartLinkParams;
 
     public EntityDetailListResponse() {}
 
@@ -52,6 +57,8 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
         }
 
         EvaluationContext subContext = new EvaluationContext(ec, ref);
+        setSmartLinkParams(subContext);
+
         ArrayList<Object> accumulator = new ArrayList<>();
         for (int i = 0; i < detailList.length; i++) {
             if (detailList[i].getNodeset() == null) {
@@ -76,6 +83,20 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
         return ret;
     }
 
+    private void setSmartLinkParams(EvaluationContext ec) {
+        try {
+            XPathExpression paramExpr = XPathParseTool.parseXPath("commcare_project");
+            String commcare_project = (String) ((XPathNodeset) paramExpr.eval(ec)).unpack();
+            if (!commcare_project.equals("")) {
+                String[] params = new String[1];
+                params[0] = commcare_project;
+                this.setSmartLinkParams(params);
+            }
+        } catch (XPathSyntaxException e) {
+            // nothing to do here
+        }
+    }
+
     @JsonGetter(value = "details")
     public EntityDetailResponse[] getEntityDetailList() {
         return entityDetailList;
@@ -94,5 +115,15 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
     @JsonSetter(value = "isPersistentDetail")
     public void setPersistentDetail(boolean persistentDetail) {
         this.isPersistentDetail = persistentDetail;
+    }
+
+    @JsonGetter(value = "smartLinkParams")
+    public String[] getSmartLinkParams() {
+        return smartLinkParams;
+    }
+
+    @JsonSetter(value = "smartLinkParams")
+    public void setSmartLinkParams(String[] params) {
+        this.smartLinkParams = params;
     }
 }
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
@@ -13,6 +13,7 @@ import org.javarosa.xpath.XPathNodeset;
 import org.javarosa.xpath.parser.XPathSyntaxException;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 
 /**
  * Created by willpride on 1/4/17.
@@ -21,7 +22,7 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
 
     private EntityDetailResponse[] entityDetailList;
     private boolean isPersistentDetail;
-    private String[] smartLinkParams;
+    private HashMap<String, String> smartLinkParams;
 
     public EntityDetailListResponse() {}
 
@@ -88,8 +89,8 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
             XPathExpression paramExpr = XPathParseTool.parseXPath("commcare_project");
             String commcare_project = (String) ((XPathNodeset) paramExpr.eval(ec)).unpack();
             if (!commcare_project.equals("")) {
-                String[] params = new String[1];
-                params[0] = commcare_project;
+                HashMap<String, String> params = new HashMap<>();
+                params.put("domain", commcare_project);
                 this.setSmartLinkParams(params);
             }
         } catch (XPathSyntaxException e) {
@@ -118,12 +119,12 @@ public class EntityDetailListResponse extends LocationRelevantResponseBean {
     }
 
     @JsonGetter(value = "smartLinkParams")
-    public String[] getSmartLinkParams() {
+    public HashMap<String, String> getSmartLinkParams() {
         return smartLinkParams;
     }
 
     @JsonSetter(value = "smartLinkParams")
-    public void setSmartLinkParams(String[] params) {
+    public void setSmartLinkParams(HashMap<String, String> params) {
         this.smartLinkParams = params;
     }
 }
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -200,7 +200,8 @@ public class MenuSessionRunnerService {
         if (command != null) {
             Endpoint endpoint = menuSession.getEndpointByCommand(command);
             if (endpoint != null) {
-                template = template.replace("---", endpoint.getId());
+                Hashtable<String, String> urlArgs = new Hashtable<String, String>();
+                urlArgs.put("endpoint", endpoint.getId());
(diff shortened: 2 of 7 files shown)


src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
```diff
@@ -76,6 +83,20 @@ public EntityDetailListResponse(Detail[] detailList, EvaluationContext ec, TreeR
         return ret;
     }
 
+    private void setSmartLinkParams(EvaluationContext ec) {
+        try {
+            XPathExpression paramExpr = XPathParseTool.parseXPath("commcare_project");
```
orangejenny, line 88: This is the one place in the formplayer code that references a project/domain. For the sake of being more generic, I'm thinking about pulling this into another app profile property, like a `cc-smart-link-param` property that would accept an XPath expression.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -181,6 +190,40 @@ private BaseResponseBean getNextMenu(MenuSession menuSession,
...
+        if (template != null) {
+            SessionWrapper session = menuSession.getSessionWrapper();
+            String command = session.getCommand();
+            if (command != null) {
+                Endpoint endpoint = menuSession.getEndpointByCommand(command);
+                if (endpoint != null) {
+                    for (String param : smartLinkParams) {
+                        template = template.replaceFirst("---", param);
+                    }
+                    template = template.replace("---", endpoint.getId());
```
orangejenny, line 208: It's a bit weird that an arbitrary number of parameters come from web apps and then one last parameter is set here. I'm debating making each of the param list attributes into just a single param. That would also make it a bit easier to turn that one web-apps-provided param into an app profile property, as described in previous comment.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -181,6 +190,40 @@ private BaseResponseBean getNextMenu(MenuSession menuSession,
...
+                    HttpUrl.Builder urlBuilder = HttpUrl.parse(template).newBuilder();
+                    OrderedHashtable<String, String> data = session.getData();
+                    for (String key : data.keySet()) {
+                        if (endpoint.getArguments().contains(key)) {
+                            urlBuilder.addQueryParameter(key, data.get(key));
+                        }
+                    }
+                    String finalUrl = urlBuilder.build().toString();
+                    BaseResponseBean responseBean = new BaseResponseBean(null, null, true);
+                    System.out.println("final url => " + finalUrl);
```
orangejenny, line 218:
```suggestion
```

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -181,6 +190,40 @@ private BaseResponseBean getNextMenu(MenuSession menuSession,
...
+            String command = session.getCommand();
+            if (command != null) {
+                Endpoint endpoint = menuSession.getEndpointByCommand(command);
+                if (endpoint != null) {
+                    for (String param : smartLinkParams) {
+                        template = template.replaceFirst("---", param);
+                    }
+                    template = template.replace("---", endpoint.getId());
+                    HttpUrl.Builder urlBuilder = HttpUrl.parse(template).newBuilder();
+                    OrderedHashtable<String, String> data = session.getData();
```
orangejenny, line 210: This doesn't work. The case id gets pushed onto the stack and then added to the session, but on the navigation request, [this block](https://github.com/dimagi/formplayer/blob/aa5ebb994318d8a920b30d4f0cb6f026a22287db/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L259-L262) executes, and ultimately it fails [this check](https://github.com/dimagi/formplayer/blob/aa5ebb994318d8a920b30d4f0cb6f026a22287db/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L74-L76) because the case isn't in the user's casedb.

I've been banging my head against the wall with this for a while and am thinking about just passing along any case-id-like selections as endpoint arguments, instead of pulling from the session.

shubham1g5:
Couple comments on the overall design without digging into specifics - 

1. I am curious how this would interact with https://github.com/dimagi/formplayer/pull/1006 where we want to navigate to the form in the current domain instead of getting smart linked to the the domain the case belongs to ? it does seem like there is a piece of config missing that allows user to choose from either of these configs. Also for my understanding confirming that the smart link redirection happens when clicking a particular module after the case selection from the case search results  ? If so, this might need a module/form level config property rather than the case list. 

2. I am wondering if it's possible to delegate the smart link formation from the template to HQ instead of Formplayer. It seems like all HQ needs to know is the endpoint id for the particular form from FP which we can pass on in fp's response to HQ. All other info like smart link template and case id selection along with the `commcare_project` the case belongs to should already be available to HQ.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5: Ahh I am running into this as well for https://github.com/dimagi/formplayer/pull/1006 as the navigation to a form break after the case selection. I am going to spend some time today and try to find a solution to this.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
snopoke: I think you'd want to use a URL template:

```java
UriComponentsBuilder urlBuilder = UriComponentsBuilder.fromHttpUrl(template)
 for (String key : data.keySet()) {
    if (endpoint.getArguments().contains(key)) {
        urlBuilder.queryParam(key, data.get(key));
    }
}
URI url = urlBuilder.build(urlArgs);
```

`urlArgs` could be a list / array or it could be a Map of named params. I think a Map is better but it would require making `smartLinkParams` into a map instead of just a list.

https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#web-uricomponents

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5: Talked to Clayton on this and it seems like we will need a special pathway for cross domain navigation that can move the user to the  query screen though an `action` without needing the usual `EntityScreen`.  @snopoke  Just flagging that this seems to be a bigger issue for menu navigation involving a case that user doesn't have in DB.  I am going to think more on how we can devise such an alternative route for cross domain case selection.

shubham1g5: @orangejenny Stumbled on an easier workaround for this which is to simply remove the `mark` and `rewind` constructs from the `stack` blocks in `action` and `remote-query` . That would mean that we always retain the intermediary query results into the session instead of cleaning them out with a `rewind`. I have not been able to think of any obvious regressions of not doing the `rewind` but do think we would need extensive case search QA to be entirely sure with this approach.

shubham1g5: Ahh apologies looks like this won't work out by itself as we somehow need to clear the `search_command` id from the stack for the session to move to the next step.

orangejenny: Rats. I did read your first comment thinking, "Surely there must be a reason we do the rewind..."

src/main/java/org/commcare/formplayer/beans/menus/EntityDetailListResponse.java
orangejenny: Leaving this alone for now.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
orangejenny: updated, both to use a template and turn params into a map

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5: @orangejenny  Simon did a restructuring of app files to address this issue in https://github.com/dimagi/commcare-hq/pull/30501 , would you be able to see if you still get blocked by this issue in the new app structure for data registry.

shubham1g5:
@orangejenny Thanks, looks great to me now. Wondering it makes sense to QA both this and https://github.com/dimagi/formplayer/pull/1006 together since it's controlled by the same config on HQ ?

orangejenny:
@shubham1g5 Yes, that sounds good. QA ticket: https://dimagi-dev.atlassian.net/browse/QA-3592

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
orangejenny: The smart links do now allow the user to jump to the other domain, instead of getting hung up on the navigation in the current domain. But, upon reaching the other domain, when the session is rebuilt [during endpoint navigation](https://github.com/dimagi/formplayer/blob/feac335727c77935c893325103ce19f705c65065/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L636), it gets to the point of realizing that all entries for the current datum contain a query, and so it sends the user to the case search screen. Basically, there's no mechanism in the query to say, "_Don't_ do this query if there's already a `case_id` in the session."

orangejenny: I'm thinking about adding a `relevant` attribute to the `query` block, so we can skip it if the case id is already provided.

shubham1g5: ahh it does seem broken. What if we make `default_search`

shubham1g5: yes, This structure definitely makes much more sense. Overall all the changes looks great to me :)

orangejenny:
Closing in favor of https://github.com/dimagi/formplayer/pull/1028

PR #1014, opened 2021-10-11
Smart linking in web apps: Stack queries

orangejenny:
See conversation in https://github.com/dimagi/formplayer/pull/1010#discussion_r723112122

This adds support for queries in stacks and adds a query to any session endpoint that references a registry. When the endpoint is executed and the session is rebuilt, the query is run to fetch the registry case.

Queries in the stack use their instance as their id and their URL as their value, similar to `RemoteQueryDatum`. They allow `<data>` elements as children to represent URL params, similar to queries in entries.

```
<endpoint id="endpoint_songs" command_id="m1">
    <argument id="case_id"/>
    <stack>
        <push>
            <datum id="case_id" value="$case_id"/>
            <command value="'claim_command.endpoint_songs.case_id'"/>
        </push>
        <push>
            <query id="results" value="http://localhost:8000/a/DOMAIN/phone/registry_case/APP_ID/">
                <data key="commcare_registry" ref="'songs'"/>
                <data key="case_type" ref="'song'"/>
                <data key="case_id" ref="$case_id"/>
            </query>
            <command value="'m1'"/>
            <datum id="case_id" value="$case_id"/>
        </push>
    </stack>
</endpoint>

```

This is branched off of both https://github.com/dimagi/formplayer/pull/1010 and https://github.com/dimagi/formplayer/pull/1006 so it can't be merged until at least one of those is.

Cross-requested with https://github.com/dimagi/commcare-core/pull/1048 and https://github.com/dimagi/commcare-hq/pull/30562


src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -77,6 +91,37 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
...
+                            queryScreen.setQueryDatum(searchDataInstance);
+                            screen = menuSession.getNextScreen(false);
+                            done = true;
+                        } catch (InvalidStructureException | IOException | XmlPullParserException | UnfullfilledRequirementsException e) {
+                            e.printStackTrace();
+                            throw new CommCareSessionException("Query response format error: " + e.getMessage());
+                        }
+                    }
+                }
+                if (done) {
```
shubham1g5, line 122: curious if this makes a difference since it's anyway at the end of the loop ?

shubham1g5 (approved):

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
snopoke: maybe this is meant to be inside the for loop above in which case it could be replaced with a `break`

ctsims: ++ to Simon's suggestion

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -77,6 +91,37 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
...
+            } else if (screen instanceof QueryScreen) {
+                QueryScreen queryScreen = (QueryScreen)screen;
+                RemoteQueryDatum neededDatum = (RemoteQueryDatum) menuSession.getSessionWrapper().getNeededDatum();
+                boolean done = false;
+                for (StackFrameStep step : steps) {
+                    if (step.getId().equals(neededDatum.getDataId())) {
+                        URI uri = null;
+                        try {
+                            uri = new URI(step.getValue());
+                        } catch (URISyntaxException e) {
```
ctsims, line 103: Echoing @shubham1g5's feedback in commcare-core that it would probably be good to push this validation to parse time (we can still and wrap here, but will push any **expected** failures to a point where they'd be caught in a cleaner state.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -77,6 +91,37 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
...
+                                queryScreen.getQueryDatum().getDataId(),
+                                queryScreen.getQueryDatum().useCaseTemplate(),
+                                uri
+                            );
+                            queryScreen.setQueryDatum(searchDataInstance);
+                            screen = menuSession.getNextScreen(false);
+                            done = true;
+                        } catch (InvalidStructureException | IOException | XmlPullParserException | UnfullfilledRequirementsException e) {
+                            e.printStackTrace();
+                            throw new CommCareSessionException("Query response format error: " + e.getMessage());
```
ctsims, line 118: Apologies if this pattern is in use elsewhere and you were just copying, but it would be best for this exception to be initialized with `e` as a source so that any reported errors actually carry the full stack trace.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -77,6 +91,37 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
...
+                    if (step.getId().equals(neededDatum.getDataId())) {
+                        URI uri = null;
+                        try {
+                            uri = new URI(step.getValue());
+                        } catch (URISyntaxException e) {
+                            e.printStackTrace();
+                            throw new CommCareSessionException("Query URL format error: " + e.getMessage());
+                        }
+                        try {
+                            ExternalDataInstance searchDataInstance = caseSearchHelper.getRemoteDataInstance(
```
ctsims, line 108: Is this performing the I/O synchronously inside of this call?

I think that burying the error handling here as such will cause issues long term. Previously this method's failures were all within its control, while now they aren't deterministic and could depend on, say, rate limiting from HQ.

This is kind of true in the UX version of these queries as well, but the difference is that there the I/O is in clear response to a user action, and is clearly idempotent. If it fails, it's in response to a user button press, and pressing the button again will produce the expected result.

In this case, it's not obvious to me how an I/O failure gets routed for a user, but I'm guessing it's "awkwardly." End-of-form nav isn't an obviously repeatable action, but the way the failure is routed should make it clear and possible for this to fail in an "Expected" way.

I think at the very least, the error handling should split here between structural failure (IE: there was an unrecoverable problem where XML payloads were un-parseable, repeating the action would be meaningless) and I/O failure (HQ responded with a 503) so that it's possible to differentiate both at a human level, and to let Formplayer try and do something smarter or more specific in response.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -77,6 +91,37 @@ public void rebuildSessionFromFrame(MenuSession menuSession) throws CommCareSess
...
+                RemoteQueryDatum neededDatum = (RemoteQueryDatum) menuSession.getSessionWrapper().getNeededDatum();
+                boolean done = false;
+                for (StackFrameStep step : steps) {
+                    if (step.getId().equals(neededDatum.getDataId())) {
+                        URI uri = null;
+                        try {
+                            uri = new URI(step.getValue());
+                        } catch (URISyntaxException e) {
+                            e.printStackTrace();
+                            throw new CommCareSessionException("Query URL format error: " + e.getMessage());
```
ctsims, line 105: Out of curiosity, are `CommCareSessionException` bubbled straight up to users as messages?

If so, I think these messages should probably be framed in more user facing language.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
snopoke: https://github.com/dimagi/commcare-core/pull/1048/commits/d1ea42af5e893cb33a097583fd40c21b89027ade

snopoke (commented):
Note: needs tests

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -549,11 +551,12 @@ public BaseResponseBean resolveFormGetNext(MenuSession menuSession) throws Excep
     }
 
     // Rebuild the session after executing any pending session stack
-    private boolean executeAndRebuildSession(MenuSession menuSession) throws CommCareSessionException {
+    private boolean executeAndRebuildSession(MenuSession menuSession) throws CommCareSessionException, RemoteInstanceFetcher.RemoteInstanceException {
         menuSession.getSessionWrapper().syncState();
+        menuSession.getSessionWrapper().prepareExternalSources(caseSearchHelper);
```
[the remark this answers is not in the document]
ctsims: Hey @snopoke - Might be worth chatting through the pattern a bit and iterating. 

In theory we could be loading the helper into the wrapper objects upfront and let the wrapper do its source prep on demand as needed in the Formplayer implementation, but my experience has been that that type of deferred / on-demand initialization can cause its own issues. 

I think the code should provide a clear signal in contexts when the source needed to be re-loaded but wasn't. I'm wondering whether the right approach might be to ensure that there are tests for all of the entry points which require re-establishing an instance, having them fail, then determining where the best place to perform initialization is. It feels like that's a good set of entry points to know somewhat more broadly.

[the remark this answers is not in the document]
snopoke: > It feels like that's a good set of entry points to know somewhat more broadly.

100% - this is exactly the context that I don't really have which I think would be very useful to know / have defined. Basically a clearer definition of the stack / session lifecycle.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
orangejenny: Simon fixed in https://github.com/dimagi/formplayer/pull/1014/commits/275f5a095680ebeb632db30fc6ce3b8e6fbb72f2

orangejenny:
Closing in favor of https://github.com/dimagi/formplayer/pull/1028

PR #1020, opened 2021-10-20
keep form session until all processing is done

snopoke:


ctsims:
@snopoke are you looking for a review on this one yet? I read the code changes and I think I understand the general structure and underlying change, but would probably want a high level description of any intended effects to know what's just code restructuring and what's intended functionality change.

PR #1026, opened 2021-11-03
move prepare external sources into the session wrapper

snopoke:
PR into https://github.com/dimagi/commcare-core/pull/1040

This moves the initialisation of external data sources to the session wrapper to ensure that they are always initialised when required. The potential downside here is that it get's called more frequently than required.

First two commits are a minor pre-factor.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -34,6 +34,9 @@ public class MenuSessionFactory {
     private InstallService installService;
 
     @Autowired
+    private CaseSearchHelper caseSearchHelper;
+
+    @Autowired
     protected FormplayerStorageFactory storageFactory;
 
     @Value("${commcarehq.host}")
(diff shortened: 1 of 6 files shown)


shubham1g5 (approved):

ctsims:
Is there a specific reason that the initialization is happening in the `getIIF()` route, rather than in the constructor?

I'm not 100% sure, but my memory is that getIIF is another call that can end up kind of buried in the core commcare code, rather than being upfront in the platform layer, which is what I presume is spurring the `@sneakythrows` wrapper.

snopoke:
> Is there a specific reason that the initialization is happening in the `getIIF()` route, rather than in the constructor?
> 
> I'm not 100% sure, but my memory is that getIIF is another call that can end up kind of buried in the core commcare code, rather than being upfront in the platform layer, which is what I presume is spurring the `@sneakythrows` wrapper.

It seemed to make sense at the time (and I don't like doing a lot of 'work' in constructors) but the fact that it can happen deep in the call stack is a good reason not to do it there.

PR #1028, opened 2021-11-04
Query Datums and Smart Linking in Web Apps

orangejenny:
This combines https://github.com/dimagi/formplayer/pull/1006, https://github.com/dimagi/formplayer/pull/1010, and https://github.com/dimagi/formplayer/pull/1014, which will be merged together and have conflicts.

QA tickets:
* https://dimagi-dev.atlassian.net/browse/QA-3592
* https://dimagi-dev.atlassian.net/browse/QA-3554


orangejenny:
@snopoke Can you check out the test failure?
```
EndOfFormNavFormLinkingWithQueryTests > testFormLinkingFromFollowupForm1ToFollowupForm2() FAILED
    org.springframework.web.util.NestedServletException: Request processing failed; nested exception is org.commcare.util.screen.CommCareSessionException: Query response format error: No re
sponse from server for case search query
```

snopoke:
@orangejenny I don't know how this worked before, perhaps I had HQ running locally and it was actually hitting HQ? Either way it revealed a bug so I've pushed the following changes:

66079903

https://github.com/dimagi/commcare-core/pull/1048/commits/3e077651bc5038a6b08d8456fd699a401ef2bbbb

I've also updated your merge branch on commcare-core and the submodule ref here

orangejenny:
@snopoke thank you!

orangejenny:
This is ready to go, anybody want to approve it?

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -290,34 +315,64 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
...
+     * @param forceManualAction Boolean passed in from the request which will prevent auto launch actions
+     * @return The next screen in the navigation cycle. This will mostly be the same value as is passed in except
+     *         where there are multiple chained queries that can be executed without user input.
+     * @throws CommCareSessionException if the was an error performing a query
+     */
+    private Screen handleQueryScreen(Screen nextScreen, MenuSession menuSession, QueryData queryData,
                                                   boolean replay, boolean forceManualAction)
             throws CommCareSessionException {
-        if (nextScreen instanceof FormplayerQueryScreen) {
+        while (nextScreen instanceof FormplayerQueryScreen) {
```
ctsims, line 343: The unbounded while here worries me a bit, since if there was a bug there's currently no way to get interrupted out of the thread, and thread locking has been an issue in the past.

Two potentially lightweight ideas: 1) checking for a thread interrupt and throwing a runtime exception would at least allow bailing out of the thread. 2) recursing instead of looping has the not quite intended (but sometimes convenient) property of throwing stack overflow exceptions and dying eventually, so it's sometimes slightly easier to write complex "We SHOULD always exit from this" walks recursively with a safe alternative if later changes break the exist logic.

snopoke: This has been refactored quite a lot here: https://github.com/dimagi/formplayer/pull/1033 (though it still has a while loop https://github.com/dimagi/formplayer/pull/1033/files#diff-585d2f4a6ce12f6bfb5a64f41142b32f6d3f0f10ee0b4303aa427cc7d6813d03R338). I could change to use recursion.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -27,12 +35,17 @@
 @Component
 public class MenuSessionFactory {
 
+    private static final String NEXT_SCREEN = "NEXT_SCREEN";
```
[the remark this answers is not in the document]
snopoke: Do you mean the whole rebuild function or the addition of the queries? I guess I don't really understand why the rebuild function is needed at all.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
ctsims: The refactor does make behavior much more clear, but it definitely still seems like a risky loop to me since what drives loop termination are return codes from a mix of different functions that could fall out of sync with each other in their meaning. I think it would be good to have some safety hook here that prevents churning indefinitely, like a safeLoop() function which checks for thread interruption and runtime exceptions if detected.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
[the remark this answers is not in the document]
ctsims: It seems like 'NEXT_SCREEN" was introduced as a formal type of stack frame step (IE: It sets the command code to a new concept, but only in formplayer, so I was hoping to understand the semantics a bit better.

"Rebuild" is just intended to be a way to encode the "meaning' of the current frame steps into a central place, rather than making it based on event-driven changes. IE: If the "current command" depends on interpreting all of the stack frames together, setting the outcome via a rebuild lets the engine put all the logic for how to make the decision in one place.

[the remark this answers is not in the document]
snopoke: This is an artefact of the way query screens are processed in the `MenuSession`. In normal navigation we don't treat query screens as navigable steps, instead we treat them as intermediate steps and don't record any state about them in the user selections (they query data is stored separately).

When adding queries to the stack this means we have to also take a different approach than what we do for other screens.

There are other side effects of this. Currently we could not have multiple query screens that need user input at the same command level (since we associate the query data with the current command).

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
snopoke: https://github.com/dimagi/formplayer/pull/1033/commits/dd1e6a277574c828d7f5eb009c97bb171bef2f0b

(or alternately: https://github.com/dimagi/formplayer/compare/sk/refator-advanceSessionWithSelections...sk/refator-advanceSessionWithSelections-safe?expand=1)

snopoke (approved):

PR #1029, opened 2021-11-05
 Switch case search to make a post request

esoergel:
https://dimagi-dev.atlassian.net/browse/USH-1308
To be merged with https://github.com/dimagi/commcare-core/pull/1054

I'm new to the codebase and to Java, so feel free to provide opinionated feedback.

### DIFF
--- a/src/main/java/org/commcare/formplayer/screens/FormplayerQueryScreen.java
+++ b/src/main/java/org/commcare/formplayer/screens/FormplayerQueryScreen.java
@@ -5,10 +5,6 @@ import org.commcare.suite.model.QueryPrompt;
 import org.commcare.util.screen.QueryScreen;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
-import org.springframework.web.util.UriComponentsBuilder;
-
-import java.net.URI;
-import java.net.URL;
 
 import com.google.common.collect.*;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/commcare/formplayer/screens/FormplayerQueryScreen.java
```diff
@@ -42,4 +45,24 @@ public URI getUri(boolean skipDefaultPromptValues) {
         }
         return builder.build().toUri();
     }
+
+    public MultiValueMap getRequestData(boolean skipDefaultPromptValues) {
```
esoergel, line 20: This is largely pulled from `getUri` above, which is deleted in a later commit

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -39,16 +38,16 @@
 
     public ExternalDataInstance getSearchDataInstance(FormplayerQueryScreen screen,
                                                       boolean skipDefaultPromptValues) {
-        URI uri = screen.getUri(skipDefaultPromptValues);
+        MultiValueMap<String, String> queryParams = screen.getRequestData(skipDefaultPromptValues);
         Cache cache = cacheManager.getCache("case_search");
-        String cacheKey = getCacheKey(uri);
+        String cacheKey = getCacheKey(queryParams.hashCode());
```
esoergel, line 43: this hash code could in theory have collisions - is that something we need to worry about?  Any other ideas for getting a unique code to use to identify a `queryParams` state?

shubham1g5: I think we would also want to include the `uri` here as part of cache key so that 2 different urls with same params don't get treated as equals.

shubham1g5: we can probably append all key-value pairs as it is in the `cacheKey` instead of the hash ? Collisions are very unlikely to happen though a bug like that would be a hard one to identify in future.

esoergel: Good point - I guess the domain is only provided in the URI, not in the parameters, and that'd be important to vary on.

I can append the key-value pairs directly - do you think there's any risk of the cache key being too big and overflowing something?  In particular, the point of switching this to a post request is to support parameter sets that are many thousands of characters long

shubham1g5: the cache key is `String`. So I think it's ok as long as we don't overshoot the `String` length limits

shubham1g5 (approved):
Looks good to me, though we should probably do a minor QA for this ?

esoergel:
This has crazy merge conflicts with the CJW work, so I'm gonna hold off QA and merge until that PR is in.

snopoke:
closed in favour of https://github.com/dimagi/formplayer/pull/1047

PR #1033, opened 2021-11-12
refator advance session with selections

snopoke:
The main impetus for this refactor was to get `cc-auto-advance-menu` to work when there are queries after the last selection. Previously the auto-advance logic was executed during screen selection processing. When there is a query after the last selection the auto-advance was skipped.

This refactor moves all non-selection based navigation to it's own method and runs the logic in a loop as long as the session was advanced which makes it a more generalised solution.

The refactor also allowed simplifying some other logic.

Best reviewed by commit.

I made this branch on top of https://github.com/dimagi/formplayer/pull/1028 since that is where the query logic is added but I do not intend to merge it into that branch and will keep it separate until #1028 is merged.

Cross request: https://github.com/dimagi/commcare-core/pull/1057
Jira: https://dimagi-dev.atlassian.net/browse/USH-1381

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -372,11 +372,10 @@ public class MenuSessionRunnerService {
         if ((queryData != null && queryData.getExecute(queryKey)) || autoSearch) {
             NotificationMessage notificationMessage = doQuery(
                     queryScreen,
-                    menuSession,
                     queryData == null ? null : queryData.getInputs(queryKey),
                     queryScreen.doDefaultSearch() && !forceManualAction
             );
-            if (notificationMessage != null && notificationMessage.isError()) {
+            if (notificationMessage.isError()) {
                 throw new CommCareSessionException(notificationMessage.getMessage());
             }
             return true;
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -475,39 +472,26 @@ private NotificationMessage doSync(FormplayerSyncScreen screen) throws Exception
...
 
-        String error = null;
         ExternalDataInstance searchDataInstance;
         try {
             searchDataInstance = searchAndSetResult(
                     screen,
                     screen.getUri(skipDefaultPromptValues));
             if (searchDataInstance == null) {
-                error = "No result from query";
+                throw new CommCareSessionException("No result from query");
```
orangejenny: Switching to an exception doesn't change the user-facing behavior, does it?

orangejenny (approved):

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -318,79 +305,116 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
...
+     * @param currentInput The current input being processed
+     * @param nextInput The next input being processed or NO_SELECTION constant
+     * @param queryData Query data from the request
+     * @param needsDetail Whether the full entity screen is required
+     * @param inputValidated Whether the input has been validated (allows skipping validation)
+     * @param forceManualAction Prevent auto execution of queries if true.
+     * @return
+     * @throws CommCareSessionException
+     */
+    private Screen autoAdvanceSession(
```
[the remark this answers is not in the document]
snopoke: The sync screen is a special case because it's really an `entry` type so we have to treat it as terminal step that rebuilds the session. If it were possible to put sync steps in between other session steps then it would make sense to put that in this method.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
snopoke: no, this would have resulted in an exception before anyway: https://github.com/dimagi/formplayer/blob/49a245a82ef33f61a05394a965c60438ddc1f7ac/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L349-L357

The exception handling is done here: https://github.com/dimagi/formplayer/blob/e638a9e628b470ce9d0df2571f1b748a815d4a5c/src/main/java/org/commcare/formplayer/application/AbstractBaseController.java#L110-L125

PR #1045, opened 2021-11-22
refactor restore factory getUrl and getHeaders

snopoke:
This is a (almost) pure refactor with the following goals:

* simplify logic
* create a single entry point for collecting request headers
* tests for the logic

One side effect here is that the `X-OpenRosa-Version` header is now always set to `3.0` where before it was sometimes `2.0`. I consider this an improvement.

The other side effect of the separation between restore type and the request headers is that HMAC auth is no longer enforced 'case restore' requests. I don't think that is and issue.

Best reviewed by commit.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -431,10 +431,9 @@ public class RestoreFactory {
     public InputStream getRestoreXml(boolean skipFixtures) {
         ensureValidParameters();
         URI url = getRestoreUrl(skipFixtures);
-        HttpHeaders headers = getRequestHeaders(url);
         recordSentryData(url.toString());
         log.info("Restoring from URL " + url);
-        InputStream restoreStream = getRestoreXmlHelper(url, headers);
+        InputStream restoreStream = getRestoreXmlHelper(url);
         setLastSyncTime();
         return restoreStream;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -429,20 +424,27 @@ public boolean isRestoreXmlExpired() {
...
+        recordSentryData(url.toString());
+        log.info("Restoring from URL " + url);
+        InputStream restoreStream = getRestoreXmlHelper(url);
         setLastSyncTime();
         return restoreStream;
     }
 
+    public HttpHeaders getRequestHeaders(URI url) {
+        HttpHeaders headers;
+        if (getHqAuth() == null) {
```
shubham1g5, line 439: It looks like Hq auth is expected to be null mostly for the sms requests. If I am right about that, it seems like the case restore requests in a normal (non-sms) flow will now won't have a Hmac auth unlike before when they were all Hmac. Do you see any issue with that ?

shubham1g5 (commented):
Looks good to me, just seeking a minor clarification.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
snopoke: > It looks like Hq auth is expected to be null mostly for the sms requests

As far as I know this is only for SMS workflows and never happens in normal user workflows.

> If I am right about that, it seems like the case restore requests in a normal (non-sms) flow will now won't have a Hmac auth unlike before when they were all Hmac.

Where are case restores coming from outside of SMS workflows? 

> Do you see any issue with that ?

Even if that is the case I think we should always prefer use auth over HMAC auth if it is available. I do see however that the case restores endpoint only accepts HMAC auth so if user auth is ever present then that will be an issue.

shubham1g5: Sounds good. I didn't realise that case restores are specific to sms workflows.

shubham1g5 (approved):

PR #1049, opened 2021-11-26
refactor and tests for 

snopoke:
Builds on https://github.com/dimagi/formplayer/pull/1046 with the goal of adding a test for the auth changes.

The refactors allow for easier testing by reducing the dependencies that are required by the controller. I'd like to split the controllers up more but that's for another day.

Best reviewed by commit.


snopoke:
This is on staging

src/main/java/org/commcare/formplayer/application/UtilController.java
```diff
@@ -35,12 +35,20 @@
  */
 @RestController
 @EnableAutoConfiguration
-public class UtilController extends AbstractBaseController {
```
snopoke, line 38: I realized that removing this base class also removes the exception handling. Solved by moving the exception handling to a global advice: https://github.com/dimagi/formplayer/pull/1056

snopoke:
Messed up a rebase: reopening as https://github.com/dimagi/formplayer/pull/1058

PR #1065, opened 2021-12-08
refactor restore URLs to use URL templates from properties

snopoke:
Moves the restore URLs outside of the code and into the properties file. This also uses URL templates which handles all the encoding etc. which means we don't have to do it.

I started working on this because of the upcoming change in the "case restore" URL: https://github.com/dimagi/commcare-hq/pull/30826.

This does change the encoded URLs slightly since it also encodes characters like `*`. This is not an issue since it is a standard encoding which Django handles transparently.


shubham1g5 (approved):
💯

snopoke:
Also tested this on staging

PR #1075, opened 2022-01-05
Use postgres form definition if available

gherceg:
This change stems from https://dimagi-dev.atlassian.net/browse/SAAS-12385 and is a followup to dimagi/formplayer#1005. This PR is based on `gh/form-def/base` which includes the previously mentioned PR since that has yet to be merged. 

This stores a serialized form definition in postgres. Previously, a form definition was saved within a `SerializedFormSession` under the column `formXml`, meaning every form session had its own copy of the associated form definition. This is an unnecessary use of space, and by moving the definition to its own table, we can setup a relationship with the `SerializedFormSession` table to reference the relevant form definition. 

A new form definition is created for every new combination of app id, app version, and form xmlns. I added the `@Cacheable` annotation since this will frequently be called for the same form definition as part of filling out a form.

Dependent on the following HQ changes
- https://github.com/dimagi/commcare-hq/pull/31384

[QA Ticket](https://dimagi-dev.atlassian.net/browse/QA-3798)


gherceg:
I realized that having caching on the FormDefinitionRepo method masked an issue with the underlying hibernate generated query. The tests for the FormDefinitionRepo pass with the same query, but does not appear to function as I had hoped when running the app. The generated query for `findByAppIdAndAppVersionAndXmlns` is 
```
Hibernate: select formdefini0_.id as id1_0_, formdefini0_.appid as appid2_0_, formdefini0_.appversion as appversi3_0_, formdefini0_.datecreated as datecrea4_0_, formdefini0_.formdef as formdef5_0_, formdefini0_.xmlns as xmlns6_0_ from form_definition formdefini0_ where formdefini0_.appid=? and formdefini0_.appversion=? and formdefini0_.xmlns=?
```
which seems strange, but is also what works in tests. I have yet to figure this out so I wanted to point it out to draw attention to the `FormDefinitionRepo` and `FormDefinitionService` implementations.

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
```diff
@@ -0,0 +1,39 @@
...
+@CacheConfig(cacheNames = {"form_definition"})
+public class FormDefinitionService {
+
+    private final Log log = LogFactory.getLog(FormDefinitionService.class);
+
+    @Autowired
+    private FormDefinitionRepo formDefinitionRepo;
+
+
+    @Cacheable
```
snopoke, line 25: No need to include the full form def in the cache key:

```suggestion
    @Cacheable(key={ #appId, #appVersion, #formXmlns }
```

snopoke (commented):
Looks OK overall but I would try to push the serialization / deserialization down into the DB / service layer so that it can be done only when necessary.

How do you plan to roll this out without breaking existing sessions?

src/main/java/org/commcare/formplayer/objects/FormDefinition.java
```diff
@@ -0,0 +1,51 @@
...
+    private String appVersion;
+
+    @Column(name="xmlns", updatable=false)
+    private String xmlns;
+
+    @CreatedDate
+    @Column(name="datecreated")
+    private Instant dateCreated;
+
+    @Setter
```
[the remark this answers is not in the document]
gherceg: e33f2195c55b9d877723166e4b17390067c05145

src/main/java/org/commcare/formplayer/repo/FormDefinitionRepo.java
```diff
@@ -0,0 +1,11 @@
+package org.commcare.formplayer.repo;
+
+import org.commcare.formplayer.objects.FormDefinition;
+import org.springframework.data.jpa.repository.JpaRepository;
+
+import java.util.List;
+
+public interface FormDefinitionRepo extends JpaRepository<FormDefinition, String> {
+    List<FormDefinition> findByAppIdAndAppVersionAndXmlns(String appId, String appVersion, String xmlns);
```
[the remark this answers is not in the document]
gherceg: Good idea. See e33f2195c55b9d877723166e4b17390067c05145

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
gherceg: Was trying to find the write docs to reference how to do this, so thanks for flagging. See e33f2195c55b9d877723166e4b17390067c05145

src/test/java/org/commcare/formplayer/repo/FormSessionRepoTest.java
```diff
@@ -186,6 +190,25 @@ public void testDeleteByDateCreatedLessThan() {
...
+                "appId",
+                "appVersion",
+                "formXmlns",
+                "formXml"
+        );
+        formDefinitionRepo.save(formDef);
+        SerializableFormSession session = getSession();
+        session.setFormDefinition(formDef);
+        formSessionRepo.save(session);
+        FormDefinition fetchedFormDef = session.getFormDefinition();
```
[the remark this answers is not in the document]
gherceg: e33f2195c55b9d877723166e4b17390067c05145

src/main/java/org/commcare/formplayer/objects/FormDefinition.java
```diff
@@ -0,0 +1,51 @@
...
+    @Column(name="formdef")
+    private String serializedFormDef;
```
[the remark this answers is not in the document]
gherceg: I like this idea, and spent part of today looking into this. I hit a blocker when realizing we rely on the serialization/deserialization as a way of making a copy of the form def that we can then inject with the instance xml. By pushing serialization down to the db layer, we still need to make a deepcopy of the form def before injecting it with the xml which effectively entails serializing/deserializing again. 

Our end goal is to cache the deserialized form def to help speed up answer requests, but caching has the same issue where a copy of the object would need to be made before modifying it. Planning to continue looking into this, but if you have any ideas I'm interested.

src/main/java/org/commcare/formplayer/db/migration/V22__create_form_definition.java
```diff
@@ -0,0 +1,26 @@
...
+        return Arrays.asList(
+                "CREATE TABLE form_definition (\n" +
+                        "    id text PRIMARY KEY,\n" +
+                        "    formsessions text,\n" +
+                        "    datecreated timestamp with time zone,\n" +
+                        "    appid text NOT NULL,\n" +
+                        "    appversion text NOT NULL,\n" +
+                        "    xmlns text NOT NULL,\n" +
+                        "    formdef text\n" +
+                        ")",
```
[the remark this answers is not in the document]
gherceg: Do we have a preference between defining constraints in sql vs hibernate? I wasn't having much success with hibernate so I opted for sql. See 785ac7334078e2856eed194cdf25cfa2c4aa72ad

gherceg:
> How do you plan to roll this out without breaking existing sessions?

Yeah this is something I overlooked. I'll add a commit where I support the previous FormSession constructor and some logic to check if a `FormDefinition` is set on the session or not, and use the `formXml` if needed.

src/main/java/org/commcare/formplayer/objects/FormDefinition.java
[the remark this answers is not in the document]
snopoke: that makes sense but I still think it would be good to try and push some of this down into the service layer to make it more consistent

src/main/java/org/commcare/formplayer/db/migration/V22__create_form_definition.java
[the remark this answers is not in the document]
snopoke: There's no need to do it in hibernate since we are managing the DB migrations separately.

gherceg:
Added support for the old workflow here 1cafa73ec55add312a5f62a8a6e0918e3b7c3f9f. Any new session that is created after this is rolled out should have a `FormDefinition` associated with it, so we should only need to cater to existing sessions that do not have a reference to a form definition. 

The `FormSession` currently doesn't know how to create a `FormDefinition`, and I don't think it is worth adding just to support transitioning sessions without a form definition reference for 1 week (when incomplete sessions that depend on the old workflow will be purged). At that point, we can remove this code since all existing sessions will have a form definition reference.

src/main/java/org/commcare/formplayer/objects/FormDefinition.java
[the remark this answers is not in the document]
gherceg: I'm definitely in favor of cleaning this up, but with some recent developments for the followup task, we are likely going to rely on an object pool that stores deserialized form defs and use that to fetch form defs on each request. 

All that to say, there are certainly going to be changes around the form def serialization/deserialization process, so I am tempted to hold off on any changes in this PR. If that doesn't sit well you with definitely let me know.

src/main/java/org/commcare/formplayer/db/migration/V22__create_form_definition.java
```diff
@@ -0,0 +1,26 @@
+package org.commcare.formplayer.db.migration;
+
+import java.util.Arrays;
+
+public class V22__create_form_definition extends BaseFormplayerMigration {
+    @Override
+    public Iterable<String> getSqlStatements() {
+        return Arrays.asList(
+                "CREATE TABLE form_definition (\n" +
+                        "    id text PRIMARY KEY,\n" +
```
snopoke, line 10: just looking at this again now I'm wondering why this needs to be a UUID. I don't think it matters too much though integer primary keys are marginally more performant. If we do stick with a UUID then this column type should be UUID and not `text` and the Java field should also be a UUID and not a `string` which would also simplify the mapping slightly:

```java
    // if we stick with a UUID then the mapping can be as follows:
    @Id
    @GeneratedValue
    private UUID id;
```

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
```diff
@@ -68,9 +73,12 @@ public NewFormResponse getResponse(NewSessionRequestBean bean, String postUrl) t
                 bean.getRestoreAs(),
                 bean.getRestoreAsCaseId());
 
+        FormDef formDef = parseFormDef(formXml);
+        String serializedFormDef = FormDefStringSerializer.serialize(formDef);
```
snopoke, line 77: This is what I was thinking could be moved into the service layer so we avoid doing the serialisation if we're loading it from the DB

src/main/java/org/commcare/formplayer/exceptions/MultipleFormDefsFoundException.java
```diff
@@ -0,0 +1,8 @@
+package org.commcare.formplayer.exceptions;
+
+public class MultipleFormDefsFoundException extends RuntimeException {
```
snopoke, line 3: this is now unused

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -303,12 +302,14 @@ private void computeDatum() {
 
     public FormSession getFormEntrySession(FormSendCalloutHandler formSendCalloutHandler,
                                            FormplayerStorageFactory storageFactory,
-                                           CaseSearchHelper caseSearchHelper) throws Exception {
+                                           CaseSearchHelper caseSearchHelper,
+                                           FormDefinitionService formDefinitionService) throws Exception {
         String formXmlns = sessionWrapper.getForm();
-        FormDef formDef = engine.loadFormByXmlns(formXmlns);
+        String serializedFormDef = FormDefStringSerializer.serialize(this.engine.loadFormByXmlns(formXmlns));
```
snopoke, line 308: same comment here about moving the serialisation into the service

src/main/java/org/commcare/formplayer/objects/FormDefinition.java
```diff
@@ -0,0 +1,46 @@
...
+    private String xmlns;
+
+    @CreatedDate
+    @Column(name="datecreated")
+    private Instant dateCreated;
+
+    @Column(name="formdef")
+    private String serializedFormDef;
+
+    protected FormDefinition(){}
```
calellowitz, line 39: Why is this constructor necessary?

src/main/java/org/commcare/formplayer/db/migration/V22__create_form_definition.java
calellowitz: Just commenting to say I agree with Simon

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -119,7 +125,7 @@ public FormSession(SerializableFormSession session,
     }
 
     public FormSession(UserSqlSandbox sandbox,
-                       FormDef formDef,
+                       FormDefinition formDefinition,
```
calellowitz, line 128: the fact that `formDef` and `formDefinition` are different types here is very confusing.

calellowitz (approved):
Few comments/questions but nothing blocking

src/main/java/org/commcare/formplayer/session/FormSession.java
snopoke: the other classes are called `SerializableFormSession` and `SerializableMenuSession`. I don't love the convention but maybe we should stick with it?

src/main/java/org/commcare/formplayer/objects/FormDefinition.java
snopoke: The default constructors are required by Hibernate

src/main/java/org/commcare/formplayer/session/FormSession.java
gherceg: Yeah I was concerned about this naming, and agree that sticking to convention is wise.

gherceg: See f209feb0bb30728edbbfc693a091904384750d35

src/main/java/org/commcare/formplayer/session/MenuSession.java
gherceg: See ed0229dc5755f8fb95512c5a60640cafa43ea457

src/main/java/org/commcare/formplayer/exceptions/MultipleFormDefsFoundException.java
gherceg: See c231fefd5255161f63c4133f825ac4a03ff516ad

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
gherceg: See ed0229dc5755f8fb95512c5a60640cafa43ea457

src/main/java/org/commcare/formplayer/db/migration/V22__create_form_definition.java
gherceg: See cc72e727ce137f742b6c4743acf2325509059aa9

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
```diff
@@ -0,0 +1,41 @@
...
+    @Cacheable(key="{#appId, #appVersion, #formXmlns}")
+    public SerializableFormDefinition getOrCreateFormDefinition(String appId, String appVersion, String formXmlns, FormDef formDef) throws IllegalArgumentException {
+        Optional<SerializableFormDefinition> optFormDef = this.formDefinitionRepo.findByAppIdAndAppVersionAndXmlns(appId, appVersion, formXmlns);
+        return optFormDef.orElseGet(() -> {
+            try {
+                String serializedFormDef = FormDefStringSerializer.serialize(formDef);
+                SerializableFormDefinition newFormDef = new SerializableFormDefinition(appId, appVersion, formXmlns, serializedFormDef);
+                return this.formDefinitionRepo.save(newFormDef);
+            } catch (IOException ex) {
+                throw new IllegalArgumentException();
```
snopoke, line 37: This isn't the right exception to throw here. Either create a new one or use `org.javarosa.core.log.WrappedException`:

```java
throw new WrappedException("Error serializing form def", e);
```

gherceg: Ah I see. Java doesn't seem to handle throwing checked exceptions from within a lambda well, so that was my workaround. Used WrapperException in c549f0b27d62bedfb8dee767e0a61b4292c50625.

src/main/java/org/commcare/formplayer/objects/SerializableFormDefinition.java
```diff
@@ -0,0 +1,42 @@
...
+    @Column(name="appid", updatable = false)
+    private String appId;
+
+    @Column(name="appversion", updatable = false)
+    private String appVersion;
+
+    @Column(name="xmlns", updatable = false)
+    private String xmlns;
+
+    @Column(name="formdef", updatable = false)
```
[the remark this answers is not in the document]
gherceg: Removed in c549f0b27d62bedfb8dee767e0a61b4292c50625

src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
```diff
@@ -65,6 +65,11 @@
     @Column(name="instancexml")
     private String instanceXml;
 
+    @Setter
+    @ManyToOne(fetch=FetchType.LAZY)
```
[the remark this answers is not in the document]
snopoke: I usually try and think about the usage scenarios. If we're always (or most often) going to need the form def then it makes sense to use eager loading. But if we only need it under certain conditions then lazy loading makes more sense. 

For this scenario I think we always need it right?

gherceg:
Thanks again @snopoke for all of the feedback. If you and/or others feel this is at a good spot for QA, I can kick that off. Is there anything unique to keep in mind before rolling this out to staging (and eventually prod) since it does entail formplayer postgres changes? Or is that all handled with the current deploy script?

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -138,13 +144,14 @@ public FormSession(UserSqlSandbox sandbox,
...
-        session = new SerializableFormSession(
+        // use this.formDef to mutate (e.g., inject instance content, set callout handler)
+        this.formDef = FormDefStringSerializer.deserialize(serializableFormDefinition.getSerializedFormDef());
+        this.session = new SerializableFormSession(
                 domain, appId, TableBuilder.scrubName(username), asUser, caseId,
-                postUrl, menuSessionId, formDef.getTitle(), oneQuestionPerScreen,
+                postUrl, menuSessionId, this.formDef.getTitle(), oneQuestionPerScreen,
                 locale, inPromptMode, sessionData, functionContext
         );
-        session.setFormXml(FormDefStringSerializer.serialize(this.formDef));
```
snopoke, line 147: I'm just wondering if we shouldn't keep saving the formDef to session directly for now otherwise it will make it very hard to roll this back if the need arises.

snopoke (commented):
I think this is ready to test on staging. I think the only outstanding thing is to think about rollback. As it stands it would be hard to roll this back since we aren't saving the data on the form session anymore. 

I've made one comment in code where I think we should keep saving the form def to the session for now purely for the sake of rollback.

src/main/java/org/commcare/formplayer/session/FormSession.java
gherceg: Realized I never responded, but addressed on 0bf6f5d48060f86769fddde01d22e9a0d1367e6b

gherceg:
So after hitting an issue with sms surveys not having access to an app version, I realized it might be a good time to switch to using the form version since that is more accurate. That is what prompted the commcare-core https://github.com/dimagi/commcare-core/pull/1079 which would require a db version bump and rebuilding local dbs, and we don't have a clear process for that. I've created a SaaS ticket to address that separately https://dimagi-dev.atlassian.net/browse/SAAS-13364, but in the meantime moved back to using `appVersion` and will resolve the sms forms issue by fixing an HQ bug that I will link here.

Notably I kept the name change from appVersion to formVersion because I believe that it accurately represents the intention of what that value should be, and appVersion isn't inaccurate, but just potentially creates more unique values than necessary.

gherceg:
Also flagging here that Cal mentioned to double check that build profiles were not used in formplayer since that would require adding that field to shared form definitions and play a role in the unique id. From what I can tell in formplayer we do not make any use of build profiles so this should not be a concern for now.

snopoke:
> Also flagging here that Cal mentioned to double check that build profiles
> were not used in formplayer since that would require adding that field to
> shared form definitions and play a role in the unique id. From what I can
> tell in formplayer we do not make any use of build profiles so this should
> not be a concern for now.
>

That's correct. Formplayer does not use build profiles.

>

gherceg:
Moved the migration out of this PR and into https://github.com/dimagi/formplayer/pull/1119. There were minor diffs that I missed in that PR (formatting issue and a docstring typo) which is why this PR still modifies the migration file and the SerializableFormSession file.

gherceg:
@snopoke @calellowitz can you re-review this when you get a chance? QA has passed on it

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -145,13 +151,16 @@ public FormSession(UserSqlSandbox sandbox,
             @Nullable SessionFrame sessionFrame,
             RemoteInstanceFetcher instanceFetcher) throws Exception {
 
-        this.formDef = formDef;
-        session = new SerializableFormSession(
+        // use this.formDef to mutate (e.g., inject instance content, set callout handler)
+        this.formDef = FormDefStringSerializer.deserialize(serializableFormDefinition.getSerializedFormDef());
```
calellowitz, line 155: It looks like we take a `FormDef`, convert it to a `serializableFormDefinition`, pass that to this constructor, the immediately derserialize it to get a `FormDef` back. Is that not expensive?

calellowitz (commented):
One question but otherwise looks good.

src/main/java/org/commcare/formplayer/session/FormSession.java
gherceg: Yeah good point. Would it be odd to add a non-postgres property to SerializableFormDefinition to hold the deserializedFormDef? I think I was hesitant to do that originally and then didn't come back to optimizing this part.

gherceg: Ok I think that isn't allowed since hibernate isn't able to determine what column to map that property to, unless there is a way to omit properties from hibernate that I don't know of. 

Alternatively, pushing the form definition service/creation of a serializable form definition into the form session seems like the next best solution. This would require passing a few additional args into the form session constructor which is already bloated, but seems worth it.

snopoke: You should be able to use the [@Transient](https://docs.oracle.com/javaee/7/api/javax/persistence/Transient.html) annotation to prevent hibernate from mapping the field.

snopoke (approved):

src/main/java/org/commcare/formplayer/session/FormSession.java
gherceg: That's perfect. Thanks for highlighting Simon. See 3d32f0ddb7ea2538e7cfedbeef8dae423c0fba3a

src/test/java/org/commcare/formplayer/repo/FormSessionRepoTest.java
```diff
@@ -198,6 +202,25 @@ public void testDeleteByDateCreatedLessThan() {
...
+                "appId",
+                "formXmlns",
+                "formVersion",
+                "formXml"
+        );
+        this.formDefinitionRepo.save(formDef);
+        SerializableFormSession session = getSession();
+        session.setFormDefinition(formDef);
+        this.formSessionRepo.save(session);
+        SerializableFormDefinition fetchedFormDef = this.formDefinitionRepo.getById(formDef.getId());
```
snopoke, line 217: shouldn't this test be checking that you can get the SerializableFormDefinition from the SerializableFormSession without having to fetch it directly?

gherceg:
I decided to revert the addition of a `formDef` attribute on the `SerializableFormDefinition` as it felt like an after thought, and also caused issues with https://github.com/dimagi/formplayer/pull/1123 since it held onto a FormDef that is _supposed_ to only be cached by session, but was being shared across sessions.

I don't love passing both the `SerializableFormDefinition` and `FormDef` into the `FormSession` constructor, but I think it'll do for now.

gherceg:
Alright I've created a new PR with this work since this has gotten a bit out of control in my opinion. That PR does not contain any differences besides more tests around the FormDefinitionService. Otherwise it is the same. I'll request QA to take another pass on staging to ensure things still look good as well. Closing this one out.

PR #1076, opened 2022-01-06
Tracing

orangejenny:
Merging this into master to reduce merge conflicts on staging. See https://dimagi-dev.atlassian.net/browse/USH-1403?focusedCommentId=182839

Deployed this to staging after switching the DD library and verified that traces are still coming in.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,6 +108,7 @@ dependencies {
     implementation group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-jdk8'
 
     //Unmanaged Dependencies
+    implementation group: 'com.datadoghq', name: 'dd-trace-ot', version: '0.75.0'
     implementation group: 'com.datadoghq', name: 'java-dogstatsd-client', version: '3.0.1'
     implementation group: 'io.sentry', name: 'sentry-spring-boot-starter', version: "${sentryVersion}"
     implementation group: 'io.sentry', name: 'sentry-logback', version: "${sentryVersion}"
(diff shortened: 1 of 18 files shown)


shubham1g5 (approved):
Looks good to me. One concern I would have on this is the impact on the mobile apk size due to addition of these libraries in the core code. Do we know if there is considerable change in the size of formplayer jar on adding these libs ?

calellowitz:
Is this going to cause traces to come in from production as well? That is the only concern I would have (as that would be really expensive), and I think is the reason I left it as a staging branch, but I don't totally remember if production is set up that these would flow in too.

calellowitz (approved):

orangejenny:
@shubham1g5 Good point. It looks like trace library adds about 128K. How significant is that for the APK?
```
(ansible3) ~/Documents/formplayer (ce/tracing) $ ls -al *.jar
-rw-r--r--  1 jschweers  staff  107566431 Jan  7 14:31 formplayer_with_trace_lib.jar
-rw-r--r--  1 jschweers  staff  107438150 Jan  7 14:29 formplayer_without_trace_lib.jar
```

shubham1g5:
@orangejenny Thanks for checking. That does look like a minor change and should not be an issue.

snopoke:
> Is this going to cause traces to come in from production as well? That is the only concern I would have (as that would be really expensive), and I think is the reason I left it as a staging branch, but I don't totally remember if production is set up that these would flow in too.

@calellowitz this should not add traces to production since we don't run formplayer with the agent in production. Quoting the [docs](https://docs.datadoghq.com/tracing/setup_overview/setup/java/?tab=containers#trace-annotation):

> Now add @Trace to methods to have them be traced when running with dd-java-agent.jar. If the Agent is not attached, this annotation has no effect on your application.

snopoke (approved):

PR #1080, opened 2022-01-19
Transition to GitHub actions

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-1502


.github/application.properties.ci
```diff
@@ -15,7 +15,7 @@ sentry.dsn=
 spring.datasource.driver-class-name=org.postgresql.Driver
 spring.datasource.url=jdbc:postgresql://localhost:5432/formplayer?prepareThreshold=0
 spring.datasource.username=postgres
-spring.datasource.password=
+spring.datasource.password=postgres
```
ctsims, line 18: Is this the recommended way to set secrets for automation tests (honestly not sure)?

snopoke: This isn't really a secret per se since it's defined in the build configuration file. I do plan to switch to using environment variables for these and dropping this file entirely but was leaving that for a future PR.

Actual secrets are set in the repo settings and accessible in the build configuration as part of the action context: https://docs.github.com/en/actions/security-guides/encrypted-secrets

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Docker image
+
+on:
+  push:
+    branches: [master]
```
shubham1g5, line 5: does this mean that the Formplayer will be continuously to production as soon as we merge into master ?

shubham1g5 (approved):

.github/workflows/publish_docker.yml
snopoke: not, it means that the docker image will be updated after each merge to master

shubham1g5: I see, purely for my own understanding it's only going to effect someone who runs FP on docker locally ? I am not sure where else does the FP docker image get used.

snopoke: currently the only use for the docker images is local development and the commcare-hq build

PR #1083, opened 2022-01-21
Respect relevancy when returning instance xml for data preview

gherceg:
https://dimagi-dev.atlassian.net/browse/SUPPORT-12366

Bug introduced by [this PR](https://github.com/dimagi/formplayer/pull/1005) which defaulted `getInstanceXml()` to serialize all data (ignore relevancy). In addition to this identified spot, I'm looking through other `getInstanceXml()` calls to determine if it is expected that they serialize all data.

I replicated Clayton's [test app](https://www.commcarehq.org/a/corpora/apps/view/2d262ab3b3c5090876b15ba1e1021ef7/form/91f50ad022bf45cbbab460b32b37a6a9/source/#form/the_hidden_value_checker) locally to reproduce the issue. 
Before the fix:
Screen Shot 2022-01-21 at 3 15 15 PM
After the fix:
Screen Shot 2022-01-21 at 3 12 39 PM

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/DebuggerController.java
+++ b/src/main/java/org/commcare/formplayer/application/DebuggerController.java
@@ -61,7 +61,7 @@ public class DebuggerController extends AbstractBaseController {
         SerializableFormSession serializableFormSession = formSessionService.getSessionById(debuggerRequest.getSessionId());
         SerializableMenuSession serializableMenuSession = menuSessionService.getSessionById(serializableFormSession.getMenuSessionId());
         FormSession formSession = getFormSession(serializableFormSession);
-        String instanceXml = formSession.getInstanceXml();
+        String instanceXml = formSession.getInstanceXml(false);
         FormattedQuestionsService.QuestionResponse response = formattedQuestionsService.getFormattedQuestions(
                 debuggerRequest.getDomain(),
                 serializableMenuSession.getAppId(),


gherceg:
There are 2 other areas with similar behavior that I'd like to clarify.

1) https://github.com/dimagi/formplayer/blob/182a52f1daa85af419767339889d234a28f41671/src/main/java/org/commcare/formplayer/beans/GetInstanceResponseBean.java#L19-L22

Appears to only be used in the context of the `get-instance` call, which maps to [this code in HQ](https://github.com/dimagi/commcare-hq/blob/2d00204fb009c80f9514831060d6e2302bf15d2f/corehq/apps/formplayer_api/smsforms/api.py#L279-L283). Based on that comment about used to store incomplete forms, continuing to serialize all data is the right approach for this case.

2) https://github.com/dimagi/formplayer/blob/eae7c7b00e4c7b4151d4b148c8dc4b9c5cae1cff/src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java#L124-L132

Appears to be used in the context of opening an incomplete form, which again would be safe to ignore relevance. https://github.com/dimagi/formplayer/blob/ef7058933b4360770be15d48e351f343a42d2b5b/src/main/java/org/commcare/formplayer/application/IncompleteSessionController.java#L44

Those were the only 2 `getInstanceXml()` calls on a `FormSession` that jumped out to me.

ctsims (approved):
This change looks good, can check the other two linked examples in a bit

ctsims:
@gherceg Answering your other two questions:

1. Agreed - Based on the documentation for [this call](https://github.com/dimagi/commcare-hq/blob/af58ee81fda9bafa9647b75e1dc10d0bde4f9b29/corehq/apps/smsforms/app.py#L92) I think it's explicitly correct to store the full copy of the form XML

2. Agreed. A NewFormResponse feels fine to store the full body.

gherceg:
Ignoring the code coverage failure as the lift to add tests to this area is too large compared to the urgency of this bug. I will create a ticket to investigate what tests should/could have been in place to catch this bug to begin with.

Tested on staging and things look good. Merging.

PR #1084, opened 2022-01-26
use postgres:9.6 in build instead of latest

snopoke:


shubham1g5 (approved):
Out of curiosity, did you run into any issues with latest ?

snopoke:
> Out of curiosity, did you run into any issues with latest ?

No but some of the dependency upgrades don't support 9.6 so I wanted to make sure the build fails for those. e.g. new version of flyway no longer supports 9.6

PR #1086, opened 2022-01-26
Gradle 7 update

shubham1g5:
cross-request: https://github.com/dimagi/commcare-core/pull/1065/files

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ allprojects{
 
 repositories {
     mavenCentral()
+    maven { url "https://jitpack.io" }
 }
 
 ext {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -43,6 +43,7 @@ allprojects{
 
 repositories {
     mavenCentral()
+    maven { url "https://jitpack.io" }
```
shubham1g5, line 46: without this the kxml dependency from https://github.com/dimagi/commcare-core/blob/gradleUpdate/build.gradle#L70 doesn't get resolved.

snopoke (approved):

PR #1087, opened 2022-01-27
Markdown Fix for Case Detail 

shubham1g5:
Reverts dimagi/formplayer#963 in favour of the fix https://github.com/dimagi/commcare-core/pull/1066

https://dimagi-dev.atlassian.net/browse/INDIV-29

The earlier PR added template-forms like `markdown` as a separate property when we already had a field for it as `details[].styles[].displayFormat` but the issue was just that it was not populating `markdown` correctly which is the issue the [core PR](https://github.com/dimagi/commcare-core/pull/1066
) addresses.  This is how the JSON response now looks like with the fix - 
```
{
  "shouldRequestLocation": false,
  "shouldWatchLocation": false,
  "details": [
    {
      "details": [
        "Empty P1",
        "*tada*"
      ],
      "entities": null,
      "styles": [
        {
          "fontSize": 0,
          "widthHint": -1,
          "displayFormat": null
        },
        {
          "fontSize": 0,
          "widthHint": -1,
          "displayFormat": "Markdown"
        }
      ],
      "headers": [
        "Name",
        "P2"
      ],  
      "title": "Case",
      "usesCaseTiles": false,
      "maxWidth": 0,
      "maxHeight": 0,
      "numEntitiesPerRow": 0,
      "tiles": null,
      "useUniformUnits": false,
      "hasInlineTile": false,
      "useNodeset": false
    }
  ],
  "isPersistentDetail": false
}
```

@mkangia  We will need to change the Web Apps JS side of it to take the format from `details[].styles[].displayFormat` instead of `details[].templateForms[]`


ctsims (approved):

mkangia:
Thanks @shubham1g5 
Added HQ changes in https://github.com/dimagi/commcare-hq/pull/31015

I am guessing this should be merged only after HQ change is deployed otherwise there might be JS errors on HQ.

shubham1g5:
@mkangia Good point, I think for the time being HQ should look for `details[].styles[].displayFormat` but fallback to `details[].templateForms[]` to avoid any errors. Once these PRs are deployed, we can follow up by removing any references to `details[].templateForms[]` in the subsequent deploy

mkangia:
@shubham1g5 i thought this PR was not adding anything new and that `details[].styles[].displayFormat` was already available? So i was thinking HQ change would go out first and then this PR can be merged at any time after that.
Isn't that true?

shubham1g5:
@mkangia  `markdown` won't be available in `details[].styles[].displayFormat ` until we deploy the corresponding [core PR](https://github.com/dimagi/commcare-core/pull/1066/files). So that PR needs to be deployed before/with the HQ change.

mkangia:
thanks @shubham1g5 
I think then it's okay to merge both PRs together then, so they get deployed together? Does not look that fallback is necessary and markdown might not work temporarily for few minutes, just need formplayer deploy to go first. `templateForms` isn't working reliably as well anyway.

shubham1g5:
> markdown might not work temporarily for few minutes 

It might be much longer in practice because of JS caching on browser. If there is a way to force refresh JS cache I would be happy without a fallback.

> templateForms isn't working reliably as well anyway.

I think it works in most cases today except when there is an empty property before the markdown formatted property

IMO, we should not introduce a bug even if it's temporary when there is an easy path to avoid it. Here is a suggestion - 

1. Impelement fallback in HQ PR 
2. Deploy core and HQ PR together
3. After a week, remove fallback in HQ and deploy it with this FP PR.

orangejenny:
> IMO, we should not introduce a bug even if it's temporary when there is an easy path to avoid it.

Seconding this. In addition to the js caching issue, web apps users don't necessarily refresh the page often, so users can still be on the "old" HQ code for some time after the deploy.

Given that this PR is just a revert + tests, couldn't we deploy this safely without adding fallback code?
1. Deploy core PR (and formplayer submodule ref update) to populate `style` correctly
2. Deploy HQ PR to start reading from `style`. Users may continue to read from `templateForms` until their next page refresh.
3. The next week, deploy this PR, so we no longer populate `templateForm`.

shubham1g5:
Ohh great, that makes a ton of sense. Agree that approach will work out without adding fallback on HQ

mkangia:
Thank you for jumping in @orangejenny. Thought a bit more about this after my last comment and that was exactly how I was thinking this should be done.
Adding a fallback wasn't really possible effectively in HQ since we were populating any existing property for markdown now.

> to the js caching issue

The javascript should expire considering it is changed. I think that should already be handled in HQ.

> users don't necessarily refresh the page often

This is good to know. I didn't account for this.

So @shubham1g5 you would need to update this PR to NOT remove templateForms yet, right?
And, The HQ change does not need any update and is ready.

mkangia:
> IMO, we should not introduce a bug even if it's temporary

Apologies. I definitely didn't mean to add a bug. I was just thinking of the time before the code update reflected for both formplayer and HQ when they are both deployed together, considering one would finish before the other.

shubham1g5:
@mkangia We are not going to deploy this PR until the core PR and the HQ change is rolled out first. So there is not any change needed. The core PR is merged and I have updated the submodule reference in https://github.com/dimagi/formplayer/commit/84b02633ce5cd7989cebc6ec7f9a3ba986ac6aee . We should merge the HQ PR after testing on Wednesday so that the core and HQ PR goes out together this Thursday.

mkangia:
Okay cool. Looks like I don't understand the code distribution before core and formplayer repo well. I leave that call to you then.

> We should merge the HQ PR after testing on Wednesday so that the core and HQ PR goes out together this Thursday

Cool.

mkangia:
Hey @shubham1g5 

This is unblocked from HQ side. There was some push back on the deploy approach (formplayer first and then HQ), so I removed the dependency on formplayer deploy by changing the HQ implementation to support both implementations, hence it was changed to
1. first check on styles.displayFormat for "Markdown" (this is currently blank in case of Markdowns)
2. else then check on templateForms. If there are no templateForms, there will not be any error.

Thinking about this now, I believe that was the fallback you were mentioning earlier which I didn't understand and thought you were mentioning the opposite (2 and then 1), and it makes sense now.
As a follow up, i will remove the support for templateForms in HQ entirely, next week.

shubham1g5:
@mkangia FYI that I am merging this for the next deploy cycle since the HQ change has been live now for couple deploy cycles.

mkangia:
Great, thanks @shubham1g5

PR #1088, opened 2022-02-02
Sets Source and target to java 11

shubham1g5:
cross-request: https://github.com/dimagi/commcare-core/pull/1067


ctsims:
Just jumping in to echo my excitement on getting things up to speed here. Awesome job, @shubham1g5 !

build.gradle
```diff
@@ -52,8 +52,8 @@ ext {
     shedlockVersion = '4.30.0'
 }
 
-sourceCompatibility = 1.8
-targetCompatibility = 1.8
+sourceCompatibility = JavaVersion.VERSION_11
+targetCompatibility = JavaVersion.VERSION_11
```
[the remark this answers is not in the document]
shubham1g5: yes, that's right.

shubham1g5:
Most tests seems to be passing on Github actions which is a good sign that there is not any API changes needed on it. Only tests that are failing are due to change in xml serializer in the [core PR](https://github.com/dimagi/commcare-core/pull/1067). These changes are although running into several issues on Apple M1 based systems. One of the issue I am facing on M1 is outlined [here](https://github.com/xerial/sqlite-jdbc/issues/638) with Spring 2.5. Though this does not happen when I update to [Spring 2.7](https://github.com/dimagi/formplayer/pull/1144) although that PR is in turn blocked on Postgres migration. Also on Spring 2.7, I am seeing another issue that causes JVM to crash on calling `SqliteConnection.close()` on M1. 

Due to these blockers, I am currently putting this PR on hold and we should circle back on it after 6 months or so to see if things are in better place.

.github/workflows/build.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v2
         with:
-          java-version: '8'
+          java-version: '11.0.15'
```
snopoke, line 51: what if we just make this change - so that we are building with v11 but still targeting v8? Then we could upgrade other infra to run v11 too.

shubham1g5: @snopoke That would mean we set `sourceCompatibility` to Java 11 while use the `targetCompatibility` as Java 8 ? I thought about that though it doesn't seem like a [good practice](https://stackoverflow.com/a/22326989/3811963) to have source and target different. Also from [this thread](https://discuss.gradle.org/t/why-cant-i-use-different-sourcecompatibility-targetcompatibility-with-hello-world/11958/2) it seems like these comptibilty settings are used to do the opposite i.e. in our case update infra to Java 11 first and have code still compiles with Java 8 but target Java 11.

snopoke: OK, I didn't know you could do that (compile with v8 but target v11)

shubham1g5:
closing in favour of https://github.com/dimagi/formplayer/pull/1292

PR #1091, opened 2022-02-10
Removed CC version string from app version text

orangejenny:
Removes CC version from web apps footer, just leaving app version. The version shown wasn't accurate, as it was pulling from [here](https://github.com/dimagi/commcare-core/blob/670ff633bf303978f96d8f2e03de205918c34941/src/cli/java/org/commcare/util/engine/CommCareConfigEngine.java#L70-L72). I considered using HQ to populate this, using the latest available CC build, but the current build is typically marked with " (dev)" which seems like a source of confusion/concern.

Screen Shot 2022-02-10 at 12 11 31 PM

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -193,7 +193,7 @@ public class MenuSessionRunnerService {
 
         menuResponseBean.setBreadcrumbs(menuSession.getBreadcrumbs());
         menuResponseBean.setAppId(menuSession.getAppId());
-        menuResponseBean.setAppVersion(menuSession.getCommCareVersionString() + ", App Version: " + menuSession.getAppVersion());
+        menuResponseBean.setAppVersion("App Version: " + menuSession.getAppVersion());
         menuResponseBean.setPersistentCaseTile(getPersistentDetail(menuSession, storageFactory.getPropertyManager().isFuzzySearchEnabled()));
         return menuResponseBean;
     }
@@ -625,7 +625,7 @@ public class MenuSessionRunnerService {
         if (menuSession.getSessionWrapper().getForm() != null) {
             NewFormResponse formResponseBean = generateFormEntrySession(menuSession);
             formResponseBean.setAppId(menuSession.getAppId());
-            formResponseBean.setAppVersion(menuSession.getCommCareVersionString() + ", App Version: " + menuSession.getAppVersion());
+            formResponseBean.setAppVersion("App Version: " + menuSession.getAppVersion());
             formResponseBean.setPersistentCaseTile(getPersistentDetail(menuSession, storageFactory.getPropertyManager().isFuzzySearchEnabled()));
             formResponseBean.setBreadcrumbs(menuSession.getBreadcrumbs());
             // update datadog/sentry metrics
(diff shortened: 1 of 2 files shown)


ctsims:
Hi Jenny,

Is the reason that the static versions listed are out of date because of the internal Formplayer branch forking? The [config engine platform version #](https://github.com/dimagi/commcare-core/blob/670ff633bf303978f96d8f2e03de205918c34941/src/cli/java/org/commcare/util/engine/CommCareConfigEngine.java#L70-L72) is the only non-APK tag in code that I'm aware of that provides a version number for the runtime, which is used semantically when we evaluate whether a profile.ccpr file is compatible with the engine, so it should be canonical and reliable, and we should improve the process to keep it updated if it's drifting.

orangejenny:
@ctsims I'm less concerned about the forking and more about the minor version. Web Apps currently references 2.50 since that was the last time we merged the cc-core master branch into the formplayer branch, but it also uses all of the code that's been added since 2.50 was released.

On the HQ side, recent features like data registries are limited to 2.53+, since that's the CC version the core code is in, which makes it confusing that web apps says it's only running 2.50. (This PR stems from a project team question about what version their web apps app should use.)

I think it's most accurate to say formplayer is running 2.53, even though the master branch of 2.53 hasn't been back merged. That implies we should update the CommCareConfigEngine constants in the formplayer branch when a new dev release is cut, and they should reference the dev version. Does that sound right?

ctsims:
Is there a reason the straightforward approach here isn't just to do a new merge from cc-core into formplayer, bringing the version code up to date? Formplayer can stay branched longer because code in Formplayer takes longer to get generalized to the rollout path for master, but I thought that getting the master branch merged back into Formplayer was a more straightforward operation that we could do regularly? 

I'd be worried about any other path where we try to ensure compatibility with Formplayer's code and whatever functionality we think is 2.53 dependent.

shubham1g5:
That sounds right. I have added a PR here to update FP - https://github.com/dimagi/commcare-core/pull/1072

orangejenny:
@shubham1g5 When do you think is the right time to merge 2.53? I'm glad 2.52 is getting merged, but it still seems confusing that web apps will be labeled 2.52 but it allows for 2.53 functionality.

shubham1g5:
@orangejenny yes it is a bit confusing.  2.53 is actually not freezed for mobile and traditionally we have been versioning the FP by the last CC release out. This merge potentially contains some of 2.53 CC code but not the potential future work that will be included in CC 2.53 when we release it for mobile. But It does make some sense to me to change the version in this merge to 2.53 directly since it does contain some of the 2.53 code. That would mean Formplayer running a version earlier than mobile in general. @ctsims  Wondering if you have thoughts on it ?

ctsims:
@shubham1g5 I'm a little confused on one point. I don't think we should be merging in the 2.53 branch into formplayer, I think we should be merging master into formplayer.

I thought that after we cut a release (IE: We cut Android 2.53.0), part of that process was updating all of the internal codes to refer to the **next** release. So after 2.53.0 is cut, on `master` the version codes for Android would be `2.54.development`. I think that would be the same on Formplayer if we're going to continue supporting pre-release functionality.

shubham1g5:
@ctsims The current [master version](https://github.com/dimagi/commcare-android/blob/master/app/AndroidManifest.xml#L6)  is 2.53 on cc-android. 2.52 was the last release we did. I am indeed merging `master` but want to confirm whether it's ok to version 2.53 on Web Apps given there may be more potential changes on mobile side in 2.53

orangejenny:
Closing this out based on conversation above about merging master into the formplayer branch a bit earlier.

PR #1093, opened 2022-02-11
Updates Style and re-format code

shubham1g5:
- Tries to sync the CommCare Coding Style with checkstyle.xml to minimize manual formatting
- Formats all java code using the Android Studio auto format
- Make some manual formatting changes


shubham1g5:
There are a couple stickler failures I am not able to get to the root of. 

> Wrong order for 'com.fasterxml.jackson.databind.ObjectMapper' import.

the [import](https://github.com/dimagi/formplayer/blob/styleAndFormat/src/main/java/org/commcare/formplayer/session/FormSession.java#L6) order seems right to me for the [check](https://github.com/dimagi/formplayer/blob/styleAndFormat/checkstyle.xml#L266). 

The other one `'array initialization' child has incorrect indentation level 32, expected level should be one of the following: 20, 24, 40, 47.` doesn't make much sense to me either. 

I am inclined to ignore these for now and revisit if we see more of them in future.

snopoke (approved):

PR #1097, opened 2022-02-21
Fix for search again workflows

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1679

Removes `force_manual_action` in the session request data in favour of  defining it per search command as 'force_manual_search' alongside 'execute' - 

````
{
  "query_data": {
    "command_key": {
    	"execute": true,
    	"force_manual_search": true
    }
  }
}
````

This is done so that we can handle 'Search Again' action per case search configuration in parent child module navigations.

### CODE REVIEW

**Comment by orangejenny** [comment]
Skimmed code, looks good. Still need to look more thoroughly at the tests.

**Comment by orangejenny** [comment]
@shubham1g5 Do you think it would be possible to make this backwards compatible so it didn't have to be coupled with the HQ side? It'd be straightforward for the HQ code to provide the flag both on the request and in queryData.

**Comment by shubham1g5** [comment]
@orangejenny I was earlier thinking if we deploy FP a week after HQ changes with HQ sending both flags, we would not need to make FP backward compatible, though since it was not too hard I added backward compatibility [here](https://github.com/dimagi/formplayer/pull/1097/commits/43e01954748cff98b4995a6dff736f944aca82b7) as well.

File: src/test/java/org/commcare/formplayer/tests/FormEntryWithQueryTests.java [hunk 1]
```diff
@@ -61,7 +61,6 @@ public void testFormEntryWithQuery() throws Exception {
         sessionNavigateWithQuery(new String[]{"1"},
                 "caseclaimquery",
                 null,
-                false,
```
orangejenny, line 64: It is rather nice seeing all these positional boolean parameters disappear.

src/test/java/org/commcare/formplayer/tests/SmartLinkTests.java
```diff
@@ -0,0 +1,100 @@
...
+import org.springframework.test.context.ContextConfiguration;
+
+import java.util.Hashtable;
+
+/**
+ * Tests for smart link workflow
+ */
+@WebMvcTest
+@ContextConfiguration(classes = TestContext.class)
+public class SmartLinkTests extends BaseTestClass {
```
orangejenny, line 26: Glad to have tests for this!

orangejenny (approved):

PR #1100, opened 2022-03-17
include queryKey in EntityList responses

snopoke:
This allows Web Apps to know what the current command is which is
necessary when sending back query data.

This is required after the changes pass `force_manual_action` in the query data:
* https://github.com/dimagi/commcare-hq/pull/31152
* https://github.com/dimagi/formplayer/pull/1097

Related HQ change: https://github.com/dimagi/commcare-hq/pull/31273

Note: must be deployed before or at the same time as the CommCare PR


shubham1g5 (approved):

orangejenny (approved):
Thoughts on adding this to `MenuBean` instead?

snopoke:
> Thoughts on adding this to `MenuBean` instead?

Certainly could do and probably makes sense. I'll update

orangejenny (approved):

PR #1101, opened 2022-03-17
Add and use forSubmission parameter to SessionRequestBean

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-13204

As the ticket describes, the issue is that form sessions that expire and submit partially completed forms do not respect relevancy. This is due to recent serialization changes, and despite inspecting this specific scenario [here](https://github.com/dimagi/formplayer/pull/1083#issuecomment-1018949957), I failed to take into consideration _all_ use cases of this request in HQ. If instance xml is being retrieved specifically for a submission, formplayer needs to be aware of this so it can properly serialize the xml.

A followup change will be made on HQ to take advantage of this new parameter when requesting an instance for submitting a partially completed form.


gherceg:
Cross-posting a question I asked in the HQ PR here:
> I am now wondering if it is helpful or adds to confusion to name this paramter for_submission vs something like serialize_all_data or respsect_relevancy.

ctsims (approved):

PR #1105, opened 2022-03-22
Sends multi select case list Params to web apps

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1773

Steps 1-2 in [Spec](https://docs.google.com/document/d/1N_Zg1vqRQofSsJo7X0Bj4WHktTroVofkpUIWt8_3Bhc/edit)

https://github.com/dimagi/commcare-core/pull/1077


snopoke (approved):

shubham1g5:
Had to bump the DB version because of model change in https://github.com/dimagi/commcare-core/pull/1079 , @snopoke  would that require us to do anything special before deploy like priming restore ?
FYI: @gherceg

snopoke:
> Had to bump the DB version because of model change in [dimagi/commcare-core#1079](https://github.com/dimagi/commcare-core/pull/1079) , @snopoke would that require us to do anything special before deploy like priming restore ? FYI: @gherceg

Hmm, that means that all the DB's will need to be rebuilt. I'm not sure if we have a standard procedure for this but we should definitely consider the impact. Priming them is likely the best option but it would need to be timed with the deploy. Let's take this off the PR to figure out.

gherceg:
@shubham1g5 I'm confused on how that PR caused a model change? `uiVersion` and `formVersion` were already properties on the `FormInstance` object.

shubham1g5:
@gherceg It breaks the deserialization from the stream that was serialized using the old code. So suppose a formInstance was stored as a byte stream using the serialised method from the old code. Now if we try to deserialise that byte stream after deploy using the new code, it will break when we try to read the `formVersion` since that was not written. In this case, FP discards the complete old user DB and refetches all resources again from HQ as a source of truth.

shubham1g5:
Updated core ref to include the [revert](https://github.com/dimagi/commcare-core/pull/1081) and reverted the DB version bump.

PR #1107, opened 2022-03-25
formplayer changes related to list query data support

snopoke:
Cross request from https://github.com/dimagi/commcare-core/pull/1078

There are no function changes here and no support for the list query data type either since that is blocked by the code to produce the 'selected-cases' instance.

### DIFF
--- a/src/main/java/org/commcare/formplayer/util/PrototypeUtils.java
+++ b/src/main/java/org/commcare/formplayer/util/PrototypeUtils.java
@@ -21,10 +21,12 @@ import org.commcare.suite.model.ComputedDatum;
 import org.commcare.suite.model.EntityDatum;
 import org.commcare.suite.model.FormEntry;
 import org.commcare.suite.model.FormIdDatum;
+import org.commcare.suite.model.ListQueryData;
 import org.commcare.suite.model.PropertySetter;
 import org.commcare.suite.model.RemoteQueryDatum;
 import org.commcare.suite.model.RemoteRequestEntry;
 import org.commcare.suite.model.Text;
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/formplayer/util/PrototypeUtils.java
```diff
@@ -138,6 +140,8 @@ public static void setupThreadLocalPrototypes() {
                 BubbleSeries.class.getName(),
                 SendAction.class.getName(),
                 XPathDistinctValuesFunc.class.getName(),
+                ValueQueryData.class.getName(),
+                ListQueryData.class.getName(),
```
snopoke, line 144: Is this right? I wasn't sure but it seems to work.

orangejenny: I don't really know how this works. If this change is removed, does something break?

orangejenny (approved):
I'm not 100% clear on the PrototypeUtils change, but I also don't think it's risky.

src/main/java/org/commcare/formplayer/util/PrototypeUtils.java
shubham1g5: yes, looks good. Serialization engine needs all implementations of `Externalizable` to be added here to be able to [instantiate classes](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/util/externalizable/PrototypeFactory.java#L135) correctly while deserializing.

shubham1g5 (approved):

PR #1109, opened 2022-03-30
Session navigation for Multi Select Case List 

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1774
https://dimagi-dev.atlassian.net/browse/USH-1786

Implements [spec](https://docs.google.com/document/d/1N_Zg1vqRQofSsJo7X0Bj4WHktTroVofkpUIWt8_3Bhc/edit#heading=h.myn5i3g8ojch)

[wiki](https://github.com/dimagi/formplayer/wiki/Multi-Select-Case-Lists)

cross-request: https://github.com/dimagi/commcare-core/pull/1082


src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -65,7 +66,7 @@
     public void rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper caseSearchHelper) throws CommCareSessionException, RemoteInstanceFetcher.RemoteInstanceException {
         Vector<StackFrameStep> steps = menuSession.getSessionWrapper().getFrame().getSteps();
         menuSession.resetSession();
-        Screen screen = menuSession.getNextScreen(false, null);
+        Screen screen = menuSession.getNextScreen(false);
```
orangejenny, line 69: This whole commit makes the previous commit much nicer.

orangejenny (approved):

src/main/java/org/commcare/formplayer/database/models/EntitiesSelectionStorage.java
```diff
@@ -0,0 +1,85 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.HashMap;
+
+/**
+ * A simple sql backed cache that stores the selectons on a multi select Entity list into user DB
+ */
+public class EntitiesSelectionStorage implements EntitiesSelectionCache {
```
snopoke, line 20: can you articulate what made you decide to use sqlite for this vs PostgreSQL?

shubham1g5: I was quite confused to be honest on where should this go and decided to go with Sqlite to have more consistency between mobile and Formplayer architecture wrt this table. That being said, I am curious to hear if there are reasons because of which we should be considering `Postgres` as the primary datastore for FP instead of Sqlite ?

snopoke: I think there are a few things to consider:
1. This further entrenches us in the 'db per machine' world
2. Having it in PG would make it possible to clear old data on a schedule (similar to what we do already with form sessions)

shubham1g5: Interesting. That's good to know about and makes sense in general. Though in context of this table I still think having ability to use Sqlite on both mobile and FP give us better code sharing opportunities between the two platforms. And since we have a bunch of other Sqlite tables It doesn't seem like adding a simple key value cache for temporal selections should introduce any additional bottlenecks if we are to make a transition at some point. 

Also we can potentially use [same cron task ](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/ScheduledTasks.java#L33)to delete the entries from Sqlite table as well. Even if that's not possible, I don't imagine having a different schedule for purging this table should cause an issue.

Let me know if you feel strongly about PG though, it should be an easy change to make.

snopoke: > Also we can potentially use [same cron task ](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/ScheduledTasks.java#L33)to delete the entries from Sqlite table as well.

The challenge with this is that it's not a single table so we'd need to query every sqlite DB on the machine individually which is much trickier.

> Though in context of this table I still think having ability to use Sqlite on both mobile and FP give us better code sharing opportunities between the two platforms.

I don't think this is a big issue, and already you've put this code in formplayer repo so it's already not shared.

In general I think the main advantage of the sqlite DBs is that we have very fast IO which is essential for doing the case DB querie. For use cases like this we don't need that since we're just loading a blob of data (and we can cache it in redis if we want).

I'd be curious if @ctsims has any thoughts.

ctsims: Right now I think the split in responsibility is that PG is managing volatile session data (menu sessions, incomplete forms, form session data, etc), and that the sqlite dbs are storing reference data. I think that makes Postgres the most congruous place to store it, since it'll make devops tasks and management the most consistent. If for instance, a sqlite bearing machine is removed from the cluster: Currently a user's form session is retained, and a they continue an existing form session once their newly provisioned sqlite bearing machine syncs their reference data. Putting the data in Postgres would enable the same with this type of data, in addition to the regular cleaning that Simon mentioned, taking advantage of an existing mechanism (the incomplete form clearing db task on startup) without needing a cron-like task.

shubham1g5: makes sense. Thanks for the inputs @ctsims and @snopoke!

shubham1g5: implemented changes in https://github.com/dimagi/formplayer/pull/1118

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -156,17 +170,23 @@ public boolean handleInput(String input, boolean needsDetail, boolean inputValid
                 EntityScreen entityScreen = (EntityScreen)screen;
                 boolean autoLaunch = entityScreen.getAutoLaunchAction() != null && allowAutoLaunch;
                 addBreadcrumb = !autoLaunch;
-                if (input.startsWith("action ") || (autoLaunch) || !inputValidated) {
+                if (input.startsWith("action ") || (autoLaunch) || !inputValidated ||
+                        input.contentEquals(USE_SELECTED_VALUES)) {
```
snopoke, line 174: Is this necessary? I think if we ensure that this value is never saved in the `cacheSessionSelections` then it shouldn't be.

I also think that we currently are saving USE_SELECTED_VALUES into the cached session selections. I think line 317 below should change:

```diff
- restoreFactory.cacheSessionSelections(selections);
+ restoreFactory.cacheSessionSelections(menuSession.getSelections());
```

I'd be interested to see without that change whether `inputValidated` is correctly set for navigation after the selections are modified. I suspect it is not since we are not caching the GUID in the session selections.

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -302,9 +355,9 @@ private void computeDatum() {
         EvaluationContext ec = sessionWrapper.getEvaluationContext();
         if (datum instanceof FormIdDatum) {
```
snopoke, line 356: aside: What is a `FormIdDatum`? I can't find any such thing in HQ.

src/main/java/org/commcare/formplayer/application/AbstractBaseController.java
```diff
@@ -56,6 +56,9 @@ public abstract class AbstractBaseController {
     @Autowired
     private NotificationLogger notificationLogger;
 
+    @Autowired
+    private VirtualDataInstanceService virtualDataInstanceService;
```
snopoke, line 60:
```suggestion
    private VirtualDataInstanceCache virtualDataInstanceService;
```

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -51,6 +52,9 @@ public class MenuSessionFactory {
     @Autowired
     protected FormplayerStorageFactory storageFactory;
 
+    @Autowired
+    private VirtualDataInstanceService virtualDataInstanceService;
```
snopoke, line 56:
```suggestion
    private VirtualDataInstanceCache virtualDataInstanceService;
```

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
```diff
@@ -51,6 +51,9 @@ public class NewFormResponseFactory {
     @Autowired
     private FormplayerStorageFactory storageFactory;
 
+    @Autowired
+    private VirtualDataInstanceService virtualDataInstanceService;
```
snopoke, line 55:
```suggestion
    private VirtualDataInstanceCache virtualDataInstanceService;
```

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
```diff
@@ -25,20 +29,37 @@ public class ScheduledTasks {
     @Autowired
     private FormSessionService formSessionService;
 
+    @Autowired
+    private VirtualDataInstanceService virtualDataInstanceService;
```
snopoke, line 33:
```suggestion
    private VirtualDataInstanceCache virtualDataInstanceService;
```

src/main/java/org/commcare/formplayer/services/VirtualDataInstanceService.java
```diff
@@ -0,0 +1,79 @@
...
+        SerializableDataInstance serializableDataInstance = cache.get(key, SerializableDataInstance.class);
+        if (serializableDataInstance == null) {
+            Optional<SerializableDataInstance> optionalSerializableDataInstance = dataInstanceRepo.findById(key);
+            if (optionalSerializableDataInstance.isPresent()) {
+                serializableDataInstance = optionalSerializableDataInstance.get();
+            }
+        }
+        if (serializableDataInstance != null
+                && ifUsernameMatches(serializableDataInstance.getUsername(), storageFactory.getUsername())
+                && serializableDataInstance.getAppId().equals(storageFactory.getAppId())) {
```
snopoke, line 60: nit: extract this to a function

In fact you could extract the whole if condition to a function like `validateInstance` which throws an `InstanceNotFoundException` if any condition fails.

src/main/java/org/commcare/formplayer/session/MenuSession.java
shubham1g5: that's a good idea, agree that it's not caching the guid selection today in case wheh the selection array have `USE_SELECTED_VALUES`

src/main/java/org/commcare/formplayer/session/MenuSession.java
shubham1g5: I am not aware of it's use case, probably a legacy thing @ctsims  ?

ctsims: This datum lets you configure an XMLNS dynamically based on session data. I think it's a legacy thing that we haven't used for a long time, and I'm not sure if HQ ever let you configure it, it might have been a remote-apps only thing.

src/main/java/org/commcare/formplayer/session/MenuSession.java
shubham1g5: https://github.com/dimagi/formplayer/pull/1109/commits/bd2469675698752c88e4fc18a2286456ea5b0932

src/main/java/org/commcare/formplayer/services/VirtualDataInstanceService.java
shubham1g5: https://github.com/dimagi/formplayer/pull/1109/commits/020d90b81b133f8ed7d87ae0750358e1517db95d

snopoke (approved):
Some small comments but nothing blocking

shubham1g5:
merging this PR and would followup on pending discussions in separate PRs.

src/main/java/org/commcare/formplayer/services/VirtualDataInstanceService.java
```diff
@@ -0,0 +1,90 @@
...
+                            serializableDataInstance.getReference(), serializableDataInstance.useCaseTemplate(),
+                            key);
+            return instanceSource.toInstance();
+        }
+        throw new InstanceNotFoundException(key);
+    }
+
+    private boolean validateInstance(SerializableDataInstance serializableDataInstance, String key) {
+        if (serializableDataInstance != null
+                && ifUsernameMatches(serializableDataInstance.getUsername(), storageFactory.getUsername())
```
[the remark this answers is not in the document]
shubham1g5: Is that required given appids are unique across domains ?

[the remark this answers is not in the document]
snopoke: probably not, especially since this change https://github.com/dimagi/formplayer/pull/1141/commits/dffd8e3c6d9ff4e6c079ba9835a00874d1206de5

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -187,28 +204,55 @@ public boolean handleInput(String input, boolean needsDetail, boolean inputValid
...
     }
 
     private void addTitle(String input, Screen previousScreen) {
-        if (previousScreen instanceof EntityScreen) {
-            try {
-                String caseName = SessionUtils.tryLoadCaseName(sandbox.getCaseStorage(), input);
-                if (caseName != null) {
-                    breadcrumbs.add(caseName);
-                    return;
+        if (previousScreen instanceof MultiSelectEntityScreen) {
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/formplayer/pull/1148/commits/2a7b09aa8347b766b5f2dc52a4816a5fac5844f3

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -295,7 +298,9 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
                 // we don't have a resolution, try rebuilding session to execute any pending ops
                 executeAndRebuildSession(menuSession);
             } else {
-                menuSession.addSelection(selection);
+                if (!selection.contentEquals(MultiSelectEntityScreen.USE_SELECTED_VALUES)) {
+                    menuSession.addSelection(selection);
+                }
```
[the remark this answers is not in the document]
shubham1g5: this worked out great, https://github.com/dimagi/formplayer/pull/1148/commits/ceb34d315553be61dfc76708db303f3101da60cd

PR #1112, opened 2022-04-01
Initializes Selected cases instance For Multi Select Case Lists

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1775

[Spec](https://docs.google.com/document/d/1N_Zg1vqRQofSsJo7X0Bj4WHktTroVofkpUIWt8_3Bhc/edit)

cross-request:https://github.com/dimagi/commcare-core/pull/1084

PR #1115, opened 2022-04-06
Tests for case claim with a multi select entity list

shubham1g5:
Tests for [case claim changes](https://docs.google.com/document/d/1N_Zg1vqRQofSsJo7X0Bj4WHktTroVofkpUIWt8_3Bhc/edit#heading=h.hfyfjlf661id) with a multi-select entity list


src/test/resources/archives/case_claim_with_multi_select/en/app_strings.txt
```diff
@@ -0,0 +1,22 @@
+app.display.name=Case Claim
```
snopoke, line 1: do we need both the `default` and the `en` app strings?

src/test/resources/archives/case_claim_with_multi_select/modules-0/forms-0.xml
```diff
@@ -0,0 +1,120 @@
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
```
snopoke, line 1: since we don't care about the form content for these test it would be nice to simplify these to the most basic form

src/test/java/org/commcare/formplayer/tests/MultiSelectCaseClaimTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private void configureSyncMock() {
+        when(webClientMock.post(anyString(), any()))
+                .thenReturn("");
+    }
+
+    private void configureQueryMock() {
+        when(webClientMock.postFormData(anyString(), any(Multimap.class)))
+                .thenReturn(FileUtils.getFile(this.getClass(),
+                        "query_responses/case_claim_multi_select_response.xml"));
```
snopoke, line 106: It's generally better practice to configure the mock to do the validation instead of doing it with captors. There's an example of that here: https://github.com/dimagi/formplayer/blob/a98a768874b9c32911566e61a048ad5d3f519318/src/test/java/org/commcare/formplayer/tests/EndOfFormNavFormLinkingWithQueryTests.java#L59-L72

Also the case claim endpoint doesn't respond with any content, just a status code of 201.

src/test/java/org/commcare/formplayer/tests/MultiSelectCaseClaimTest.java
```diff
@@ -0,0 +1,113 @@
...
+        assertEquals(requestData.get("case_id"), casesToBeClaimed);
+    }
+
+    private void configureSyncMock() {
+        when(webClientMock.post(anyString(), any()))
+                .thenReturn("");
+    }
+
+    private void configureQueryMock() {
+        when(webClientMock.postFormData(anyString(), any(Multimap.class)))
```
snopoke, line 104: I also notice that this is mocking the `postFormData` call but I think the claim request just uses `post`.

Looking more at the test I think you've got the naming of the mock methods swapped. This one should be `configureSyncMock`.

shubham1g5: `configureQueryMock` is supposed to mock the case search request which is of type `postFormData`. `configureSyncMock` is for case claim which is as you said `post`.

snopoke: I see, so the names are derived from the names of the 'screen' classes. I don't find the names line up very well with what they do so I'd recommend a rename of these functions:

* `configureSyncMock` -> `configureCaseClaimMockRequest`
* `configureQueryMock` -> `configureCaseSearchMockRequest`

orangejenny: +1 for renaming "sync" to either "claim" or "post"

src/test/java/org/commcare/formplayer/tests/MultiSelectCaseClaimTest.java
```diff
@@ -102,4 +102,9 @@ private void configureQueryMock() {
                 .thenReturn(FileUtils.getFile(this.getClass(),
                         "query_responses/case_claim_multi_select_response.xml"));
     }
+
+    @Override
+    protected boolean useCommCareArchiveReference() {
```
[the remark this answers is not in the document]
shubham1g5: no objections though this was [recently implemented](https://github.com/dimagi/formplayer/pull/1112/commits/575d7a7490c3c91b9b8e9ffb3133cc2a584228d4) and is not yet in `master`.  SInce multiselect work will take a while to merge, I can do a separate PR for this if you need these changes.

src/test/java/org/commcare/formplayer/tests/MultiSelectCaseClaimTest.java
shubham1g5: I looked into this though it turned out it would not be a strict enough test for the claim request. The reason being even if we don't mock the case claim request the test will pass without any failure as we don't really use the return values from the [claim request](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L486) unlike the case query request. So configuring the mock behaviour won't guarantee that the webClient was fired with those specific arguments only and not some other arguments unless we also configure the mock to throw an exception when the arguments do not match. 

I also like the argument capture style in this case since it's more assertive. This test's sole purpose is to check for the correct claim request and making assertions on argument conveys that much better.

src/test/java/org/commcare/formplayer/tests/MultiSelectCaseClaimTest.java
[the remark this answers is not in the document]
orangejenny: nah, don't worry about it. We can do it on an as-needed basis as we update tests, or I'll work on it some late Friday afternoon.

src/test/java/org/commcare/formplayer/tests/MultiSelectCaseClaimTest.java
```diff
@@ -0,0 +1,107 @@
...
+        // cases that are owned should not be in the claim request
+        List<String> casesToBeClaimed = Arrays.asList("0156fa3e-093e-4136-b95c-01b13dae66c7",
+                "0156fa3e-093e-4136-b95c-01b13dae66c8");
+        assertEquals(requestData.get("case_id"), casesToBeClaimed);
+    }
+
+    private void configureCaseSearchMock() {
+        when(webClientMock.postFormData(anyString(), any(Multimap.class)))
+                .thenReturn(FileUtils.getFile(this.getClass(),
+                        "query_responses/case_claim_multi_select_response.xml"));
```
snopoke, line 100: nit: rename to match request
```suggestion
                        "query_responses/case_search_multi_select_response.xml"));
```

src/test/resources/archives/case_claim_with_multi_select/suite.xml
```diff
@@ -0,0 +1,313 @@
...
+        <locale id="forms.m0f0"/>
+      </text>
+    </command>
+    <instance id="casedb" src="jr://instance/casedb"/>
+    <instance id="commcaresession" src="jr://instance/session"/>
+   <session>
+      <datum id="case_id_new_case_0" function="uuid()"/>
+      <query url="http://www.example.com/a/domain/phone/get_case/" storage-instance="registry" template="case" default_search="true">
+        <data key="case_type" ref="'case'"/>
+      </query>
```
snopoke, line 197: It would be nice to remove elements that are not needed for the test to make it easier to see what's going on. This query is for data registry workflows which is not part of this at all. The entry stack frames are not needed either.

snopoke (approved):

PR #1117, opened 2022-04-08
Dont sync on 204 code from case claim request

AddisonDunn:
[Jira ticket](https://dimagi-dev.atlassian.net/browse/USH-1833)

Under the [case search -> case claim](https://confluence.dimagi.com/display/saas/Case+Search+and+Claim) workflow. 

Small change. When formplayer sends a request to HQ to claim a case, HQ will return a 204 status if no cases were claimed (see above ticket for the previous work where this was implemented). If this is the case, we don't need to sync the user's restore file. This code implements the "not syncing if we get back a 204" logic.

Tested locally thus far. From what I can tell, no automated tests to really work off of.

I will try to make these changes for commcare-core as well in another PR.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/FormattedQuestionsService.java
+++ b/src/main/java/org/commcare/formplayer/services/FormattedQuestionsService.java
@@ -43,7 +43,7 @@ public class FormattedQuestionsService {
         body.add("xmlns", xmlns);
         body.add("appId", appId);
 
-        String responseBody = webclient.post(getFormattedQuestionsUrl(host, domain), body);
+        String responseBody = (String) webclient.post(getFormattedQuestionsUrl(host, domain), body).getBody();
         JSONObject responseJSON = new JSONObject(responseBody);
         return new QuestionResponse(
                 responseJSON.getString("form_data"),
--- a/src/main/java/org/commcare/formplayer/services/FormattedQuestionsService.java
+++ b/src/main/java/org/commcare/formplayer/services/FormattedQuestionsService.java
@@ -43,7 +43,7 @@ public class FormattedQuestionsService {
         body.add("xmlns", xmlns);
         body.add("appId", appId);
 
-        String responseBody = (String) webclient.post(getFormattedQuestionsUrl(host, domain), body).getBody();
+        String responseBody = webclient.post(getFormattedQuestionsUrl(host, domain), body);
         JSONObject responseJSON = new JSONObject(responseBody);
         return new QuestionResponse(
                 responseJSON.getString("form_data"),
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -482,8 +482,9 @@ public class MenuSessionRunnerService {
     }
 
     private NotificationMessage doSync(FormplayerSyncScreen screen) throws Exception {
+        Boolean shouldSync = true;
         try {
-            webClient.post(screen.getUrl(), screen.getQueryParams());
+            shouldSync = webClient.caseClaimPost(screen.getUrl(), screen.getQueryParams());
         } catch (RestClientResponseException e) {
             return new NotificationMessage(
                     String.format("Case claim failed. Message: %s", e.getResponseBodyAsString()), true,
@@ -492,7 +493,9 @@ public class MenuSessionRunnerService {
             return new NotificationMessage("Unknown error performing case claim", true,
                     NotificationMessage.Tag.sync);
         }
-        restoreFactory.performTimedSync(false, false, false);
+        if (shouldSync) {
+            restoreFactory.performTimedSync(false, false, false);
+        }
         return null;
     }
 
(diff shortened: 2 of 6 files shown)


ctsims:
Hi Addison - Can you clarify the spec behavior for this process a bit?

Specifically is the 204 response triggered based on the set of cases based on the user's sync token, or based on the user's "sandbox" state (IE: the set of cases they would receive in a fresh restore)? 

If it's the latter I could see a potential failure mode here where a user is using two devices (say, a phone and a web apps browser) and on one of them tries to 'claim' a case that's currently in their sandbox based on actions which they haven't synced locally yet. In that case the sever could reply with a 204 (since there's already a claim case owned by the user), but without a sync that claim case might not come down. 

Does that concern make sense?

AddisonDunn:
@ctsims That seems like a very valid concern. I haven't reviewed the code thoroughly on the HQ side enough to know but I think @dominicochoa-dimagi or @orangejenny would be able to answer. For reference, I think the relevant updates were made in [this PR](https://github.com/dimagi/commcare-hq/pull/31304).

dominicochoa-dimagi:
@ctsims @AddisonDunn 204 is returned when no _new_ cases are claimed. Meaning, if someone tries to only claim cases they already own, we return a 204. 

If someone tries to claim a combination of new cases and already claimed cases, or **only** new cases, we return a 201.

snopoke:
@ctsims makes a very valid point. I don't think there is a way we can safely skip the sync without a check to verify that the cases are actually in the local case DB.

Currently the 'post' request is a generic XML structure that could contain any data. If we want to be able to perform a check of cases locally we'd need to make formplayer aware that data being sent is case IDs. I don't particularly like the idea of doing that but I also don't see a specific issue with it.

orangejenny:
Argh, yes, @ctsims you're totally right, thank you for raising.

@snopoke I don't love the idea of making formplayer more aware of what the post request is, either. I'm inclined to close this PR. I think this could have been a nice little improvement but it isn't worth making the code more complex / less generic since it isn't a known, significant performance problem. At this point this is a USH-only feature, and I don't think we're expecting sync times to be nearly as bad for this upcoming project as they have been for the CICT projects.

snopoke (commented):
I'm going to make a general comment about this code even though the PR is going to be closed for the sake of learning.

Rather than changing the return type of `WebClient.post` I would have recommended a new function that is more specific to case claim which will perform the request and interpret the result (returning a boolean).

ctsims:
One other idea here: The mobile can send over the Sync Token ID (it may already)  or the Sync State hash (which just verifies that all of the Case ID's present in the sandbox are the same) along with the sync request, which means HQ may be able to inherently identify if the local sync state on its side to determine the 204 response without needing any additional context from mobile.

snopoke:
> One other idea here: The mobile can send over the Sync Token ID (it may already) or the Sync State hash (which just verifies that all of the Case ID's present in the sandbox are the same) along with the sync request, which means HQ may be able to inherently identify if the local sync state on its side to determine the 204 response without needing any additional context from mobile.

We could definitely do this. @orangejenny if we want to continue with this we can split the work out for the HQ side and keep the formplayer piece on hold until that has been merged.

orangejenny:
@snopoke I'm not very familiar with the sync tokens. Does this mean that formplayer would send its current sync token (which looks like it's in the request header, based on `SyncTokenMiddleware`, though I'm not sure if it's in all formplayer requests), HQ would compare that to its own token (the user's most recent SyncLogSQL entry), and it would only send the 204 if no cases are claimed **and** those two tokens are the same?

snopoke:
> @snopoke I'm not very familiar with the sync tokens. Does this mean that formplayer would send its current sync token (which looks like it's in the request header, based on `SyncTokenMiddleware`, though I'm not sure if it's in all formplayer requests),

Formplayer does always send the sync token if it's available (https://github.com/dimagi/formplayer/blob/dffa67cb818c8474eed05ec6983c8a0ad7483198/src/main/java/org/commcare/formplayer/services/RestoreFactory.java#L604). 

>  HQ would compare that to its own token (the user's most recent SyncLogSQL entry), and it would only send the 204 if no cases are claimed **and** those two tokens are the same?

Not quite. HQ would fetch the SyncLogSQL and check that all claimed cases are present in the synclog:
```python
synclog = get_properly_wrapped_sync_log(request.last_sync_token)

all_cases_exist_on_phone = all(synclog.phone_is_holding_case(case_id) for case_id in claimed_case_ids)
should_sync = not all_cases_exist_on_phone

# Alternatively
missing_case_ids_on_phone = set(claimed_case_ids) - synclog.case_ids_on_phone
should_sync = bool(missing_case_ids_on_phone)
```

orangejenny:
@snopoke ohhh, I see now, thank you. And that seems quite doable. @AddisonDunn what do you think, would you be interested in owning the HQ change or would you rather stick to the Java side of this one?

AddisonDunn:
That makes sense to me too! @orangejenny happy to handle.

To make sure I've got the right idea, we want to:

1. Keep the code in this PR (or at least the functionality to not sync on 204)
2. Implement this for commcare-core
3. In HQ, change the way we detect whether a user needs to sync or not, by using the sync token

orangejenny:
@AddisonDunn Yep, exactly. And the HQ PR will need to be deployed before it's safe to deploy the changes to formplayer and commcare-core.

AddisonDunn:
I reset to the original state before this PR then added same functionality but using a new client post method as @snopoke requested.

Won't merge until the [HQ side changes to check the synclog](https://github.com/dimagi/commcare-hq/pull/31458) are deployed.

snopoke (approved):

PR #1118, opened 2022-04-09
Switch Entities selection table to Postgress

shubham1g5:
Context: https://github.com/dimagi/formplayer/pull/1109/files#r840569109


src/main/java/org/commcare/formplayer/services/EntitiesSelectionService.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.stereotype.Service;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+@Service
+@CacheConfig(cacheNames = {"entities_selection"})
```
shubham1g5, line 25: @snopoke I am not quite sure if we actually need a cache on top of the postgres table. Do you have an opinion on it ?

src/main/java/org/commcare/formplayer/db/migration/V22__init_entities_selection.java
```diff
@@ -0,0 +1,19 @@
+package org.commcare.formplayer.db.migration;
+
+import java.util.Arrays;
+
+public class V22__init_entities_selection extends BaseFormplayerMigration {
```
shubham1g5, line 5: @gherceg I am unsure how to best manage this migration without collision with https://github.com/dimagi/formplayer/pull/1075. There are a few challenges here - 
1.  The version number here depends on what PR will go out first. 
2. I want to put this on staging though staging already has '22' as schema version. 

It seems like best course of action here is for one of these PRs to rebase on top of the other. @gherceg Is there a possibility to depriortize roll out of Form def work so that we roll out this PR first (I think it can be 2-4 weeks before we roll this out) and then for you to base off your branch on top of this ? Or we can do the other way rounds if your changes are more closer to being rolled out. 

FYI: @snopoke

src/main/java/org/commcare/formplayer/db/migration/V22__init_entities_selection.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @Override
+    public Iterable<String> getSqlStatements() {
+        return Arrays.asList(
+                "CREATE TABLE entities_selection (\n" +
+                        "    id text NOT NULL,\n" +
```
snopoke, line 11: Since this is auto-generated and will always be a UUID I think the column should be `UUID` type.

src/main/java/org/commcare/formplayer/services/EntitiesSelectionService.java
```diff
@@ -0,0 +1,75 @@
...
+    @CacheEvict(allEntries = true)
+    public int purge() {
+        Instant cutoff = Instant.now().minus(7, ChronoUnit.DAYS);
+        long start = System.currentTimeMillis();
+        int deletedRows = entitiesSelectionRepo.deleteSessionsOlderThan(cutoff);
+        long elapsed = System.currentTimeMillis() - start;
+        log.info(String.format("Purged %d stale entities selections in %d ms", deletedRows, elapsed));
+        if (datadogStatsDClient != null) {
+            datadogStatsDClient.time("PostgresEntitiesSelectionRepo.purge.timeInMillis", elapsed);
+        }
```
snopoke, line 72: Now that there are 2 purge methods that duplicate a lot of the same code it would be nice to move most of this to the caller so that it's not duplicated and so that the cutoff time is kept the same:

```
doTimedPurge(formSessionService, "deletedRows");
doTimedPurge(entitiesSelectionService, "deletedEntitiesSelectionsRows");
```

src/main/java/org/commcare/formplayer/services/EntitiesSelectionService.java
snopoke: I think it makes sense to cache it for the sake of faster access.

src/main/java/org/commcare/formplayer/db/migration/V22__init_entities_selection.java
shubham1g5: makes sense. separately do you think we need to store info like `user` , `domain` and `appid` in the table. I see that we do that for `MenuSession` even if they are not used explicity.

snopoke: Good point. We should probably store some information that allows us to verify that the user has access to the selections. So storing the userid / username should be sufficient.

Storing the domain has no functional use right at the present but it could be useful in the future either functionally or else for maintenance / metrics.

src/main/java/org/commcare/formplayer/db/migration/V22__init_entities_selection.java
orangejenny: (FYI I pulled the https://github.com/dimagi/formplayer/pull/1075 branch off of staging in order to put this PR on staging, after checking with QA that 1075 was okay to remove.)

gherceg: Thanks for flagging @shubham1g5. The shared form definition work is close to complete (final stages of QA), but I should have separated out the migration from the code anyway. Created https://github.com/dimagi/formplayer/pull/1119 to isolate the migration so we can safely merge that and avoid nasty migration conflicts.

[Ramblings] Maybe we can learn from this and do the same with this work as well? I do think having the migration and code together adds value to reviewers and likely results in better feedback, but this could be achieved by branching off the branch with the migration and adding dependent code there, and waiting for reviews before merging the migration. Probably not a bad habit to pickup despite the fact that we don't add formplayer migrations very often.

shubham1g5: Thanks a bunch Graham! That does solve it for us and yeah agree that if possible we should merge the migrations sooner than later.

src/main/java/org/commcare/formplayer/db/migration/V22__init_entities_selection.java
shubham1g5: https://github.com/dimagi/formplayer/pull/1118/commits/39d6a28383256b4aa0172ea52878cec32b19d8e6

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
```diff
@@ -25,20 +29,37 @@ public class ScheduledTasks {
...
+        long elapsed = System.currentTimeMillis() - start;
+        log.info(String.format("Purged %d records in %d ms for %s", deletedRows, elapsed, tag));
         datadogStatsDClient.count(
-                String.format("%s.%s", Constants.SCHEDULED_TASKS_PURGE, "deletedRows"),
+                String.format("%s.%s.%s", Constants.SCHEDULED_TASKS_PURGE, "deletedRows", tag),
                 deletedRows
         );
-        datadogStatsDClient.increment(
-                String.format("%s.%s", Constants.SCHEDULED_TASKS_PURGE, "timesRun")
+        datadogStatsDClient.time(
```
shubham1g5, line 60: this results in change in the Datadog keys from before, is that alright ?

src/main/java/org/commcare/formplayer/services/EntitiesSelectionService.java
```diff
@@ -0,0 +1,72 @@
...
+        EntitiesSelection entitiesSelection = entitiesSelectionRepo.save(entitySelection);
+        Cache cache = cacheManager.getCache("entities_selection");
+        cache.put(entitiesSelection.getId(), entitiesSelection.getEntities());
+        return entitiesSelection.getId();
+    }
+
+    @Nullable
+    @Override
+    @Cacheable
+    public String[] read(UUID key) {
```
snopoke, line 55: this should validate that the values found belong to the current user

src/main/java/org/commcare/formplayer/services/EntitiesSelectionService.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Autowired(required = false)
+    private StatsDClient datadogStatsDClient;
+
+    @Autowired
+    CacheManager cacheManager;
+
+    @Override
+    public UUID write(String[] values) {
+        EntitiesSelection entitySelection =  new EntitiesSelection(storageFactory.getUsername(),
```
shubham1g5, line 44: I am a bit unsure if `storageFactory` is the right object to take the metadata from and have not been able to test this yet. Though hoping to do it soon.

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
snopoke: I think that's OK

src/main/java/org/commcare/formplayer/services/EntitiesSelectionService.java
snopoke: I actually think the restore factory is a better choice at the moment.

shubham1g5: Ahh ok, curious if you see anything wrong with `storageFactory` ? It seems like it exists to provide this data and seems to be more light weight than `restoreFactory`.

snopoke: I think it's fine to use either provided it is available in all the requests where it's needed (which it seems like it should be). Ultimately I want to extract this out into the auth objects.

shubham1g5: Sounds good!

src/main/java/org/commcare/formplayer/services/EntitiesSelectionService.java
shubham1g5: decided to [implement the validation](https://github.com/dimagi/formplayer/pull/1118/commits/318cca9530133616045156802a946476f3ed0beb) against username, appid pair as same user can exist through multiple apps

snopoke (commented):
Just thinking about future work to make the case search inputs available in forms etc (https://dimagi-dev.atlassian.net/browse/USH-1754) I think we could generalise this data store to store any virtual instance.

shubham1g5:
Ahh that's a great point. That would mean storing the serialized instance xml tree in the DB rather than the string array right ? I think that would be significant rework but seems like the right thing to do if we have to do it for case search instance anyway.

snopoke:
> Ahh that's a great point. That would mean storing the serialized instance xml tree in the DB rather than the string array right ? I think that would be significant rework but seems like the right thing to do if we have to do it for case search instance anyway.

Yes, I think it's the way to go. It probably makes sense to store the instance ID as well.

shubham1g5:
Sounds good, Is it ok to merge this PR for now in the parent `multiSelectList` branch and then do the rework on top of that ?

PR #1121, opened 2022-04-13
FormDef Object Pooler

gherceg:
[SAAS-13404](https://dimagi-dev.atlassian.net/browse/SAAS-13404)

This is the initial implementation for a `FormDef` pool that will help boost performance between requests. We cannot rely on simply caching `FormDef` objects, because they must be copied (serialized/deserialized) which is costly. This allows us to avoid the serialization process when possible. The pool has a couple of tunable constraints, max total size and max size per id, which should allow us to keep this from taking up too much space in memory.

This is obviously incomplete on its own, but I wanted to create this PR in isolation of the future work that will come. To provide context, the idea is to add one more layer above the pool that is responsible for interfacing with both the pool and if necessary, the db to fetch the form definition for the given (app id, form xmlns, form version) combo. Once that layer is complete, other parts of formplayer can interact with that layer to obtain form defs when needed.

I've added unit tests for this, though am still concerned/would like to add more tests around the immutability of the root `TreeElement` as this is essential to this functioning properly. The biggest risk is leaking user data from one form session into another by not properly resetting a `FormDef` object.

### DIFF
--- a/src/main/java/org/commcare/formplayer/exceptions/AlreadyExistsInPoolException.java
+++ b/src/main/java/org/commcare/formplayer/exceptions/AlreadyExistsInPoolException.java
@@ -1,0 +1,7 @@
+package org.commcare.formplayer.exceptions;
+
+public class AlreadyExistsInPoolException extends Exception {
+    public AlreadyExistsInPoolException() {
+        super("Attempted to add form def object to pool that has already been added.");
+    }
+}
--- a/src/main/java/org/commcare/formplayer/exceptions/ExceedsMaxPoolSizeException.java
+++ b/src/main/java/org/commcare/formplayer/exceptions/ExceedsMaxPoolSizeException.java
@@ -1,0 +1,7 @@
+package org.commcare.formplayer.exceptions;
+
+public class ExceedsMaxPoolSizeException extends Exception {
+    public ExceedsMaxPoolSizeException(int totalSize) {
+        super("Exceeds max pool size " + totalSize);
+    }
+}
--- a/src/main/java/org/commcare/formplayer/exceptions/ExceedsMaxPoolSizePerId.java
+++ b/src/main/java/org/commcare/formplayer/exceptions/ExceedsMaxPoolSizePerId.java
@@ -1,0 +1,7 @@
+package org.commcare.formplayer.exceptions;
+
+public class ExceedsMaxPoolSizePerId extends Exception {
+    public ExceedsMaxPoolSizePerId(int maxNum, String id) {
+        super("Max number (" + maxNum + ") of entries for id " + id + " exceeded.");
+    }
+}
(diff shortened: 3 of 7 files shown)


snopoke (commented):
Is it necessary to create our own object pool implementation? There are a number of object pools already available:

* https://commons.apache.org/proper/commons-pool/
* https://danielw.cn/fast-object-pool/
* https://chrisvest.github.io/stormpot/
* http://www.vibur.org/vibur-object-pool/

ctsims:
@Graham Herceg ***@***.***>

Do we have performance data on the cost for the pure bytestream
deserialization step of getting a formdef object back?

I think object pooling is *great* as a general approach in Java, but I am
skeptical that figuring out ways to make the in-memory version of the
FormDef object immutable is where the performance savings are going to
come, since we are already round-tripping the entire instance XML into
serialization/deserialization every time we make a request anyway.

It would be good to know specifically which part of the request lifecycle
this pooling is intended to optimize. If, for instance, it's the
deserialization of the object in between requests from the same user /
session it feels like caching the object in memory in between requests (if
available) is a safer play. If it's loading the initial FormDef when a form
session is initiated, I'm skeptical that there's an enormous cost saving to
the deserialization step compared to the actual initialization.

There's also another potential approach we could take here WRT pooling
where instead of re-using objects we keep hot spares cached, with the
creation of new spares deferred as part of the request.

IE: For FormDef Pool P:

   - When form X is requested for the first time: Deserialize X from
   bytestream and return to the request. At the same time, kick of an async
   process to deserialize 5 fresh copies of X to add as hot spares
   - When X is requested later, return a hot spare and kick off an async
   process to create a replacement
   - If the pool is ever empty, Deserialize a fresh copy, and double the
   baseline (5) number of copies that get added to seed the pool
   - Expire X from the pool if no spares have been requested for some window

That process doesn't need to worry about Immutability of objects or
introduce risks of data leakage, and we could tune it as needed over time
to get the ~same steady state performance as a pool without introducing
breakdown costs.

gherceg:
> Do we have performance data on the cost for the pure bytestream
deserialization step of getting a formdef object back?

I do not have any data, but will look into this.

> If, for instance, it's the
deserialization of the object in between requests from the same user /
session it feels like caching the object in memory in between requests (if
available) is a safer play

This is likely where the performance gains would be felt the most (for example, back to back answer requests by the same user in the same session). I previously only considered caching by app id, form xmlns, and form version which would cause issues since multiple sessions could try to access the same mutated form def. In this case, adding the session id to that combination allows safely caching form defs. However, I'm unsure of the behavior of using a form def that has already been mutated/injected with instance xml, and if it would be necessary to bypass that process when using a cached form def.

The hot spares idea is interesting and certainly sounds safer than the typical object pool. I'll look into the above solution first as that appears to have the best value to effort ratio. Thanks for the input @ctsims

calellowitz:
> It would be good to know specifically which part of the request lifecycle this pooling is intended to optimize. If, for instance, it's the deserialization of the object in between requests from the same user / session it feels like caching the object in memory in between requests (if available) is a safer play

Just chiming in to say this is where my previous investigations have shown the largest performance costs (specifically large amounts of time in the `readExternal` method), and just caching that object per session seems like it would solve 90% of the issue. Only starting the session would have to read the definition from the db, all the answers, where the performance penalty really matters because they are otherwise such short requests, could just use the copy in memory. I am not sure why I didn't think of that but this seems like a really nice solution, and probably much more straightforward to implement, in addition to the safety benefits.

PR #1122, opened 2022-04-14
Fixes for multi select Case list

shubham1g5:
Some minor fixes that came up during testing. 

https://github.com/dimagi/commcare-core/pull/1097

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -1,8 +1,9 @@
 package org.commcare.formplayer.services;
 
+import datadog.trace.api.Trace;
 import com.google.common.collect.ImmutableMap;
 import com.timgroup.statsd.StatsDClient;
-
+import io.sentry.SentryLevel;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -11,7 +12,6 @@ import org.commcare.core.parse.ParseUtils;
 import org.commcare.formplayer.api.process.FormRecordProcessorHelper;
 import org.commcare.formplayer.auth.HqAuth;
 import org.commcare.formplayer.beans.AuthenticatedRequestBean;
-import org.commcare.formplayer.beans.InstallRequestBean;
 import org.commcare.formplayer.engine.FormplayerTransactionParserFactory;
 import org.commcare.formplayer.exceptions.AsyncRetryException;
 import org.commcare.formplayer.exceptions.SQLiteRuntimeException;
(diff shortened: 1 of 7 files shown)


src/test/java/org/commcare/formplayer/tests/MultiSelectCaseListTest.java
```diff
@@ -65,6 +68,19 @@ public void testNormalMultiSelectCaseList() throws Exception {
...
+    @Test
+    public void testMultiSelectCaseListWithPreValidatedInput() throws Exception {
+        Mockito.doReturn(true)
+                .when(restoreFactoryMock).isConfirmedSelection(any());
+        String[] selections = new String[]{"0", "1", "use_selected_values"};
+        String[] selectedValues =
+                new String[]{"5e421eb8bf414e03b4871195b869d894", "3512eb7c-7a58-4a95-beda-205eb0d7f163"};
+        NewFormResponse formResp = sessionNavigateWithSelectedValues(selections, APP, selectedValues,
+                NewFormResponse.class);
+        sessionNavigateWithSelectedValues(selections, APP, selectedValues,
```
snopoke, line 80: how did this fail before? It's not clear from the test what's actually being tested.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -178,6 +183,9 @@ public void configure(AuthenticatedRequestBean authenticatedRequestBean, HqAuth
         log.info(String.format("configuring RestoreFactory from authed request with arguments " +
                         "username = %s, asUsername = %s, domain = %s",
                 username, asUsername, domain));
+        if (authenticatedRequestBean instanceof InstallRequestBean) {
+            storageFactory.configure(((InstallRequestBean)authenticatedRequestBean));
```
snopoke, line 187: this is already happening in the [AppInstallAspect](https://github.com/dimagi/formplayer/blob/dffa67cb818c8474eed05ec6983c8a0ad7483198/src/main/java/org/commcare/formplayer/aspects/AppInstallAspect.java#L43). I think the best approach for now is to add the `@AppInstall` annotation to whatever view it is missing from.

src/test/java/org/commcare/formplayer/tests/MultiSelectCaseListTest.java
shubham1g5: it was hitting a session navigation exception due to the datum being set as  `use_selected_values` [here](https://github.com/dimagi/formplayer/blob/fixesForMultiSelect/src/main/java/org/commcare/formplayer/session/MenuSession.java#L176) . Probably [this](https://github.com/dimagi/formplayer/pull/1122/commits/4b585d8cb76a493b00f23e00cdf37b6f7a52856e) is better failure mode.

src/test/resources/archives/case_claim_with_multi_select/suite.xml
```diff
@@ -232,7 +232,7 @@
     <command id="m1-f1"/>
   </menu>
   <remote-request>
-    <post url="http://localhost:8000/a/test/phone/claim-case/" relevant="true()">
+    <post url="http://localhost:8000/a/test/phone/claim-case/" relevant="$case_id != ''">
```
snopoke, line 235: Does `$case_id` refer to the value of the data element? Where is that set in the eval context?

snopoke: Ah: https://github.com/dimagi/commcare-core/pull/1097

snopoke (approved):

PR #1123, opened 2022-04-15
Cache FormDef by session

gherceg:
As mentioned in https://github.com/dimagi/formplayer/pull/1121#issuecomment-1099297838
> It would be good to know specifically which part of the request lifecycle
this pooling is intended to optimize. If, for instance, it's the
deserialization of the object in between requests from the same user /
session it feels like caching the object in memory in between requests (if
available) is a safer play.

this is the implementation for caching FormDefs for specific sessions in between requests. As Cal mentions further in the thread, this would solve the majority of performance issues with regards to FormDef serialization. 

I based this on top of the existing shared form definition work, even though it isn't necessarily directly related. The SerializableFormDefinition does have all of the information needed in one place which made this easier. This does feel a bit clunky so if anyone has cleaner ideas on how to implement this, I'm all ears.

**I didn't add the user id as a cache key because it seemed safe to assume a session would be specific to a user**. If that is not the case definitely flag it.

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
+++ b/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
@@ -10,6 +10,7 @@ import org.commcare.formplayer.repo.FormDefinitionRepo;
 import org.commcare.formplayer.util.serializer.FormDefStringSerializer;
 import org.javarosa.core.log.WrappedException;
 import org.javarosa.core.model.FormDef;
+import org.javarosa.core.util.externalizable.DeserializationException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cache.annotation.CacheConfig;
 import org.springframework.cache.annotation.Cacheable;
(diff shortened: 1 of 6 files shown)


src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
```diff
@@ -59,4 +60,31 @@ public SerializableFormDefinition getOrCreateFormDefinition(
...
+     * @param formXmlns xmlns for specific form
+     * @param formVersion version for specific form
+     * @param sessionId form session id
+     * @param serializedFormDef serialized FormDef object
+     * @return deserialized FormDef object
+     */
+    @Cacheable(key = "{#appId, #formXmlns, #formVersion, #sessionId}")
+    public FormDef getFormDef(String appId,
+            String formXmlns,
+            String formVersion,
```
gherceg, line 79: The serializable form session ID is unique across all sessions, so we should be able to just cache on that right?

gherceg:
Tested locally and confirmed that the `getFormDef` method is only run for the first answer request, and then the cached FormDef is returned on subsequent answer requests. 

2 potential issues with this current implementation: 
- Since we need to first save the SerializableFormSession to obtain the id that we will cache on, we cannot take advantage of caching a session until the second answer request, and will still need to deserialize a form def during the first answer request. We could call `getFormDef` after creating a SerializableFormSession with the only intention being to prime the cache. This would mean deserializing a FormDef twice during the session creation, though I suppose that upfront cost is potentially worth it.
- Potentially related to the above point, we do have access to a deserialized FormDef when creating the session. This solution doesn't feel clever enough to take advantage of that, unless there is a way to manually update the cache for the `getFormDef` method with that existing object? 🤔  I'll look into that.

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
gherceg: See 0cefa6fb921734998a392837e4f06c12aa23126f

PR #1126, opened 2022-04-19
Use shared form definition from postgres

gherceg:
[SAAS-13377](https://dimagi-dev.atlassian.net/browse/SAAS-13377)

### Context
This is a re-do of https://github.com/dimagi/formplayer/pull/1075 since the commit history got pretty messy. I know this causes an inconvenience to reviewers, but I think the clean commit history is better in the long run. The link to the closed PR will still be available here if more context if needed, but I'll try to summarize everything below. The only difference between this PR and the previous is that this includes the FormDefinitionServiceTest class. Everything else remains the same. I've tested this locally, and plan to run another pass of QA on this to ensure nothing went wrong in creating this new branch.

### Summary
This PR makes use of the newly added SerializableFormDefinition to store unique form definitions in postgres. A form definition is considered unique based on the app id, form xmlns, and form version. So for a new value in any of those three variables, a new SerializableFormDefinition is created.

Notably, the current implementation in this PR uses the _app_ version in place of the _form_ version due to an existing bug that is being tracked in [SAAS-13364](https://dimagi-dev.atlassian.net/browse/SAAS-13364) where the form version is not set on the FormInstance object. This only results in creating more SerializableFormDefinitions than potentially necessary, as the app version is updated for any change to any form or app setting, whereas the form version is only set to the latest app version when a change to that specific form is made.

### Rollback Plan
This PR is **safe to rollback** as it continues to write to both the SerializableFormSession and SerializableFormDefinition, and falls back on reading from the SerializableFormSession if it cannot find a matching SerializableFormDefinition, which means it will not break existing incomplete sessions.

### QA
https://dimagi-dev.atlassian.net/browse/QA-3798

### Tests
I've added tests for the FormDefinitionRepo and FormDefinitionService.

### DIFF
--- a/src/main/java/org/commcare/formplayer/session/FormSession.java
+++ b/src/main/java/org/commcare/formplayer/session/FormSession.java
@@ -14,6 +14,7 @@ import org.commcare.formplayer.beans.FormEntryNavigationResponseBean;
 import org.commcare.formplayer.beans.FormEntryResponseBean;
 import org.commcare.formplayer.objects.FormVolatilityRecord;
 import org.commcare.formplayer.objects.FunctionHandler;
+import org.commcare.formplayer.objects.SerializableFormDefinition;
 import org.commcare.formplayer.objects.SerializableFormSession;
 import org.commcare.formplayer.sandbox.UserSqlSandbox;
 import org.commcare.formplayer.services.FormplayerStorageFactory;
@@ -105,9 +106,17 @@ public class FormSession {
         restoreFactory.setPermitAggressiveSyncs(false);
 
         this.sandbox = restoreFactory.getSandbox();
-        this.formDef = FormDefStringSerializer.deserialize(session.getFormXml());
-        loadInstanceXml(formDef, session.getInstanceXml());
-        formDef.setSendCalloutHandler(formSendCalloutHandler);
+
+        if (session.getFormDefinition() != null) {
+            this.formDef = FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
+        } else {
+            // DEPRECATED: leave to allow rollback if needed
+            // Will remove once https://github.com/dimagi/formplayer/pull/1075 is safe
+            this.formDef = FormDefStringSerializer.deserialize(session.getFormXml());
+        }
+
+        loadInstanceXml(this.formDef, session.getInstanceXml());
+        this.formDef.setSendCalloutHandler(formSendCalloutHandler);
         setupJavaRosaObjects();
 
         if (session.isOneQuestionPerScreen() || session.isInPromptMode()) {
@@ -126,6 +135,7 @@ public class FormSession {
     }
 
     public FormSession(UserSqlSandbox sandbox,
+            SerializableFormDefinition serializableFormDefinition,
             FormDef formDef,
             String username,
             String domain,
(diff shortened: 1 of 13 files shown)


snopoke (approved):

gherceg:
I requested QA run another pass on this yesterday which they did and all looks good!

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -105,9 +106,17 @@ public FormSession(SerializableFormSession session,
         restoreFactory.setPermitAggressiveSyncs(false);
 
         this.sandbox = restoreFactory.getSandbox();
-        this.formDef = FormDefStringSerializer.deserialize(session.getFormXml());
-        loadInstanceXml(formDef, session.getInstanceXml());
-        formDef.setSendCalloutHandler(formSendCalloutHandler);
+
+        if (session.getFormDefinition() != null) {
+            this.formDef = FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
```
shubham1g5, line 111: @gherceg Think we should handle code breaking due to deserialization here. There can be changes in the underlying serialization methods (`readExternal` and `writeExternal`) due to which deserializing already serialized form def will break.  

Separately curious if we should clear `form_definition` table when user perform 'clear data' ?

snopoke: With the changes in https://github.com/dimagi/formplayer/pull/1127 I think it's simpler though we might have to store the raw XML as well so that we can re-parse it. This would be required especially for SMS sessions where we don't have the XML in an app DB.

shubham1g5: Does SMS sessions send the raw xml in the request data ? I am seeing only 2 path ways for someone to enter into a form - 
1. (NewFormResponseFactory](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java#L56) which uses xml from the request bean itself
2. [Normal menu navigation](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/session/MenuSession.java#L327) which seems to be getting the raw `FormDef` from the DB already

Does the sms uses sessions use the first pathway here ?

Would a solution here be just catching the deserialization exception and then reserialise raw form xml from one of the 2 sources (DB or request data) and store it in the `FormDefinition` postgress table ?

shubham1g5: > I think the form XML is only sent in the first request for SMS forms

Curious if that's the case why doesn't it error [here](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java#L63) already ?

snopoke: That code is only called on the first request (when the session is started). After that it calls the 'answer', 'current', 'next' and 'get-instance' endpoints which require an existing session.

gherceg: Thanks for raising Shubham and the additional feedback Simon. Just so I understand how this fits in with the changes made so far in #1127, at the moment that removes fetching the raw xml from the local db even in the case of normal menu navigation, and solely relies on the serialized xml from postgres which is risky. So the two changes that need to be made there are:
- fetch raw xml from db if deserialization fails in the FormDefinitionService (related to [this thread](https://github.com/dimagi/formplayer/pull/1127#discussion_r870368438))
- save raw xml to db when opening sms sessions so the previous step can be successful for sms sessions as well

gherceg: See 87f7c5039d82ff4f6c2a91ddbd62e7a47762bd8f

PR #1127, opened 2022-04-20
Cache deserialized FormDef by session id

gherceg:
I had to create a new PR since the previous PR was based off of a branch that will no longer be merged. See https://github.com/dimagi/formplayer/pull/1123 for any potentially missing context.

This work is a response to [Clayton's suggestion](https://github.com/dimagi/formplayer/pull/1121#issuecomment-1099297838) to target what specifically we want to optimize. In this case, our main goal is to improve answer timings since a user filling out a form should not have to deserialize a FormDef on every request. This still results in a FormDef being deserialized upon session creation, but that is less likely to experience noticeable gains than answer requests and submission are, since it is an already slower request.

6cd6bc1eb0d9df31a39fc6f17518c4b17cb5f06f is the bulk of this PR, but has a crucial missing component which is that we would not call `getFormDef` until the first answer request, which means we wouldn't be taking advantage of a cached FormDef until the second answer request.

Since we have to deserialize a FormDef upon session creation, it is best to prime the `form_definition` cache with that FormDef which is what 54aff2bfb2d56d7ccbff1c516594115220dbf762 does.

QA: https://dimagi-dev.atlassian.net/browse/QA-4079

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
+++ b/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
@@ -99,7 +99,8 @@ public class FormDefinitionService {
             formDef = FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
         } catch (Exception e) {
             String xmlns = session.getFormDefinition().getFormXmlns();
-            formDef = this.getFormDefStorage().getRecordForValue(FormDef.STORAGE_KEY, xmlns);
+            // TODO: getStorageManager() returns different instance that does not have FORMDEF registered
+            formDef = this.getFormDefStorage().getRecordForValue("XMLNS", xmlns);
         }
         return formDef;
     }
--- a/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
+++ b/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
@@ -7,6 +7,7 @@ import org.commcare.formplayer.session.FormSession;
 import org.commcare.formplayer.util.serializer.FormDefStringSerializer;
 import org.javarosa.core.log.WrappedException;
 import org.javarosa.core.model.FormDef;
+import org.javarosa.core.services.storage.IStorageUtilityIndexed;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cache.annotation.CacheConfig;
 import org.springframework.cache.annotation.CachePut;
@@ -25,6 +26,9 @@ public class FormDefinitionService {
 
     @Autowired
     private FormDefinitionRepo formDefinitionRepo;
+
+    @Autowired
+    private FormplayerStorageFactory storageFactory;
 
     /**
      * Checks if an entry for this (appId, formXmlns, formVersion) combination already exists, and returns if so
@@ -59,6 +63,18 @@ public class FormDefinitionService {
     }
 
     /**
+     * Ensure raw xml is accessible locally in case serialization changes which would break the ability to
+     * deserialize the serialized form def in postgres
+     */
+    public void writeToLocalStorage(FormDef formDef) {
+        IStorageUtilityIndexed<FormDef> formStorage = this.getFormDefStorage();
+        String xmlns = formDef.getMainInstance().schema;
+        if (formStorage.getRecordForValue("XMLNS", xmlns) == null) {
+            formStorage.write(formDef);
+        }
+    }
+
+    /**
      * This method uses the sessionId to cache deserialized FormDefs
      * Deserializing a serialized FormDef object is costly, and this avoids doing so in between requests
      * within the same session
@@ -82,7 +98,8 @@ public class FormDefinitionService {
         try {
             formDef = FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
         } catch (Exception e) {
-            formDef = null;
+            String xmlns = session.getFormDefinition().getFormXmlns();
+            formDef = this.getFormDefStorage().getRecordForValue(FormDef.STORAGE_KEY, xmlns);
         }
         return formDef;
     }
@@ -96,5 +113,9 @@ public class FormDefinitionService {
     @CachePut(key = "#session.getSessionId()")
     public FormDef cacheFormDef(FormSession session) {
         return session.getFormDef();
+    }
+
+    private IStorageUtilityIndexed<FormDef> getFormDefStorage() {
+        return this.storageFactory.getStorageManager().getStorage(FormDef.STORAGE_KEY);
     }
 }
--- a/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
+++ b/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
@@ -66,8 +66,18 @@ public class FormDefinitionService {
      * @param session session that contains session id and serialized formDef
      * @return deserialized FormDef object
      */
-    @Cacheable(key = "#session.id")
     public FormDef getFormDef(SerializableFormSession session) {
+        FormDef formDef = this.internalGetFormDef(session);
+        // ensure previous tree references are cleared (only necessary when retrieving from cache)
+        formDef.getMainInstance().cleanCache();
+        return formDef;
+    }
+
+    /**
+     * Always use public getFormDef to ensure internal FormDef cached references are cleared
+     */
+    @Cacheable(key = "#session.id")
+    private FormDef internalGetFormDef(SerializableFormSession session) {
         FormDef formDef;
         try {
             formDef = FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
--- a/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
+++ b/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
@@ -75,19 +75,12 @@ public class FormDefinitionService {
      * @return True if the value was written to storage or False if it already exists in storage.
      */
     public boolean writeToLocalStorage(FormDef formDef) {
-        IStorageUtilityIndexed<FormDef> formStorage = this.getFormDefStorage();
         String xmlns = formDef.getMainInstance().schema;
-        FormDef existing;
-        try {
-            existing = formStorage.getRecordForValue("XMLNS", xmlns);
-        } catch (NoSuchElementException e) {
-            existing = null;
+        if (getFormDefFromStorage(xmlns).isPresent()) {
+            return false;
         }
-        if (existing == null) {
-            formStorage.write(formDef);
-            return true;
-        }
-        return false;
+        writeFormDefToStorage(formDef);
+        return true;
     }
 
     /**
@@ -99,7 +92,7 @@ public class FormDefinitionService {
      * @return deserialized FormDef object
      */
     public FormDef getFormDef(SerializableFormSession session) {
-        FormDef formDef = this.internalGetFormDef(session);
+        FormDef formDef = this.getFormDefWithCache(session);
         // ensure previous tree references are cleared (only necessary when retrieving from cache)
         formDef.getMainInstance().cleanCache();
         return formDef;
@@ -110,18 +103,20 @@ public class FormDefinitionService {
      *
      * We can't use @{@link Cacheable} here since we want to keep this method private.
      */
-    private FormDef internalGetFormDef(SerializableFormSession session) {
+    private FormDef getFormDefWithCache(SerializableFormSession session) {
         Cache cache = caches.getCache("form_definition");
-        FormDef formDef = cache.get(session.getId(), () -> {
-            try {
-                return FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
-            } catch (Exception e) {
-                String xmlns = session.getFormDefinition().getFormXmlns();
-                // TODO: getStorageManager() returns different instance that does not have FORMDEF registered
-                return this.getFormDefStorage().getRecordForValue("XMLNS", xmlns);
-            }
-        });
-        return formDef;
+        return cache.get(session.getId(), () -> getFormDefFromSession(session));
+    }
+
+    private FormDef getFormDefFromSession(SerializableFormSession session) {
+        try {
+            return FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
+        } catch (Exception e) {
+            String xmlns = session.getFormDefinition().getFormXmlns();
+            return getFormDefFromStorage(xmlns).orElseThrow(() -> {
+                return new WrappedException("Unable to load form def after serialization error", e);
+            });
+        }
     }
 
     /**
@@ -135,7 +130,21 @@ public class FormDefinitionService {
         return session.getFormDef();
     }
 
+    private Optional<FormDef> getFormDefFromStorage(String xmlns) {
+        // TODO: getStorageManager() returns different instance that does not have FORMDEF registered
+        try {
+            return Optional.of(getFormDefStorage().getRecordForValue("XMLNS", xmlns));
+        } catch (NoSuchElementException e) {
+            return Optional.empty();
+        }
+    }
+
+    private void writeFormDefToStorage(FormDef formDef) {
+        getFormDefStorage().write(formDef);
+    }
+
     private IStorageUtilityIndexed<FormDef> getFormDefStorage() {
         return this.storageFactory.getStorageManager().getStorage(FormDef.STORAGE_KEY);
     }
+
 }
--- a/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
+++ b/src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
@@ -79,7 +79,7 @@ public class FormDefinitionService {
             throw new WrappedException("Error serializing form def", e);
         }
         formDefinition.setSerializedFormDef(serializedFormDef);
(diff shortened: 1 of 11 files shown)


src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
```diff
@@ -54,4 +54,24 @@ public SerializableFormDefinition getOrCreateFormDefinition(
...
+     *
+     * @param sessionId form session id
+     * @param serializedFormDef serialized FormDef object
+     * @return deserialized FormDef object
+     */
+    @Cacheable(key = "#sessionId")
+    public FormDef getFormDef(String sessionId, String serializedFormDef) {
+        FormDef formDef;
+        try {
+            formDef = FormDefStringSerializer.deserialize(serializedFormDef);
```
snopoke, line 71: You could simplify this for callers by taking in the whole session object instead of passing the ID and serialized formDef separately:

```suggestion
    @Cacheable(key = "#session.id")
    public FormDef getFormDef(SerializableFormSession session) {
        FormDef formDef;
        try {
            formDef = FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
```

gherceg: Good point. See 5068f204cea26929ec844d477ed79d290fb0567c

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
```diff
@@ -118,6 +123,9 @@ public NewFormResponse getResponse(FormSession formEntrySession) throws Exceptio
 
         SerializableFormSession serializedSession = formEntrySession.serialize();
         formSessionService.saveSession(serializedSession);
+        // prime the form def cache for this session
+        Cache cache = this.cacheManager.getCache("form_definition");
+        cache.put(formEntrySession.getSessionId(), formEntrySession.getFormDef());
```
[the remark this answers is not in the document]
gherceg: That does simplify things. See bdd1518830166ae3b955e1ed0f7b32a6d561f98a

gherceg:
> what's the status of this PR?

On the verge of getting to the bottom of the issue that arose in QA when caching sessions. There's a strange edge case where the question node from the previous request is referenced when answering questions with a null value. Typical behavior is for this question node to be copied as part of the `loadInstanceXml` method, and to set the value on this fresh copy of the question node.

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
```diff
@@ -54,4 +57,34 @@ public SerializableFormDefinition getOrCreateFormDefinition(
...
+     * @param session session that contains session id and serialized formDef
+     * @return deserialized FormDef object
+     */
+    @Cacheable(key = "#session.id")
+    public FormDef getFormDef(SerializableFormSession session) {
+        FormDef formDef;
+        try {
+            formDef = FormDefStringSerializer.deserialize(session.getFormDefinition().getSerializedFormDef());
+        } catch (Exception e) {
+            formDef = null;
```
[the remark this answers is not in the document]
gherceg: That makes sense to me especially with the concerns you brought up in the #1126

gherceg:
Finally got to the bottom of the issue. Answers were being set on a node that was cached from the previous answer request. I wasn't aware of this lower level cache so it took a bit of digging. I think the best bet is to always ensure this cache is cleared prior to returning the FormDef from `getFormDef`. I figured it wasn't worth adding a check to determine if the FormDef came from cache or not, since a newly deserialized FormDef doesn't have anything in the cache yet anyway. See a4e94177caee451ada47eba4016fe090821e174f

shubham1g5:
> when writing to local storage it isn't clear how that xmlns value would get mapped to allow us to fetch FormDefs

@gherceg [This is where](https://github.com/dimagi/commcare-core/blob/formplayer/src/main/java/org/javarosa/core/model/FormDef.java#L1671) the xmlns gets mapped.

gherceg:
So I took a simple approach to the StorageManager issue and just registered FormDef.STORAGE_KEY as needed in 51c94a657be652c3d2cb0479a6498045352a27af (with a minor change in 6ac2848ca04a1d1ae43d34ef30dbecfba5151b7b). I did this because I noticed a similar pattern in the [CommCarePlatform](https://github.com/dimagi/commcare-core/blob/9a4192a98e640b1fcf0e8b1fc67b13c4f663de96/src/main/java/org/commcare/util/CommCarePlatform.java#L56) which is setup when initializing a `FormSession`. The reason I didn't add it there is because that occurs after we would have already wanted to read the form def from local storage.

Then knowing that the sms entry point did not follow the same pattern as other form entry points with the `ConfigureStorageFromSession` decorator (requires an already existing session), I used the same `configure` line found in the [ConfigureStorageFromSession](https://github.com/dimagi/formplayer/blob/0a6850fdbeb61d8c5dedcda4c4c338e350fbd275/src/main/java/org/commcare/formplayer/aspects/ConfigureStorageFromSessionAspect.java#L29) and added that after the `SerializableFormSession` was created in the `NewFormResponseFactory`. See 07d76fe2a6924089c2b76d0750efa4170cf7e24c.

Lastly, I added a bit of code to ensure the serialized FormDef is updated on the shared form_definition table if deserialization fails in the first place. See 296854b3f1c65fb483fb9624cde9868022225234

snopoke (commented):
I think it would be better to register the storage preemptively since it makes the code easier to reason about.

There are 2 places I think it should go:
https://github.com/dimagi/formplayer/blob/6cd6bc1eb0d9df31a39fc6f17518c4b17cb5f06f/src/main/java/org/commcare/formplayer/application/AbstractBaseController.java#L92-L95

and

https://github.com/dimagi/formplayer/blob/0783752d088d3f2c34937c722f80c44ae146293a/src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java#L69-L75

src/test/java/org/commcare/formplayer/services/FormDefinitionServiceTest.java
```diff
@@ -162,20 +176,104 @@ public void testGetOrCreateFormDefinitionNewVersionNotEqual() {
...
+        getFormDefStorage().write(this.formDef);
+        assertThat(this.formDef.getID()).isNotNull();
+
+        FormDef reSerializedFormDef = this.formDefinitionService.getFormDef(session);
+        assertThat(reSerializedFormDef.getID()).isEqualTo(this.formDef.getID());
+
+    }
+
+    @Test
+    public void testGetFormDefFromStorageUpdatesSerializableFormDefinition() throws Exception {
```
shubham1g5, line 231: we can probably abstract out the common setup between this and `testGetFormDefBrokenSerialization` into another method.

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
```diff
@@ -54,4 +67,102 @@ public SerializableFormDefinition getOrCreateFormDefinition(
...
+     * @param session grab the id to cache on and the deserialized form def from the session
+     * @return deserialized FormDef object
+     */
+    @CachePut(key = "#session.getSessionId()")
+    public FormDef cacheFormDef(FormSession session) {
+        return session.getFormDef();
+    }
+
+    private Optional<FormDef> getFormDefFromStorage(String xmlns) {
+        this.storageFactory.getStorageManager().registerStorage(FormDef.STORAGE_KEY, FormDef.class);
```
shubham1g5, line 152: doesn't `storageFactory` already have the storage registered ?

snopoke: no, that's the whole crux of the matter. Specifically for SMS forms there is no app install and so the storage never get's registered.

gherceg:
@snopoke I agree that we should register FormDef storage at a higher level, but related to what I mentioned [here](https://github.com/dimagi/formplayer/pull/1127#discussion_r874827942), 
>  FormDef storage is only registered when setting up a FormplayerConfigEngine, which only happens during navigate menu session requests.

the AbstractBaseController is only used during menu session navigation, so I'm leaning towards adding in the [ConfigureStorageFromSessionAspect](https://github.com/dimagi/formplayer/blob/0a6850fdbeb61d8c5dedcda4c4c338e350fbd275/src/main/java/org/commcare/formplayer/aspects/ConfigureStorageFromSessionAspect.java#L29) instead.

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
gherceg: A `CommCarePlatform` is setup during initialization of a FormSession, but that does not register FormDef storage, only [registers PropertyManager storage](https://github.com/dimagi/commcare-core/blob/9a4192a98e640b1fcf0e8b1fc67b13c4f663de96/src/main/java/org/commcare/util/CommCarePlatform.java#L56). This happens _after_ we need access to a local form def, so we need to register the FormDef storage prior to this point anyway. 

In the context of the FormController, storage is configured due to the `@ConfigureStorageFromSession` decorator used on what appears to be all endpoints except sms. Registering FormDef storage after configuring the factory [here](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/aspects/ConfigureStorageFromSessionAspect.java#L29) makes the most sense to me.

I've been reproducing locally by modifying the FormDefinitionService to attempt to access the local storage on any `getFormDef`, though you really just need to be able to inspect the storage factory at that point. You'll see that the StorageManager returned from the factory does not have any registered storage yet.

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
```diff
@@ -106,6 +105,10 @@ public NewFormResponse getResponse(NewSessionRequestBean bean, String postUrl) t
                 caseSearchHelper
         );
 
+        // can only setup local storage once a form session has been created
+        this.storageFactory.configure(formSession.getSerializableSession());
```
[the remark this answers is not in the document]
gherceg: Totally missed that. I think I got excited cause I tested this locally and it worked, but in hindsight, in order to test sms locally I need to run a menu session first so that localization is properly setup (should be able to put out a fix for that), and I'm wondering if that is the same case here. Is it possible performing an app install for a menu session prior to an sms form session allows the sms form session to have access to a sqlite db?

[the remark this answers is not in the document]
snopoke: I don't think that would work for SMS since it uses a different data directory so a previous app install wouldn't be accessible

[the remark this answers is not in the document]
snopoke: For normal sessions a previous app install of the same app for that user would have created the DBs (though each request still needs to register them with the storage manager) and it's quite possible this is only happening after we load the form def from the DB.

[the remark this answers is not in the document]
ctsims: @gherceg It's quite likely that a previous app install is resulting in a thread later having a Localizer. In Spring the worker threads don't terminate after completion (they are reused between requests). If the issue here is simply that a Localizer was never created on the SMS thread, a previously successful menusession request that created a threadstatic localizer would likely change the behavior for a subsequent sms session request.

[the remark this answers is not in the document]
gherceg: Yeah I figured the localizer mechanism and the app db mechanisms are quite different. I'm fairly confident what you are describing is exactly what happens in the case of the localizer.

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
```diff
@@ -54,4 +67,102 @@ public SerializableFormDefinition getOrCreateFormDefinition(
...
+            SerializableFormDefinition formDefinition,
+            FormDef formDef) {
+        String serializedFormDef;
+        try {
+            serializedFormDef = FormDefStringSerializer.serialize(formDef);
+        } catch (IOException e) {
+            throw new WrappedException("Error serializing form def", e);
+        }
+        formDefinition.setSerializedFormDef(serializedFormDef);
+        return this.formDefinitionRepo.save(formDefinition);
```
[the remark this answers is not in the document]
gherceg: Ah wasn't sure what the java best practices were around the use of `this`. Out of curiosity do we have a formplayer convention? Or is it just use when necessary and that's it?

src/test/java/org/commcare/formplayer/services/FormDefinitionServiceTest.java
gherceg: Agreed that there is room for abstraction, but it's one line that is specifically related to setting up a broken serialization. I also added an additional test that relies on this setup. While all three tests could be consolidated into one test with multiple assertions, in HQ we tend to keep assertions in separate tests to make it easy to reason with.

All that to say I don't mind a bit of repetition if it keeps tests relatively simple. I don't see a super obvious abstraction to be made that wouldn't overcomplicate things, so feel free to highlight that or suggest a change if you want. Otherwise I'll likely leave as is.

src/main/java/org/commcare/formplayer/services/FormDefinitionService.java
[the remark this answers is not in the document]
shubham1g5: > Or is it just use when necessary and that's it

that's what I follow, I think it's better to omit `this` when not necessary to keep the code less verbose.

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
```diff
@@ -71,10 +71,11 @@ public NewFormResponse getResponse(NewSessionRequestBean bean, String postUrl) t
                 bean.getSessionData().getAppId(),
                 bean.getRestoreAs(),
                 bean.getRestoreAsCaseId());
+        storageFactory.getStorageManager().registerStorage(FormDef.STORAGE_KEY, FormDef.class);
```
snopoke, line 74: What do you think about putting this in a method in `FormplayerStorageFactory` with a javadoc to give a bit more context of why it's needed? It would also tie the two usages together.

src/test/java/org/commcare/formplayer/services/FormDefinitionServiceTest.java
```diff
@@ -162,20 +177,133 @@ public void testGetOrCreateFormDefinitionNewVersionNotEqual() {
...
+        session.setFormDefinition(formDef);
+        getFormDefStorage().write(this.formDef);
+
+        FormDef reSerializedFormDef = this.formDefinitionService.getFormDef(session);
+
+        SerializableFormDefinition updatedFormDef = this.formDefinitionService.getOrCreateFormDefinition(
+                this.appId, this.formXmlns, this.formVersion, this.formDef
+        );
+        assertEquals(updatedFormDef.getSerializedFormDef(),
+                FormDefStringSerializer.serialize(reSerializedFormDef));
```
snopoke, line 247: I don't think this is actually testing anything since the broken form def was never saved to the DB. I think you'd need to save the broken SerializableFormDefinition before calling `formDefinitionService.getFormDef(session)`.

You can confirm this by removing the call to `this.formDefinitionService.getFormDef(session)` and seeing if the test passes.

snopoke: Same goes for `testGetFormDefFromStorageUpdatesCache`

src/test/java/org/commcare/formplayer/services/FormDefinitionServiceTest.java
snopoke: You could extract out the test setup:

```java
public SerializableFormDefinition createBrokenFormDef() {
    SerializableFormDefinition formDef = this.formDefinitionService.getOrCreateFormDefinition(
            this.appId, this.formXmlns, this.formVersion, this.formDef
    );
    ReflectionTestUtils.setField(formDef, "serializedFormDef", "not a form def");

    // this line added based on PR comments below
    formDefinitionRepo.save(formDef);

    SerializableFormSession session = new SerializableFormSession(UUID.randomUUID().toString());
    session.setFormDefinition(formDef);
    getFormDefStorage().write(this.formDef);
    
    return formDef
}

Comment (gherceg) [reply src/test/java/org/commcare/formplayer/services/FormDefinitionServiceTest.java:221]: Fair point 😅  see 70111ce3e775fe57f4a4bda134179ab20b554d5d

File: src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java (hunk 8 above)
Comment (gherceg) [reply src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java]: Yeah anything to make this less confusing in the future sounds good to me. See d82f76ea27cde69ac1bd3b1180d744162ebd9307 and let me know if the javadoc could be more informative.

File: src/main/java/org/commcare/formplayer/services/FormplayerStorageFactory.java [hunk 10]
```diff
@@ -102,6 +105,15 @@ public void preDestroy() {
         }
     }
 
+    /**
+     * Call after configuring factory if form def storage needs to be accessible
+     * Menu navigation requests setup form def storage as part of MenuSession initialization
+     * Form controller requests do not setup form def storage as part of FormSession initialization
```
snopoke, line 111: nit:
```suggestion
     * Call after configuring factory if form def storage needs to be accessible.
     * Menu navigation requests setup form def storage as part of MenuSession initialization.
     * Form controller requests do not setup form def storage as part of FormSession initialization.
```

snopoke (approved):

PR #1130, opened 2022-04-21
Fixes user equality match

shubham1g5:
https://dimagi-dev.atlassian.net/browse/QA-4081

`storageFactory` seems to be injected with scrubbed username sometimes but not always. So doing a euality check by scrubbing both sides of the equation.


orangejenny (approved):

snopoke (approved):
That's very unfortunate!

PR #1132, opened 2022-04-26
Remove formxml related code

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-13495

Followup to https://github.com/dimagi/formplayer/pull/1126. That PR was deployed last Thursday, the 21st. We need to wait at least 7 days from that date before merging/deploying this to ensure all incomplete form sessions have been purged. This change is _not_ backwards compatible in that a form session created prior to shared form definitions would not be able to resume properly. This is because we are no longer going to reference a serializable form session's `formxml` property.

Once this has been deployed and deemed safe, the last step will be to deploy a migration that removes the `formxml` column from `SerializableFormSession`.


snopoke (approved):
I think it would be prudent to wait more than 7 days to give a bigger window for issues to be raised.

gherceg:
@snopoke I agree. I'll update the ticket to wait until at least next week which would give us 14 days. Thanks for adding the label 👍🏻

ctsims:
Quick question / note - Will anyone self-hosting get this maintenance for free as part of an automated migration, or are we manually managing this process in some way that won't come through in a normal deploy?

ctsims:
If someone has changed that value to longer than 2 weeks, they would need to manually clear any older entries, right? Is there a specific way for someone to do that? If there is, and we specifically communicate it upfront in the changelog before an automated update would apply, that fulfills our obligation.

I think it's ok (although not ideal) to push manual maintenance to teams who are doing something different than a default, but if there's no practical way to accomplish that manual step then I think the situation gets more complicated.

gherceg:
The risk is if there were incomplete form sessions created before https://github.com/dimagi/formplayer/pull/1126 was deployed, that **still exist** after this PR is deployed, those incomplete sessions would not function properly. Manual clearing of older form sessions would be necessary only if we view that as a solution to broken incomplete sessions.

In addition to the possibility of self-hosters modifying the purge window, and perhaps more concerning, there is also the variable of when they actually deployed https://github.com/dimagi/formplayer/pull/1126.

To summarize there are a few solutions:
1) (mentioned above) provide an easy way to delete existing form sessions that do not reference a shared form definition, which does not eliminate the risk of broken form sessions but gives self-hosters the ability to remove them easily
2) Taking that one step farther, automate the cleanup of form sessions that do not reference a shared form definition in this PR via a migration, which would be a potential breaking change that would need to be signed off by other divisions and communicated properly. If we waited long enough, it's possible this task wouldn't find any broken form sessions as they've all been purged on all envs.
3) Coordinate with _all_ self-hosters, ensuring #1126 is deployed, all hosters have all waited for the purge window to cleanup old form sessions, and then we could merge this PR

I think the pressure to roll this out has subsided since other DB/storage gains have been made recently, so we could go with option 2 + waiting an adequate amount of time (e.g., a month) to merge this.

ctsims:
>
> there is also the variable of when they actually deployed #1126
> <https://github.com/dimagi/formplayer/pull/1126>.
>
Our SoP (and something we're working towards codifying more publicly) is
that we expect a routine deploy every 2 weeks, to resolve any ambiguity
with sequencing. A deploy should always be able to assume that any other
code that is > 2 weeks old has been deployed.

Just in general, btw, it's worth excluding from consideration any planning
based on "coordination with all self hosters." For scalability and for lack
of ambiguity for people performing work, we should always assume that
people are following the maintenance expectations
<https://docs.google.com/document/d/1EXVoCheX44bVs_GxVL2d7LOTsDZybKPbQ8l4kWU2v8M/edit>
rather
than coming up with bespoke plans for rolling out changes.

If the failure mode from this is "incomplete forms from before XYZ just
won't load anymore" and not that Formplayer itself is broken for new work,
my sense is that we shouldn't be putting much work into mitigation. There
are other factors which can break incomplete forms, and I think it's a
small practical concern for temporary data models from 2 weeks ago to no
longer be available.

Based on everything we've discussed, my sense of the most practical move
based on the impact and amounts of effort is:

   1. Wait 4 weeks after 1126 was deployed (2 weeks for the code to get
   onto a machine, and then 2 weeks for the incomplete form window to expire)
   2. Announce in the changelog that any servers where the incomplete form
   retention window has been extended past 2 weeks will need to clear
   incomplete forms older than 2 weeks in the next deploy (expecting people to
   do it from the db is probably fine without providing a tool, given how
   minor the impact is and how unlikely that is)
   3. Roll out the removal as-is with no further mitigation

I was initially thinking it would have been good to mention that anyone who
has changed the form retention window will have an upcoming "action needed"
deploy when we rolled out #1126, but I'm not sure how actionable that will
be.

-Clayton

gherceg:
Thanks for the feedback Clayton.

> the failure mode from this is "incomplete forms from before XYZ just
won't load anymore"

Confirming this is the only failure mode.

> Wait 4 weeks after 1126 was deployed (2 weeks for the code to get
   onto a machine, and then 2 weeks for the incomplete form window to expire)

Th incomplete form window is only 1 week. Simon's suggestion to wait an additional week was to ensure any potential issues were raised before removing backwards compatibility which I think overlaps with the additional 2 weeks needed to ensure code is deployed by self-hosters. All that to say I think 3 weeks is sufficient, so noting that we would roll this out on May 12th at the earliest.

> Announce in the changelog that any servers where the incomplete form
   retention window has been extended
   
   As far as timing this goes, can this be announced in the changelog the same day it is scheduled to be rolled out? Or is there a certain amount of notice we are required to give?

snopoke:
> I think it's ok (although not ideal) to push manual maintenance to teams who are doing something different than a default, but if there's no practical way to accomplish that manual step then I think the situation gets more complicated.

Just noting that if they have changed the purge window they would have to fork the code and therefore also would need the have their own process for pulling in changes from this repo. That is all to say that I don't think we owe anyone in that situation more than the basic changelog.

snopoke (approved):
Code changes look good. It would be good to put this on staging as a final check.

gherceg:
Deployed to staging and things look good 👍🏻

PR #1135, opened 2022-04-28
Added breadcrumb support for MultiSelectEntityScreen

orangejenny:
PR into https://github.com/dimagi/formplayer/pull/1109

Screen Shot 2022-04-28 at 2 42 26 PM

Depends on https://github.com/dimagi/commcare-core/pull/1105

### DIFF
--- a/src/main/java/org/commcare/formplayer/session/MenuSession.java
+++ b/src/main/java/org/commcare/formplayer/session/MenuSession.java
@@ -53,6 +53,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.UUID;
 
 import datadog.trace.api.Trace;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -210,20 +211,43 @@ public boolean autoAdvanceMenu(Screen screen, boolean autoAdvanceMenu)
     }
 
     private void addTitle(String input, Screen previousScreen) {
-        if (previousScreen instanceof EntityScreen) {
-            try {
-                String caseName = SessionUtils.tryLoadCaseName(sandbox.getCaseStorage(), input);
+        if (previousScreen instanceof MultiSelectEntityScreen) {
+            UUID referenceId = ((MultiSelectEntityScreen)previousScreen).getStorageReferenceId();
+            String[] values = entitiesSelectionCache.read(referenceId);
```
snopoke, line 216: This is no longer correct. We now store the full instance in the DB:
```java
VirtualDataInstance instance = virtualDataInstanceService.read(referenceId);
TreeElement root = instance.getRoot();
```

You'll probably have to evaluate an xpath expression (or you could just use TreeElement methods like 'getNumChildren' and 'getChildAt').

snopoke (changes requested):

src/main/java/org/commcare/formplayer/session/MenuSession.java
orangejenny: ah, I was working off outdated code. Updated in 15034ca9c00318843df954c6df7a4870374ad2e0, and exposed a `getVirtualInstance` method in https://github.com/dimagi/commcare-core/pull/1105

snopoke (approved):

PR #1137, opened 2022-05-02
Added conditionally required query prompts

orangejenny:
Formplayer support for https://github.com/dimagi/commcare-hq/pull/31568

Depends on https://github.com/dimagi/commcare-core/pull/1106


snopoke (approved):

shubham1g5 (commented):
A simple test for it can be added in [here](https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/CaseClaimTests.java).

orangejenny:
@shubham1g5 This is all web apps UI, though - all formplayer does is pass the values along. Does it seem worth testing?

shubham1g5:
@orangejenny I think we just want to test whether the query response has right value of `required`, for example how we test the blank value [here](https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/CaseClaimTests.java#L188).

orangejenny:
Added tests in 9bc22b1f

shubham1g5 (approved):

PR #1144, opened 2022-05-25
Spring boot 2.7

shubham1g5:
Updates to Spring Boot 2.7.0 and bumps PG for github builds to 14.2 in order to avoid ` org.flywaydb.core.internal.license.FlywayEditionUpgradeRequiredException: Flyway Teams Edition or PostgreSQL upgrade required: PostgreSQL 9.6 is no longer supported by Flyway Community Edition, but still supported by Flyway Teams Edition.
`

https://dimagi-dev.atlassian.net/browse/USH-1440


shubham1g5:
> I thought we couldn't do this upgrade until our infra was on > 9.6

Ahh I see, Docker seems to pull PG 14.2 locally for me which is why I thought our infra might already be on PG 14. But I guess we don't expect to have the same PG version locally as on production ?

snopoke:
> > I thought we couldn't do this upgrade until our infra was on > 9.6
> 
> Ahh I see, Docker seems to pull PG 14.2 locally for me which is why I thought our infra might already be on PG 14. But I guess we don't expect to have the same PG version locally as on production ?

From what I can tell we are using PG 10 on our infra so it's just blocking on the upgrade notice to 3rd parties. I've commented on the ticket about that.

shubham1g5:
@snopoke  https://dimagi-dev.atlassian.net/browse/SAAS-11793 seems to be done, does it mean we are unblocked on this ?

.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:9.6
+        image: postgres:14.2
```
snopoke, line 14: I think we should stick with 10 to match what's being run in production

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
 }
 
 plugins {
-    id 'org.springframework.boot' version '2.5.6'
+    id 'org.springframework.boot' version '2.7.0'
```
snopoke, line 8: might as well make this 2.7.2

snopoke (commented):
@shubham1g5 yes, I think we can go ahead with this now

snopoke (approved):

shubham1g5:
tested a bit on staging and didn't see any issues.

PR #1146, opened 2022-05-26
use query data in debugger endpoints

snopoke:
Companion to https://github.com/dimagi/commcare-hq/pull/31687 which passes query data to FP. These changes will use the query data when replaying the session to ensure correct session state.


shubham1g5 (approved):
💯

PR #1147, opened 2022-05-26
clear remote instance cache after submission

snopoke:
I've been thinking about doing this for some time, I have not fully tested it (especially with entry stack ops) but I'm putting it up for some initial feedback.

This is intended to solve this issue (as described by USH delivery):

> There is an issue that I believe is related to caching when using Case Search. The following issue can occur
Search for records with a certain criteria using “Search All Cases”
>  * A set of cases appears
>  * You update a case so that it no longer appears in the module
>  * You search for records with the same criteria 
>  * The case that you removed from the case list is returned

The first commit is a refactor, the last is an optimisation so actual commit of interest is aaabe47ac6507113ac2815877aa4881a14035efd

I'm putting this on staging for testing.


src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -91,7 +98,8 @@ public SubmitResponseBean submitForm(@RequestBody SubmitRequestBean submitReques
                 stepFactory.makeStep("processFormXml", this::processFormXml, PROCESSED_XML),
                 stepFactory.makeStep("updateVolatility", this::updateVolatility),
                 stepFactory.makeStep("performSync", this::performSync),
-                stepFactory.makeStep("doEndOfFormNav", this::doEndOfFormNav, PROCESSED_STACK)
+                stepFactory.makeStep("doEndOfFormNav", this::doEndOfFormNav, PROCESSED_STACK),
+                stepFactory.makeStep("clearCaches", this::clearCaches)
```
shubham1g5, line 102: I am concerned that this is going to cause [this issue](https://github.com/dimagi/formplayer/blob/autoLaunchTest/src/test/java/org/commcare/formplayer/tests/MultiSelectCaseClaimTest.java#L111) for session navigation requests after EOF navigation to a screen that comes after a case search result screen. Curious if it's possible to clear the cache is at the start of a new menu session instead ?

snopoke: The stack for EOF nav does separate queries which should not match the cache keys for the existing searches. You can see an example here: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/tests/test_suite_registry_search.py#L208-L221

Obviously I'll need to test that but I don't think it should be an issue.

snopoke:
> There is an issue that I believe is related to caching when using Case Search. The following issue can occur Search for records with a certain criteria using “Search All Cases”
> 
> * A set of cases appears
> * You update a case so that it no longer appears in the module
> * You search for records with the same criteria
> * The case that you removed from the case list is returned

With more investigation this change only solves the issue for 'inline search' since normal case search does not results in instances being saved in the session (due to the session rewind). I'm going to have to thing of an alternative approach to solve that.

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
shubham1g5: got it, these points make perfect sense. thanks!

shubham1g5 (approved):
looks great.

PR #1162, opened 2022-06-22
clear volatile session elements after stack ops

snopoke:
This replaces the cache clearing in https://github.com/dimagi/formplayer/pull/1147 with a more generic mechanism that clears the cache for any remote instances that were removed from the session.

This uses the StackObserver from https://github.com/dimagi/commcare-core/pull/1127 to find which steps have been removed and inspect them to see if they contain remote instances.

Note that this does not remove virtual instances, it only clears the case search result cache for remote instances.


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -557,20 +559,35 @@ public BaseResponseBean resolveFormGetNext(MenuSession menuSession) throws Excep
...
     }
 
+    private void clearVolatiles(SessionWrapper session, StackObserver observer) {
+        session.clearVolatiles();
+        for (StackFrameStep abandonedStep : observer.getRemovedSteps()) {
+            for (ExternalDataInstanceSource source : abandonedStep.getDataInstanceSources().values()) {
+                try {
+                    caseSearchHelper.clearCacheForInstanceSource(source);
+                } catch (InvalidStructureException e) {
+                    log.warn("Error clearing virtual instance cache");
```
orangejenny, line 585: Should this say remote not virtual? PR description says this only clears remote instances.

orangejenny (approved):

shubham1g5 (approved):
wow! great turnaround time on these :)

PR #1166, opened 2022-06-23
rename field and make mock service behave like the real one

snopoke:
pure code refactor with no functional changes


src/main/java/org/commcare/formplayer/objects/SerializableDataInstance.java
```diff
@@ -35,7 +35,7 @@ public class SerializableDataInstance {
 
     @Setter
     @Column(name = "key", updatable = false)
-    private String key;
+    private String namespacedKey;
```
snopoke, line 38: no DB change for this since the db column is still called 'key'

shubham1g5 (approved):

PR #1171, opened 2022-06-29
expire case search cache 60s after last access

snopoke:
https://dimagi-dev.atlassian.net/browse/USH-2210


orangejenny:
https://github.com/dimagi/formplayer/pull/1172 supersedes this one, right?

snopoke:
> #1172 supersedes this one, right?

Yes but I think it's still useful to be able to adjust the cache setting independently. I've reverted the timeout change but kept the other changes so this should now be a non-functional change.

PR #1175, opened 2022-07-01
Revert "Revert "Changes virtual instance source reference""

snopoke:
Reverts dimagi/formplayer#1174


snopoke:
Confirmed that this PR was not the cause of the issue.

PR #1179, opened 2022-07-08
test changes to use real cache configuration

snopoke:
A recent cache configuration refactor resulted in a broken cache configuration because test were all using mock / test only cache configurations.

This PR addresses that by making all tests use the actual cache config defined in `application.properties`. 

It also adds some tests to ensure that all cache configurations are covered.

There are a few other commits which are refactors of tests classes.

🐡 best reviewed by commit


shubham1g5 (approved):
nice changes 👏

PR #1181, opened 2022-07-14
logging / sentry changes to reduce multiple events for the same thing

snopoke:
Looking at Sentry today I noticed that there are often 2 or more events for the same 'error'. Sentry is configured to create new events for unhandled exceptions as well as for log events at the 'ERROR' level. In some cases we are logging to ERROR and also raising exceptions or manually capturing the Sentry events which is unnecessary.

I have tried to ensure that there is still useful context being output to the console while reducing the Sentry events.

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/GlobalDefaultExceptionHandler.java
+++ b/src/main/java/org/commcare/formplayer/application/GlobalDefaultExceptionHandler.java
@@ -51,9 +51,8 @@ public class GlobalDefaultExceptionHandler {
             NoLocalizedTextException.class})
     @ResponseBody
     public ExceptionResponseBean handleApplicationError(HttpServletRequest request, Exception exception) {
-        log.error("Request: " + request.getRequestURL() + " raised " + exception);
+        log.info("Application error", exception);
         datadog.incrementErrorCounter(Constants.DATADOG_ERRORS_APP_CONFIG, request);
-        FormplayerSentry.captureException(exception, SentryLevel.INFO);
         return getPrettyExceptionResponse(exception, request);
     }
 
@@ -74,8 +73,7 @@ public class GlobalDefaultExceptionHandler {
     @ResponseBody
     public ExceptionResponseBean handleHttpRequestError(HttpServletRequest req, HttpClientErrorException exception) {
         datadog.incrementErrorCounter(Constants.DATADOG_ERRORS_EXTERNAL_REQUEST, req);
-        log.error(String.format("Exception %s making external request %s.", exception, req));
-        Sentry.captureException(exception);
+        log.error(String.format("Exception %s making external request %s.", exception, req), exception);
         return new ExceptionResponseBean(exception.getResponseBodyAsString(), req.getRequestURL().toString());
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/formplayer/application/GlobalDefaultExceptionHandler.java
```diff
@@ -51,9 +51,8 @@ public class GlobalDefaultExceptionHandler {
             NoLocalizedTextException.class})
     @ResponseBody
     public ExceptionResponseBean handleApplicationError(HttpServletRequest request, Exception exception) {
-        log.error("Request: " + request.getRequestURL() + " raised " + exception);
+        log.info("Application error", exception);
         datadog.incrementErrorCounter(Constants.DATADOG_ERRORS_APP_CONFIG, request);
-        FormplayerSentry.captureException(exception, SentryLevel.INFO);
```
snopoke, line 56: This was creating an event before but now will not. I don't think there is value in having application error events in Sentry.

The log event is still captured in Sentry but as breadcrumbs

src/main/java/org/commcare/formplayer/application/GlobalDefaultExceptionHandler.java
```diff
@@ -98,17 +96,16 @@ public RetryExceptionResponseBean handleAsyncRetryException(HttpServletRequest r
     @ExceptionHandler({FormattedApplicationConfigException.class})
     @ResponseBody
     public HTMLExceptionResponseBean handleFormattedApplicationError(HttpServletRequest req, Exception exception) {
-        log.error("Request: " + req.getRequestURL() + " raised " + exception);
+        log.info("Application configuration error", exception);
         datadog.incrementErrorCounter(Constants.DATADOG_ERRORS_APP_CONFIG, req);
-        FormplayerSentry.captureException(exception, SentryLevel.INFO);
```
snopoke, line 103: same as above, no application error events in Sentry

src/main/java/org/commcare/formplayer/services/FormplayerLockRegistry.java
```diff
@@ -88,15 +88,15 @@ public FormplayerReentrantLock obtain(Object lockKey) {
...
+        log.warn(String.format("Thread %s owns expired lock with lock key %s.", ownerThread, lockKey));
         ownerThread.interrupt();
         try {
             ownerThread.join(5000);
         } catch (InterruptedException e) {
             throw new InterruptedRuntimeException(e);
         }
         if (ownerThread.isAlive()) {
-            log.error(String.format(
+            log.warn(String.format(
```
snopoke, line 99: since the exception is captured below I've changed this to a 'warn' which will still output in the console but won't results in a Sentry event on it's own

src/main/java/org/commcare/formplayer/services/InstallService.java
```diff
@@ -93,10 +93,8 @@ public Pair<FormplayerConfigEngine, Boolean> configureApplication(String referen
             installTimer.record();
             return new Pair<>(engine, newInstall);
         } catch (UnresolvedResourceException e) {
-            log.error("Got exception " + e + " while installing reference " + reference + " at path " + sqliteDB.getDatabaseFileForDebugPurposes());
```
snopoke, line 96: The info in these logs is already logged above (reference and database file). Since we are re-throwing the exceptions there's no need to log them here. They will get logged by the main exception handler.

src/main/java/org/commcare/formplayer/application/GlobalDefaultExceptionHandler.java
shubham1g5: Breadcrumbs are only available in sentry when there is an error event right ?

shubham1g5 (approved):
Looks good, thanks for the detailed explanations.

PR #1198, opened 2022-08-03
Add support for Media Upload

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-2189

Implements support for media upload in Formplayer as per [spec](https://docs.google.com/document/d/1R2vMPurumINVhOFiV111vNBEvkjWeGjOxtetzYUWjTA/edit#heading=h.15dhyt7hay0). Involves - 

- Performing Any [validations](https://docs.google.com/document/d/1R2vMPurumINVhOFiV111vNBEvkjWeGjOxtetzYUWjTA/edit#heading=h.6hw0wbs8b3ct) on the file including file size and extension check
- Save the file on the filesystem as outlined [here](https://docs.google.com/document/d/1R2vMPurumINVhOFiV111vNBEvkjWeGjOxtetzYUWjTA/edit#heading=h.rn9s7uqdc20b) 
- Save the answer to the media question in Xform as an unique uuid ‘mediaId’ generated on FP to reference the file.
- Clean up any old files captured for the same media question

https://github.com/dimagi/commcare-core/pull/1144


src/test/java/org/commcare/formplayer/tests/MediaCaptureTest.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    fun testImageCapture_fileSaveAndReplace() {
+        val formResponse = startImageCaptureForm()
+        val responseBean: FormEntryResponseBean
+        try {
+            responseBean = saveImage(formResponse, "media/valid_image.jpg", "valid_image.jpg")
+        } catch (e: Exception) {
+            fail("Unable to save a valid file due to " + e.message)
+        }
```
snopoke, line 36: why not just let the exception get raised?

src/test/java/org/commcare/formplayer/tests/MediaCaptureTest.kt
```diff
@@ -0,0 +1,109 @@
...
+            expectedErr,
+            exception.cause!!.message
+        )
+    }
+
+    @Test
+    fun testImageCapture_FileWithInvalidExtensionErrors() {
+        val formResponse = startImageCaptureForm()
+        val exception = assertThrows<java.lang.Exception> {
+            saveImage(formResponse, "media/invalid_extension.jppg", "invalid_extension.jppg")
```
snopoke, line 80: looks like the filename on disk is actually `invalid_extension.jppg.jpg`

snopoke (commented):
Mostly small feedback but I'm also wondering how the files get cleaned up after form submission (or when incomplete forms get deleted)?

src/test/java/org/commcare/formplayer/tests/MediaCaptureTest.kt
shubham1g5: I was trying to be more assertive here with the primary intent of this test being testing the file save.

shubham1g5:
> Mostly small feedback but I'm also wondering how the files get cleaned up after form submission (or when incomplete forms get deleted)?

@snopoke Since we would be implementing the media metadata table I was thinking we can do an independent cron task that loops over the metadata rows and if the form session no longer exists delete the media from filesystem and subsequently the corresponding metadata record.

src/main/java/org/commcare/formplayer/services/MediaHandler.kt
```diff
@@ -15,28 +18,29 @@ class MediaHandler(val file: MultipartFile) {
...
+            FileUtils.copyFile(file.inputStream, desintationFile)
             return fileId
         } catch (e: IOException) {
             throw IOException("Could not copy file to destination due to " + e.message, e)
         }
     }
 
     private fun validateFile() {
-        if (isUnSupportedFileExtension()) {
+        if (isUnSupportedFileExtension() && isUnsupportedMimeType()) {
```
snopoke, line 43: should this be an `or`

src/main/java/org/commcare/formplayer/services/MediaHandler.kt
```diff
@@ -45,22 +49,24 @@ class MediaHandler(val file: MultipartFile) {
         }
     }
 
-    fun getMediaFilePath(parentDir: String, fileId: String): String {
-        return parentDir + fileId
+    private fun isUnsupportedMimeType(): Boolean {
+        val mimeType: String? = URLConnection.guessContentTypeFromName(file.name)
```
snopoke, line 53: since this is just using the file name I think it's basically a duplicate of the extension check

src/main/java/org/commcare/formplayer/services/MediaHandler.kt
shubham1g5: I mostly tried to replicate what mobile is doing [today](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/utils/FormUploadUtil.java#L423). Seems like we would be ok as long as the mime type is correct even if the extension is not in our list.

src/main/java/org/commcare/formplayer/services/MediaHandler.kt
shubham1g5: That's true in a way although I was assuming that purpose of the mime type check is to cover for the extensions that are not added in our supported extensions list. For eg. even though `webp` is not in our list explicitly, user will still be able to upload the file because it's the valid mimetype. 

Although it seems like we want a more strictier check on content type by checking the byte content of the file and not trust the filename provided by user for the right extension ? I think that should be fine though I think we should still fallback to `guessContentTypeFromName` if we get a null content type from `guessContentTypeFromStream` ?

src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -95,7 +118,25 @@ public FormEntryResponseBean changeLocale(@RequestBody ChangeLocaleRequestBean c
...
+        return saveAnswer(answerQuestionBean, null);
+    }
+
+    @RequestMapping(value = Constants.URL_ANSWER_MEDIA_QUESTION, method = RequestMethod.POST)
+    @UserLock
+    @UserRestore
+    @ConfigureStorageFromSession
+    public FormEntryResponseBean answerMediaQuestion(
+            @RequestParam("file") MultipartFile file,
+            @RequestBody AnswerQuestionRequestBean answerQuestionBean,
```
snopoke, line 132: I think this should be using `@RequestParam("answer")` since the whole body will be the multipart request

src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -95,7 +118,25 @@ public FormEntryResponseBean changeLocale(@RequestBody ChangeLocaleRequestBean c
     @UserRestore
     @ConfigureStorageFromSession
     public FormEntryResponseBean answerQuestion(@RequestBody AnswerQuestionRequestBean answerQuestionBean,
-                                                @CookieValue(name = Constants.POSTGRES_DJANGO_SESSION_ID, required = false) String authToken) throws Exception {
+            @CookieValue(name = Constants.POSTGRES_DJANGO_SESSION_ID, required = false) String authToken)
+            throws Exception {
+        return saveAnswer(answerQuestionBean, null);
+    }
+
+    @RequestMapping(value = Constants.URL_ANSWER_MEDIA_QUESTION, method = RequestMethod.POST)
```
snopoke, line 126: this should probably include a `consumes` arg to set the expected content type.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -1130,6 +1165,15 @@ private <T> T generateMockQuery(ControllerType controllerType,
                 break;
         }
         switch (requestType) {
+            case MULTIPART:
+                result = controller.perform(
+                        multipart(urlPrepend(urlPath))
+                                .file(file)
+                                .contentType(MediaType.APPLICATION_JSON)
+                                .cookie(new Cookie(Constants.POSTGRES_DJANGO_SESSION_ID, "derp"))
+                                .content((String)bean));
```
snopoke, line 1174: I don't think this mocking is correct. I think it should be:
```java
controller.perform(
            multipart(urlPrepend(urlPath))
                .file(file)
                // contentType(...)  don't override content type
                .param(“answer”, (String) bean)  // the JSON data must be a multipart param
                .cookie(new Cookie(Constants.POSTGRES_DJANGO_SESSION_ID, "derp"))
)
```

snopoke (changes requested):
You will also need to modify the auth filters to handle multipart requests:

https://github.com/dimagi/formplayer/blob/63834033d59161ac4c1de9d0326f6104dc07c7f7/src/main/java/org/commcare/formplayer/auth/CommCareSessionAuthFilter.java#L65

https://github.com/dimagi/formplayer/blob/63834033d59161ac4c1de9d0326f6104dc07c7f7/src/main/java/org/commcare/formplayer/auth/HmacAuthFilter.java#L132

(see `HmacAuthTests` and `SessionAuthTests`)

src/main/java/org/commcare/formplayer/application/FormController.java
shubham1g5: Yes @orangejenny  did flag this to me and I am definitely confused between the request format here.  Looks like you are saying there will be 2 parts in this request - The media file and the answer Json which makes sense to me. 

Although I am testing this with JS side of code [here](https://github.com/dimagi/commcare-hq/compare/jls/web-apps-upload#diff-db36924530834745eae7459c406c8e8bdae13c959bc24f2ab93fe9db18a033f0R126) although it seems like JS is currently sending a part for every property of `AnswerQuestionRequestBean`. This is how the payload looks like in chrome console - 

````
answer
------WebKitFormBoundaryUpptPtsoIV8XP82O
Content-Disposition: form-data; name="ix"

0
------WebKitFormBoundaryUpptPtsoIV8XP82O
Content-Disposition: form-data; name="answer"

sas
------WebKitFormBoundaryUpptPtsoIV8XP82O
Content-Disposition: form-data; name="answersToValidate"

[object Object]
------WebKitFormBoundaryUpptPtsoIV8XP82O
Content-Disposition: form-data; name="oneQuestionPerScreen"

false
------WebKitFormBoundaryUpptPtsoIV8XP82O
Content-Disposition: form-data; name="form_context"

undefined
------WebKitFormBoundaryUpptPtsoIV8XP82O
Content-Disposition: form-data; name="domain"

shubham
------WebKitFormBoundaryUpptPtsoIV8XP82O
Content-Disposition: form-data; name="username"

....
````

Is this wrong ? If yes, is it feasible to change it on HQ side so that there are only ever 2 parts as mentioned above.

One other question, it seems like on the JS side we are going to modify all answer requests to be multipart requests instead of just the media answer ones. As such it doesn't seem to make sense to have 2 separate endpoints one for media and one for normal answers ? We will need to support the earlier format to start with as well though I think we can just have one endpoint `answer` and remove `answer_media` ?

Comment (shubham1g5) [reply src/main/java/org/commcare/formplayer/application/FormController.java:132]: > Is this wrong ? If yes, is it feasible to change it on HQ side so that there are only ever 2 parts as mentioned above.

Just saw [this](https://github.com/dimagi/commcare-hq/pull/32026#discussion_r947817490) which resolves this question for me.

**Comment by shubham1g5** [comment]
> You will also need to modify the auth filters to handle multipart requests:

@snopoke I am looking into this though any idea why doesn't the requests from `MediaCaptureTest` doesn't go through auth filters or how do I configure those tests so that they go through auth filters ? I tried adding the `WebSecurityConfig` to the `ContextConfiguration` on `MediaCaptureTest`  but it doesn't seem to do the trick.

**Comment by snopoke** [comment]
> 

I would probably add tests to the existing `HmacAuthTests` and `SessionAuthTests` instead of trying to get your media upload test working with the security configuration. But you can look at those tests to see how they are set up

File: src/test/java/org/commcare/formplayer/tests/BaseTestClass.java [hunk 8]
```diff
@@ -1130,6 +1163,16 @@ private <T> T generateMockQuery(ControllerType controllerType,
                 break;
         }
         switch (requestType) {
+            case MULTIPART:
+                MockMultipartFile answer = new MockMultipartFile("answer", "answer",
+                        "application/json", ((String)bean).getBytes());
+                result = controller.perform(
+                        multipart(urlPrepend(urlPath))
+                                .file(file)
+                                .file(answer)
```
snopoke, line 1172: this shouldn't be a file, it should just be a normal 'part':
```suggestion
                                .part(new MockPart("answer",  ((String)bean).getBytes()))
```

shubham1g5: https://github.com/dimagi/formplayer/pull/1198/commits/36ca13ffb061768da7237041d4e961d6b49f2480

shubham1g5:
@snopoke @orangejenny I am testing this with HQ locally and is now running into an issue `Caused by: org.apache.tomcat.util.http.fileupload.FileUploadException: the request was rejected because no multipart boundary was found` . [SO suggests](https://stackoverflow.com/a/38013585/3811963) that we should not need to explicitly add content type `multipart/form-data` on requests although on [removing this line ](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/cloudcare/static/cloudcare/js/form_entry/web_form_session.js#L130) from HQ code, the `answer_media` request ends up in having a content type `application/json`. Any thoughts if I am missing something here ? If not, I will continue to look for other solutions for `no multipart boundary was found` error

snopoke:
> @snopoke @orangejenny I am testing this with HQ locally and is now running into an issue `Caused by: org.apache.tomcat.util.http.fileupload.FileUploadException: the request was rejected because no multipart boundary was found` . [SO suggests](https://stackoverflow.com/a/38013585/3811963) that we should not need to explicitly add content type `multipart/form-data` on requests although on [removing this line ](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/cloudcare/static/cloudcare/js/form_entry/web_form_session.js#L130) from HQ code, the `answer_media` request ends up in having a content type `application/json`. Any thoughts if I am missing something here ? If not, I will continue to look for other solutions for `no multipart boundary was found` error

What does the request look like in the browser debug tools?

shubham1g5:
This is the `Payload` tab - 

````
"file"; filename="Run2.jpeg"
Content-Type: image/jpeg


------WebKitFormBoundaryvSaUAS2cSSUz4HO3
Content-Disposition: form-data; name="answer"

{"action":"answer_media","ix":"1","answer":"Run2.jpeg","answersToValidate":{},"oneQuestionPerScreen":false,"domain":"shubham","username":"admin@dimagi.com","restoreAs":null,"session-id":"4bbd1556-02ea-43fe-ad1c-a49009070847","session_id":"4bbd1556-02ea-43fe-ad1c-a49009070847","debuggerEnabled":true,"tz_offset_millis":19800000,"tz_from_browser":"Asia/Calcutta"}
------WebKitFormBoundaryvSaUAS2cSSUz4HO3--
````

**Comment by snopoke** [comment]
> > on [removing this line ](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/cloudcare/static/cloudcare/js/form_entry/web_form_session.js#L130)from HQ code, the answer_media request ends up in having a content type application/json
> 
> Little correction that the content type comes as `application/x-www-form-urlencoded; charset=UTF-8` on removing `multipart/form-data` from HQ

I got this working locally by setting `contentType=false` and removing the `dataType` param from the ajax request.

I'm still getting an error from Formplayer because of the auth filters not being able to handle multipart requests.

**Comment by shubham1g5** [comment]
Thanks Simon, that was super helpful. On the multipart auth handling the [code](https://github.com/dimagi/formplayer/blob/saveMediaAttachment/src/main/java/org/commcare/formplayer/util/RequestUtils.java#L36) current checks for a exact match for content type `multipart/form-data` while the request from HQ seems to have a content type like `multipart/form-data; boundary=----WebKitFormBoundaryoveAfsS5eV2xHjBF` . Is it ok to just check for content type starting with `multipart/form-data` here ?

Secondly, [this code](https://github.com/dimagi/formplayer/blob/saveMediaAttachment/src/main/java/org/commcare/formplayer/util/RequestUtils.java#L38) looks for a request part with content type `application/json` to get the auth info from.  Though currently there is no content type set on the `answer` request part from HQ. Is it possible to add a content type to request parts on HQ side or should we hardcode the FP side to look for a part with name `answer` ?

**Comment by snopoke** [comment]
> Thanks Simon, that was super helpful. On the multipart auth handling the [code](https://github.com/dimagi/formplayer/blob/saveMediaAttachment/src/main/java/org/commcare/formplayer/util/RequestUtils.java#L36) currently checks for a exact match for content type `multipart/form-data` while the request from HQ seems to have a content type like `multipart/form-data; boundary=----WebKitFormBoundaryoveAfsS5eV2xHjBF` . Is it ok to just check for content type starting with `multipart/form-data` here ?

Yes that's fine - you could also split it using `;` and look at the first part.

> Secondly, [this code](https://github.com/dimagi/formplayer/blob/saveMediaAttachment/src/main/java/org/commcare/formplayer/util/RequestUtils.java#L38) looks for a request part with content type `application/json` to get the auth info from. Though currently there is no content type set on the `answer` request part from HQ. Is it possible to add a content type to request parts on HQ side or should we hardcode the FP side to look for a part with name `answer` ?

I've made a change that adds the content type to the request: https://github.com/dimagi/commcare-hq/pull/32063

File: src/test/java/org/commcare/formplayer/auth/HmacAuthTests.java [hunk 9]
```diff
@@ -115,6 +118,14 @@ public void testFullAuthEndpoint_WithHmacAuthAndUserDetails_Succeeds() throws Ex
         this.testEndpoint(builder, status().isOk());
     }
 
+    @Test
+    public void testFullAuthMultipartEndpoint_WithHmacAuthAndUserDetails_Succeeds() throws Exception {
+        String hmac = RequestUtils.getHmac(formplayerAuthKey, FULL_AUTH_BODY);
```
snopoke, line 123: for multipart requests the HMAC should be calculated on the full body, not just the JSON part. I think you will have to re-factor the `RequestUtils` so that you can get both the full body and the JSON part:

```java
RequestUtils.getBody(request)  // returns the full body content and does't care about the content type
RequestUtils.getJsonBodyPart(request)  // returns the JSON request part which might be the full body or just a part of it
```

shubham1g5:
Noting that I had to [reorder args](https://github.com/dimagi/formplayer/pull/1198/commits/bb00618456101c19bcf284a038a02432a52f0c5c) for answer media controller method as currently Aspects are [programmed](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/aspects/UserRestoreAspect.java#L59) to expect `AuthenticatedRequestBean` and hq auth token(sessionid cookie) to  be the first two arguments.

src/test/java/org/commcare/formplayer/auth/HmacAuthTests.java
```diff
@@ -120,9 +122,10 @@ public void testFullAuthEndpoint_WithHmacAuthAndUserDetails_Succeeds() throws Ex
 
     @Test
     public void testFullAuthMultipartEndpoint_WithHmacAuthAndUserDetails_Succeeds() throws Exception {
-        String hmac = RequestUtils.getHmac(formplayerAuthKey, FULL_AUTH_BODY);
-        MockHttpServletRequestBuilder builder = getMultipartRequestBuilder(getClass(), FULL_AUTH_BODY)
-                .header(Constants.HMAC_HEADER, hmac);
+        MockHttpServletRequestBuilder builder = getMultipartRequestBuilder(getClass(), FULL_AUTH_BODY);
+        MockHttpServletRequest request = builder.buildRequest(new MockServletContext());
+        String hmac = RequestUtils.getHmac(formplayerAuthKey, RequestUtils.getBody(request.getInputStream()));
```
snopoke, line 127: This looked a bit suspicious to me so I dug into it and found that `MockHttpServletRequest.getInputStream()` is returning an empty input stream.

I tried to figure out how we could do this better but there doesn't seem to be an easy way. I was able to hack it together by creating a real mutlpart body and adding that to the mock request as well as the individual parts:

```java
        MultipartBodyBuilder builder = new MultipartBodyBuilder();
        builder.part(PART_ANSWER, FULL_AUTH_BODY).contentType(MediaType.APPLICATION_JSON);

        Resource image = new ClassPathResource("media/valid_image.jpg");
        builder.part(PART_FILE, image).contentType(MediaType.IMAGE_JPEG);

        MultiValueMap<String, HttpEntity<?>> multipartBody = builder.build();
        RequestCallback callback = restTemplate.httpEntityCallback(multipartBody);
        MockClientHttpRequest request = new MockClientHttpRequest();
        callback.doWithRequest(request);

        String multiPartBody = request.getBodyAsString();

        MockPart answer = new MockPart(PART_ANSWER, FULL_AUTH_BODY.getBytes());
        answer.getHeaders().setContentType(MediaType.APPLICATION_JSON);
        MockHttpServletRequestBuilder request1 = multipart(String.format(String.format("/%s", "mock_mulipart_request")))
                .file(
                        new MockMultipartFile(
                                PART_FILE,
                                "valid_image.jpg",
                                MediaType.IMAGE_JPEG_VALUE,
                                image.getInputStream())
                )
                .part(answer)
                .contentType(request.getHeaders().get("Content-Type").get(0))
                .content(multiPartBody)
                .with(SecurityMockMvcRequestPostProcessors.csrf());

        String hmac = RequestUtils.getHmac(formplayerAuthKey, multiPartBody);
        request1.header(Constants.HMAC_HEADER, hmac);
        this.testEndpoint(request1, status().isOk());
```

I really don't like this approach though. I'm not sure what the best thing to do here. On the one hand we could disable multipart requests with HMAC auth (we don't currently have a use case for it) or we could abandon the idea of testing it properly for now.

shubham1g5: Ahh that's interesting! Thanks for digging deep into this. I would support disabling multipart requests with HMAC auth  if we don't see a plausible use case for it. I am assuming it would be comparatively low effort to do so and a cleaner approach amongst all.

snopoke: > Ahh that's interesting! Thanks for digging deep into this. I would support disabling multipart requests with HMAC auth if we don't see a plausible use case for it. I am assuming it would be comparatively low effort to do so and a cleaner approach amongst all.

👍🏻 HMAC auth should only be used for SMS forms which don't support MM questions

shubham1g5: Sounds promising. Do we need to make a HQ change as well for it ? Or Just changing FP to bypass Hmac auth on multipart requests should be enough ?

src/main/java/org/commcare/formplayer/services/MediaValidator.kt
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Utility functions to validate media attachments in a form
+ */
+object MediaValidator {
+
+    // 3 MB size limit
+    private const val MAX_BYTES_PER_ATTACHMENT = (3 * 1048576 - 1024).toLong()
+    private val SUPPORTED_FILE_EXTS = ImmutableList.of(
+        ".jpg",
```
orangejenny, line 15: why does just this one have a dot?

Robert-Costello: Maybe typo? I don't think it needs the dot. I've tested it without as part of frontend file type validation and it works fine with or without.

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -84,14 +105,17 @@ public class FormSubmissionController extends AbstractBaseController {
 
     private final Log log = LogFactory.getLog(FormSubmissionController.class);
 
+    @Value("${form.submit.max_attachments}")
```
shubham1g5, line 108: @Robert-Costello Requesting one more change here. We should rename this property here to `formplayer.form.submit.max_attachments` to have it namespaced as `formplayer`. I missed out on this feedback from Simon earlier.

shubham1g5:
@snopoke This looks good to be merged to me. Would you be able to have a final look here ?

Also I do imagine a certain impact on disk space as more projects start to use this feature. This feature is behind a [feature flag](https://www.commcarehq.org/hq/flags/edit/web_apps_upload_questions/) currently which I think is a good approach for now to limit the # of projects using this feature. But we probably want to start a separate conversation with SAAS if they see value in GAing this and would be willing to monitor any impact on disk space due to this feature. @orangejenny

src/test/java/org/commcare/formplayer/auth/AuthTestUtils.java
```diff
@@ -0,0 +1,49 @@
...
+     * Use the 'answer_media' endpoint for 'full auth' with multipart
+     */
+    public static MockHttpServletRequestBuilder getMultipartRequestBuilder(Class c, String body)
+            throws IOException {
+        InputStream fis = FileUtils.getFileStream(c, "media/valid_image.jpg");
+        MockMultipartFile file = new MockMultipartFile(PART_FILE, "valid_image.jpg", MediaType.IMAGE_JPEG_VALUE,
+                fis);
+        MockPart answer = new MockPart(PART_ANSWER, body.getBytes());
+        answer.getHeaders().setContentType(MediaType.APPLICATION_JSON);
+        return multipart(String.format("/%s", "mock_mulipart_request"))
```
snopoke, line 44: Is the string format here doing anything?

src/test/java/org/commcare/formplayer/auth/AuthTestUtils.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public static MockHttpServletRequestBuilder getFullAuthRequestBuilder(String body) {
+        return post(String.format("/%s", Constants.URL_CLEAR_USER_DATA))
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(body)
+                .with(SecurityMockMvcRequestPostProcessors.csrf());
+    }
+
+    /**
+     * Use the 'answer_media' endpoint for 'full auth' with multipart
```
snopoke, line 35: I think this comment is wrong. This specifically creates a request for the `mock_mulipart_request` endpoint right?

src/test/java/org/commcare/formplayer/junit/MockTimezoneProvider.java
```diff
@@ -5,6 +5,7 @@
 /**
  * Mock timezone provider
  */
+
```
snopoke, line 8: Not for future reference, making unnecessary changes in the PR makes it harder to review.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -946,6 +955,16 @@ private <T> T generateMockQuery(ControllerType controllerType,
...
         switch (requestType) {
+            case MULTIPART:
+                MockPart answer = new MockPart(PART_ANSWER, ((String)bean).getBytes());
+                answer.getHeaders().setContentType(MediaType.APPLICATION_JSON);
+                result = controller.perform(
+                        multipart(urlPrepend(urlPath))
+                                .file(file)
+                                .part(answer)
+                                .cookie(new Cookie(Constants.POSTGRES_DJANGO_SESSION_ID, "derp")));
+                break;
```
snopoke, line 966: Is this change still necessary?

src/test/java/org/commcare/formplayer/auth/AuthTestUtils.java
shubham1g5: Not really, I think I just copied it from the similar `getFullAuthRequestBuilder` present in this class. Do you think it's better to use String concatenation instead of `.format` ?

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -248,38 +276,100 @@ private SubmitResponseBean processFormXml(FormSubmissionContext context) throws
...
+        Path mediaDirPath = formSession.getMediaDirectoryPath(restoreFactory.getDomain(),
+                restoreFactory.getUsername(), restoreFactory.getAsUsername(), storageFactory.getAppId());
+        File mediaFile = mediaDirPath.toFile();
+        if (mediaFile.exists()) {
+            File[] files = Objects.requireNonNull(mediaDirPath.toFile().listFiles());
+            if (files.length > maxAttachmentsPerForm) {
+                failWithError("form.upload.attachments.limit.exceeded", maxAttachmentsPerForm.toString());
+            }
+            for (File file : files) {
+                validateFile(file);
```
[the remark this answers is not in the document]
shubham1g5: We are validating max no. of files per form and then each file size separately, so I think validating total upload size would be redundant.

orangejenny:
@shubham1g5 Agreed that disk space will be something to consider, possibly even if the feature flag doesn't get GAed soon, as we've had some USH requirements conversations involving relatively large files (think 50-page documents).

src/test/java/org/commcare/formplayer/auth/AuthTestUtils.java
snopoke: Oh right, I see this is adding the '/'. This is fine then.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
shubham1g5: https://github.com/dimagi/formplayer/pull/1198/commits/324daa004a2120df5492e165150960b84c953ecf

src/main/java/org/commcare/formplayer/services/MediaHandler.kt
```diff
@@ -0,0 +1,103 @@
...
+            asUser,
+            domain,
+            appId
+        )
+        mediaMetaDataService.saveMediaMetaData(mediaMetaData)
+
+        return fileIdWithExt
+    }
+
+    private fun validateFile() {
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/formplayer/pull/1198/commits/916c66aec6eb351e1077828415e4a8ee20766802

snopoke (approved):

PR #1206, opened 2022-08-12
Remove seemingly out of place files uatbc{,2}.ccz

dannyroberts:
## Technical Summary
I made this PR mainly to test the checklist check https://dimagi-dev.atlassian.net/browse/SAAS-13858.

### Safety story
These appear to be unreferenced, but if they're used for anything at all, I would imagine it would be for tests, in which case the tests would fail.

### Rollback instructions

### Review


joxl (approved):
Thanks!

PR #1209, opened 2022-08-16
Form submission with Media

shubham1g5:
## Product Description

Adds support for submitting a form media to HQ as part of form submission. 

## Technical Summary

Implements form submission request as a multipart/form-data request using specs [here](https://confluence.dimagi.com/display/commcarepublic/Submission+API#SubmissionAPI-FormSubmissionAPI)

## Safety Assurance

This will be QAed to check for any regressions

### Safety story

- Tested Locally
- Planning to have it go through QA

### Automated test coverage

- Added a unit test to test for correct request params
- We are lacking any integration tests in this area and I will be relying on QA automated scripts plus manual testing for them

### QA Plan

To be added

- Will need to update submodule to include https://github.com/dimagi/commcare-core/pull/1147 before merging

### Rollback instructions

### Review

cross-request: https://github.com/dimagi/commcare-core/pull/1147


src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -252,6 +263,44 @@ private SubmitResponseBean processFormXml(FormSubmissionContext context) throws
...
+        File mediaFile = mediaDirPath.toFile();
+        if(mediaFile.exists()) {
+            for (File file : Objects.requireNonNull(mediaDirPath.toFile().listFiles())) {
+                String contentType = FileUtils.getContentType(file);
+                if (!StringUtils.hasText(contentType)) {
+                    contentType = "application/octet-stream";
+                }
+                HttpEntity<Object> filePart = createFilePart(file.getName(), file.getName(),
+                        file, contentType);
+                body.add(file.getName(), filePart);
```
snopoke, line 281: I think there should be some additional validation before we submit these files in case there are other files in this directory for some reason. I'm thinking:
* max number of attachments
* max attachment size

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -252,6 +263,44 @@ private SubmitResponseBean processFormXml(FormSubmissionContext context) throws
...
+        MultiValueMap<String, Object> body = new LinkedMultiValueMap<>();
+
+        // Add any media files associated with the form
+        Path mediaDirPath = formSession.getMediaDirectoryPath(restoreFactory.getDomain(),
+                restoreFactory.getUsername(), restoreFactory.getAsUsername(), storageFactory.getAppId());
+        File mediaFile = mediaDirPath.toFile();
+        if(mediaFile.exists()) {
+            for (File file : Objects.requireNonNull(mediaDirPath.toFile().listFiles())) {
+                String contentType = FileUtils.getContentType(file);
+                if (!StringUtils.hasText(contentType)) {
```
snopoke, line 276: how would this not be able to get the content type if we are only writing files to this directory that match our content type filtering?

src/main/java/org/commcare/formplayer/services/SubmitService.java
```diff
@@ -27,11 +27,11 @@ public class SubmitService {
 
     private CategoryTimingHelper.RecordingTimer submitTimer;
 
-    public String submitForm(String formXml, String submitUrl) {
+    public String submitForm(Object body, String submitUrl) {
```
snopoke, line 30: Should this not be `MultiValueMap`? I don't see this function used anywhere else.

src/test/java/org/commcare/formplayer/tests/SavedFormDefTest.java
```diff
@@ -86,7 +88,7 @@ public void testIrrelevantDataIsNotIncludedInSubmission() throws Exception {
                 "requests/submit/submit_hidden_value_form.json", sessionId);
         assertEquals("success", submitResponseBean.getStatus());
 
-        Mockito.verify(this.submitServiceMock).submitForm(formSession.getInstanceXml(false),
-                formSession.getPostUrl());
+        Mockito.verify(this.submitServiceMock).submitForm(any(),
```
snopoke, line 91: This change greatly reduces the value of this test. I think the whole point of the test was to check that the XML submitted matched expectations which it no longer does. I think you'll need to use `argThat` with a custom argument matcher to validate the XML in the mulipart body.

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
shubham1g5: Sounds good, Would `10` be a good number for max no. of attachments ? Or we want to set it to much higher like `100` which more strongly signifies that something has gone wrong with the media directory.

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
shubham1g5: We write to the directory when either one of extension match or MIME type match is successful. So it is potentially feasible for us to write a file for which content type methods return empty or null.

src/main/java/org/commcare/formplayer/services/SubmitService.java
shubham1g5: https://github.com/dimagi/formplayer/pull/1209/commits/23a5881c68c9c4cb9b844c02506e6085caff0f41

src/test/java/org/commcare/formplayer/tests/SavedFormDefTest.java
shubham1g5: Thanks for flagging, decided to use an argument captor instead - https://github.com/dimagi/formplayer/pull/1209/commits/ca4949eaab8ab1b0a827e7d7f0dbcc548f891c37

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
snopoke: We should make it configurable via the properties file with a sane default. 10 seems like it could be too low - maybe 50? Or how about limiting on both number and total size - 50 attachments or total size < 50MB?

@orangejenny any thoughts about the max number of attachments to allow per form submission?

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
snopoke: But the file utils method falls back to extension anyway right?

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
shubham1g5: 50 sounds like a good threshold to go with. We do a size check per file while saving media, should we go with the same size check per file while submission as well so that it's more consistent ?

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
shubham1g5: It does. Although JVM mapping for extension to mime types doesn't seem to be [very reliable](https://bugs.openjdk.org/browse/JDK-8273655). So if we start supporting `.doc` with Java8, we would get `null` result from `guessContentTypeFromName`.  

It might make more sense though to only worry about these extensions if and when we start supporting them and just throw an exception at the moment for these cases ?

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
shubham1g5: Added a threshold of 50 and replicated normal file validations as well to protect against any malicious files - https://github.com/dimagi/formplayer/pull/1209/commits/e6bca3951eb723ab90c69072db79961dd2f22583

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
shubham1g5: I am keeping it as it as for now as mobile also seems to be using the [same default content type](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/utils/FormUploadUtil.java#L348). Happy to revisit if there are any objections though.

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -252,6 +274,65 @@ private SubmitResponseBean processFormXml(FormSubmissionContext context) throws
...
+
+        // Add the form xml
+        HttpEntity<Object> xmlPart = createFilePart("xml_submission_file", "xml_submission_file.xml",
+                formSession.getInstanceXml(false), "text/xml");
+        body.add("xml_submission_file", xmlPart);
+        return body;
+    }
+
+    private void validateFile(File file) throws FileNotFoundException {
+        if (isUnSupportedFileExtension(file.getName()) && isUnsupportedMimeType(new FileInputStream(file),
```
snopoke, line 311: should this be 'or'?:
```suggestion
        if (isUnSupportedFileExtension(file.getName()) || isUnsupportedMimeType(new FileInputStream(file),
```

src/test/resources/formplayer_translatable_strings.txt
```diff
@@ -4,3 +4,6 @@ ethiopian_months=Mäskäräm,T’ïk’ïmt,Hïdar,Tahsas,T’ïr,Yäkatit,Mäga
 nepali_months=Baishakh,Jestha,Ashadh,Shrawan,Bhadra,Ashwin,Kartik,Mangsir,Poush,Magh,Falgun,Chaitra
 form.attachment.invalid=Invalid filetype choosen. Please select a valid multimedia file.
 file.oversize.error.message=Selected file exceeds the file size limit of 3 MB. Please select a file whose size is less than 3 MB.
+form.upload.attachments.threshold.error.message=More than ${0} attachments found for this form.
+form.upload.attachment.invalid=Invalid file ${0} attached with this form.
+form.upload.attachment.oversize=File ${0} is more than the maximum limit of 3 MB.
```
snopoke, line 9: is it necessary to duplicate these here? If we remove this file it should pick up the one from `src/main/resources`

snopoke (commented):
Looks good, a few small questions. Also ... stickler...

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
shubham1g5: I think we would want to accept files that has either supported extension or mime type (https://github.com/dimagi/formplayer/blob/saveMediaAttachment/src/main/java/org/commcare/formplayer/services/MediaHandler.kt#L43) , so we only error out when both are invalid.

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -85,6 +104,9 @@ public class FormSubmissionController extends AbstractBaseController {
 
     private final Log log = LogFactory.getLog(FormSubmissionController.class);
 
+    @Value("${max.attachments.per.form}")
+    private Integer maxAttachmentsPerForm;
```
shubham1g5, line 108: @snopoke I am struggling with an issue with this property not bein initialised in the `MediaCaptureTest` . I tried adding annotations like `@TestPropertySource(properties = ["max.attachments.per.form=1"])` or `@TestPropertySource("classpath:application.properties")` to `MediaCaptureTest` , though this property remains `null`. Would you be able to have a look at it whenever you have some space.

snopoke: `@TestPropertySource("classpath:application.properties")` is what I've used before. You might need to add ` @ContextConfiguration` to the class as well.

shubham1g5: thanks I did try it but it doesn't seem to work. My suspicion is somehow `@WebMvcTest` is interferring with setting the property but I am out of ideas on how to debug it.

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -252,38 +275,100 @@ private SubmitResponseBean processFormXml(FormSubmissionContext context) throws
...
+        MultiValueMap<String, HttpEntity<Object>> body = new LinkedMultiValueMap<>();
+
+        // Add any media files associated with the form
+        Path mediaDirPath = formSession.getMediaDirectoryPath(restoreFactory.getDomain(),
+                restoreFactory.getUsername(), restoreFactory.getAsUsername(), storageFactory.getAppId());
+        File mediaFile = mediaDirPath.toFile();
+        if (mediaFile.exists()) {
+            File[] files = Objects.requireNonNull(mediaDirPath.toFile().listFiles());
+            if (files.length > maxAttachmentsPerForm) {
+                failWithError("form.upload.attachments.threshold.error.message", maxAttachmentsPerForm.toString());
```
snopoke, line 289: nit: update locale message to be more concise and consistent with the other error messages:
```suggestion
                failWithError("form.upload.attachments.limit_exceeded", maxAttachmentsPerForm.toString());
```

src/main/java/org/commcare/formplayer/application/FormSubmissionController.java
```diff
@@ -252,38 +275,100 @@ private SubmitResponseBean processFormXml(FormSubmissionContext context) throws
...
+                file.getName())) {
+            failWithError("form.upload.attachment.invalid", file.getName());
+        } else if (isFileTooLarge(file.length())) {
+            failWithError("form.upload.attachment.oversize", file.getName());
+        }
+    }
+
+    private void failWithError(String localeKey, String... args) {
+        String attachmentsThresholdError = Localization.get(localeKey, args);
+        throw new RuntimeException(attachmentsThresholdError);
```
snopoke, line 323: Can we use a custom exception class rather than `RuntimeException`. How about `FormAttachmentException`?

snopoke (approved):
Small nits but nothing blocking

shubham1g5:
Going to merge this in the main feature branch for the sake of completion, happy to take more feedback on the parent branch if any.

PR #1216, opened 2022-08-30
update readme for mac specific issue

Robert-Costello:
## Technical Summary

Mac users running formplayer outside of Docker (all mac users?) need to remove references to `micrometer` to run formplayer locally. I was reminded of this when helping new devs setup their environments. 

## Safety Assurance
Just a readme change
### QA Plan

no QA
### Migrations

### Rollback instructions

### Review

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ Download submodule dependencies
 
     $ git submodule update --init --recursive
 
+Mac users will need to comment out references to micrometer in the following files:
+
+    build.gradle
+
+    src/main/java/org/commcare/formplayer/application/WebAppContext.java
+
 To make properties file:
 
     $ cp config/application.example.properties config/application.properties  # Update properties as necessary (the defaults are fine for running locally)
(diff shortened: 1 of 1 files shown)


orangejenny:
Could this be an OS version issue? I'm on Monterey (12.5) and don't need to do this.

stephherbers:
hm I also don't have to do this-- interesting. Is it a M1 chip issue?

Robert-Costello:
Interesting, thanks for the input. This was a fix that Sravan came up with last year so it could be an OS version thing. I'm on an intel machine running Big Sur 11.3, so I don't think it's exclusive to M1 machines. I'll check with Evan and Jonathan to see what version they're running.

Robert-Costello:
Evan is on Monterey 12.5.1

orangejenny:
Weird. How about just updating the docs to say "Mac users **may** need..."?

Robert-Costello:
Thanks @AddisonDunn. That deepens the mystery, but is definitely helpful info.

orangejenny (approved):

shubham1g5:
This might also be related to the combination of M1 chip and the Java variant they are using. Do we know if they are using a Java distribution other than OpenJDK ?

PR #1218, opened 2022-08-31
add state hash to restore URL

snopoke:
## Technical Summary
This adds the state hash to the restore URL. This should allow better validation that HQ and formplayer are aligned on the users case DB state.

Initially any errors will only be logged to avoid breaking changes (see https://github.com/dimagi/commcare-hq/pull/32089).

The intention is to surface sync bugs sooner so that they can be fixed.

### Safety story
Once the HQ PR is deployed this should be safe since a state hash mismatch will still not result in an error.

The only question I have is related to the performance of computing the state hash. Is that something we should be concerned about? Perhaps @shubham1g5 @ctsims have more insight.

### Automated test coverage
Tests added to make sure a blank state hash is not included and a non-blank one is included correctly

### QA Plan
None

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -7,6 +7,7 @@ import io.sentry.SentryLevel;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.commcare.cases.util.CaseDBUtils;
 import org.commcare.cases.util.InvalidCaseGraphException;
 import org.commcare.core.parse.ParseUtils;
 import org.commcare.formplayer.api.process.FormRecordProcessorHelper;
@@ -663,6 +664,11 @@ public class RestoreFactory {
         if (syncToken != null && !"".equals(syncToken)) {
             params.put("since", syncToken);
         }
+        String caseStateHash = getCaseDbHash();
+        if (!caseStateHash.isEmpty()) {
+            params.put("state", caseStateHash);
+        }
+
         if (asUsername != null) {
             String asUserParam = asUsername;
             if (!asUsername.contains("@")) {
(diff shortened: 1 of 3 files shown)


shubham1g5 (approved):
Looks good to me, I have not seen computing hash causing issues on mobile. Although FP will be computing the hash for large number of users, it would be spread out across time so I don't think it should affect performance significantly.

ctsims:
Since we shifted to [index-only iterators](https://github.com/dimagi/commcare-android/blob/35627e56cbe693cf8951e0315ca2f33057b3fc5c/app/src/org/commcare/engine/cases/CaseUtils.java#L54) in the db, this hasn't been much of a meaningful hit on Android, but it'll be a bit slower on the [native commcare-core library](https://github.com/dimagi/commcare-core/blob/d9066ea561f4ce85717b25263f9ddad695c98b29/src/main/java/org/commcare/cases/util/CaseDBUtils.java#L13) because it's doing full record trawls. 

I imagine this may have some performance implications as a result, but it's not always trivial to guess what operations which are slow on Android are quite fast on FP. My intuition is that this kind of linear walk will be a meaningful but not massive performance hit, which means it might be worth a second look. It's the kind of "ounces are pounds, and pounds are pain" kind of incremental performance hit that can add up over time.

snopoke:
> Since we shifted to [index-only iterators](https://github.com/dimagi/commcare-android/blob/35627e56cbe693cf8951e0315ca2f33057b3fc5c/app/src/org/commcare/engine/cases/CaseUtils.java#L54) in the db, this hasn't been much of a meaningful hit on Android, but it'll be a bit slower on the [native commcare-core library](https://github.com/dimagi/commcare-core/blob/d9066ea561f4ce85717b25263f9ddad695c98b29/src/main/java/org/commcare/cases/util/CaseDBUtils.java#L13) because it's doing full record trawls.

The mobile one definitely seems more efficient. Is there a reason we can't do the same in core?

> I imagine this may have some performance implications as a result, but it's not always trivial to guess what operations which are slow on Android are quite fast on FP. My intuition is that this kind of linear walk will be a meaningful but not massive performance hit, which means it might be worth a second look. It's the kind of "ounces are pounds, and pounds are pain" kind of incremental performance hit that can add up over time.

:+1:

ctsims:
> The mobile one definitely seems more efficient. Is there a reason we can't do the same in core?

The Android build has one or two storage layer optimizations which aren't pushed down to core due to coupling with the underlying database. Index-only iterators might be one of those, but I can't quite recall with confidence whether that's the case.

shubham1g5 (approved):

PR #1222, opened 2022-09-13
Change #,# delimiter in multi-select case search fields

orangejenny:
## Product Description
https://dimagi-dev.atlassian.net/browse/USH-2382

## Technical Description
Interdependent with https://github.com/dimagi/commcare-core/pull/1149

### Safety story
There's reasonable test coverage of this logic.

### Automated test coverage
In PR

### QA Plan
Requesting QA

This needs to be deployed at the same time as  and https://github.com/dimagi/commcare-hq/pull/32139

### Rollback instructions
If this is rolled back, https://github.com/dimagi/commcare-hq/pull/32139 also needs to be rolled back.

This isn't ideal, but I'm not terribly worried about this set of changes needing to be rolled back.

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
@@ -8,6 +8,7 @@ import org.commcare.session.RemoteQuerySessionManager;
 import org.commcare.suite.model.QueryPrompt;
 import org.commcare.util.screen.QueryScreen;
 import org.javarosa.core.model.utils.ItemSetUtils;
+import org.javarosa.core.util.DataUtil;
 import org.javarosa.core.util.OrderedHashtable;
 
 import java.util.ArrayList;
@@ -63,8 +64,7 @@ public class QueryResponseBean extends MenuBean {
                     }
                 }
                 if (indicesForSelectedChoices.size() > 0) {
-                    currentAnswer = String.join(RemoteQuerySessionManager.ANSWER_DELIMITER,
-                            indicesForSelectedChoices);
+                    currentAnswer = DataUtil.joinWithSpaces(indicesForSelectedChoices.toArray(new String[0]));
                 }
                 choiceLabels = ItemSetUtils.getChoiceLabels(queryPromptItem.getItemsetBinding());
             }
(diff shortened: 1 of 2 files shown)


snopoke (approved):

src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
```diff
@@ -63,8 +64,7 @@ public QueryResponseBean(QueryScreen queryScreen, SessionWrapper session) {
                     }
                 }
                 if (indicesForSelectedChoices.size() > 0) {
-                    currentAnswer = String.join(RemoteQuerySessionManager.ANSWER_DELIMITER,
-                            indicesForSelectedChoices);
+                    currentAnswer = DataUtil.joinWithSpaces(indicesForSelectedChoices.toArray(new String[0]));
```
shubham1g5, line 67: we should initialise the array with the exact size like - 
```suggestion
                    currentAnswer = DataUtil.joinWithSpaces(indicesForSelectedChoices.toArray(new String[indicesForSelectedChoices.size()]));
```

shubham1g5 (changes requested):

src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
orangejenny: b744e383

shubham1g5 (approved):

src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java
```diff
@@ -63,8 +64,8 @@ public QueryResponseBean(QueryScreen queryScreen, SessionWrapper session) {
                     }
                 }
                 if (indicesForSelectedChoices.size() > 0) {
-                    currentAnswer = String.join(RemoteQuerySessionManager.ANSWER_DELIMITER,
```
shubham1g5, line 66: We would need this to be backward compatible as well to start with to remove the dependency between HQ and Formplayer deploy.

shubham1g5 (changes requested):

orangejenny:
Closing, see discussion on ticket.

PR #1227, opened 2022-09-23
Initial updates to make tests follow Spring testing patterns

snopoke:
Easiest reviewed by commit

I've put most of this code in a separate PR so that it can be merged independently of the MM submissions changes: https://github.com/dimagi/formplayer/pull/1228

## Technical Summary
Our current controller tests mock the controller manually which does work but has some limitations (e.g. properties don't work https://github.com/dimagi/formplayer/pull/1209#discussion_r965834890).

This PR makes a start to bring out tests inline with Spring testing patterns.

The primary change is to stop mocking the controllers and allow Spring to set them up:

```
@WebMvcTest
@ContextConfiguration(classes = [TestContext::class, CacheConfiguration::class])
@Import(FormController::class, FormSubmissionController::class)
class MediaCaptureTest {
```

This test uses the `@Import` annotation to bring the controllers under test into the web app context and have them configured by Sprint (technically you should be able to pass the controllers directly to the `@WebMvcTest` annotation but for some reason that isn't working).

The dependencies for the controllers (e.g. FormSessionService) are provided by the context classes passed to `@ContextConfiguration`.

With this setup all that's needed is configuring other mocks like `RestoreFactory`.

**Configuring mocks**
To handle this I've tried a different approach rather than having a very large base class that does all the setup. I've split the mock setup into separate classes for each mock and made them into Junit5 extensions which can be used to annotate the test class.

This allows us to keep the mock setup compartmentalised and also makes it possible to use only those that are needed for a test.

For convenience we can bundle a number of extension together using an annotation which I've done here (see `FormSessionTest`).

I have not attempted to change any other tests but I have replaced relevant parts of the `BaseTestClass` with the new mock setup classes to reduce code duplication.

**Making requests**
The other thing that I've done is to extract the code which makes the mock MVC requests into their own classes instead of using the generic methods in the base class which are complex and hard to follow. See the `org.commcare.formplayer.junit.request` package.

### Safety story
Only affects tests

### QA Plan
If the build passes

### Rollback instructions

### Review

### DIFF
--- a/src/test/java/org/commcare/formplayer/junit/FormDefSessionServiceExtension.kt
+++ b/src/test/java/org/commcare/formplayer/junit/FormDefSessionServiceExtension.kt
@@ -16,7 +16,6 @@ import org.springframework.cache.CacheManager
 import org.springframework.test.context.junit.jupiter.SpringExtension
 import org.springframework.test.util.ReflectionTestUtils
 import java.io.IOException
-import java.util.*
 
 /**
  * Junit extension that configures the mock FormDefinitionService
@@ -33,7 +32,8 @@ class FormDefSessionServiceExtension : BeforeAllCallback, BeforeEachCallback {
     private var currentFormDefinitionId = 1L
 
     override fun beforeAll(context: ExtensionContext) {
-        formDefinitionService = SpringExtension.getApplicationContext(context).getBean(FormDefinitionService::class.java)
+        formDefinitionService =
+            SpringExtension.getApplicationContext(context).getBean(FormDefinitionService::class.java)
         cacheManager = SpringExtension.getApplicationContext(context).getBean(CacheManager::class.java)
 
         // manually wire this in. The autowiring doesn't work here since we've made it a mock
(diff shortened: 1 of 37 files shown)


src/test/java/org/commcare/formplayer/junit/FormSessionServiceExtension.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.test.util.ReflectionTestUtils
+import java.util.*
+
+/**
+ * Junit extension that configures the mock FormDefinitionService
+ *
+ * Setup mocking for the FormSessionService that allows saving and retrieving sessions.
+ * The 'persisted' sessions are cleared at the start of each test.
+ */
+class FormSessionServiceExtension : BeforeAllCallback, BeforeEachCallback {
```
snopoke, line 22: mostly copied from https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/BaseTestClass.java#L325-L362

src/test/java/org/commcare/formplayer/junit/FormDefSessionServiceExtension.kt
```diff
@@ -0,0 +1,95 @@
...
+import java.io.IOException
+import java.util.*
+
+/**
+ * Junit extension that configures the mock FormDefinitionService
+ *
+ * Setup mocking for the FormDefinitionService that allows saving and retrieving form definitions.
+ * The 'persisted' definitions are cleared at the start of each test.
+ */
+class FormDefSessionServiceExtension : BeforeAllCallback, BeforeEachCallback {
```
snopoke, line 27: mostly copied from https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/BaseTestClass.java#L368-L410

src/test/java/org/commcare/formplayer/junit/InitializeStaticsExtension.java
```diff
@@ -0,0 +1,29 @@
...
+import org.commcare.formplayer.util.PrototypeUtils;
+import org.javarosa.core.model.utils.DateUtils;
+import org.javarosa.core.reference.ReferenceManager;
+import org.javarosa.core.services.locale.Localization;
+import org.javarosa.core.services.locale.LocalizerManager;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class InitializeStaticsExtension implements BeforeEachCallback {
```
snopoke, line 14: taken from https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/BaseTestClass.java#L289-L295

src/test/java/org/commcare/formplayer/junit/MockTimezoneProvider.java
```diff
@@ -0,0 +1,17 @@
+package org.commcare.formplayer.junit;
+
+import org.javarosa.core.model.utils.TimezoneProvider;
+
+public class MockTimezoneProvider extends TimezoneProvider {
```
snopoke, line 5: moved from `BaseTestClass`

shubham1g5 (approved):

PR #1233, opened 2022-09-29
move plugin versions to gradle.properties

snopoke:
## Technical Summary

The main reason for this is to allow multiple plugin e.g. kotlin
to use the same version

https://github.com/dimagi/formplayer/pull/1231
https://github.com/dimagi/formplayer/pull/1232

### Safety story
build change only, no code change

### Rollback instructions

### Review

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,4 @@
+sprintBootPluginVersion=2.7.3
+gradleGitPropertiesPluginVersion=2.4.1
+lombokPluginVersion=6.5.1
+kotlinPluginVersion=1.7.10
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,7 @@ jobs:
           distribution: 'adopt'
       - name: Setup config
         run: |
-          cp .github/gradle.properties.ci gradle.properties
+          cp .github/gradle.properties.ci $HOME/gradle.properties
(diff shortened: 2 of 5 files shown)


gradle.properties
```diff
@@ -0,0 +1,4 @@
+sprintBootPluginVersion=2.7.3
```
shubham1g5, line 1: we can also store these versions inside the `build.gradle` itself like [here](https://github.com/dimagi/commcare-android/blob/master/build.gradle#L32). I like that style better since versions are maintained in the same file and also remove dependency on `gradle.propeties` which can be modified locally to add more attributes sometimes. Thoughts ?

snopoke: I'm open to changing this but as far as I know that doesn't work for plugin versions. From the docs: `«plugin version» must be constant, literal, strings`.

This is the recommended way to manage plugin versions: https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_version_management

Local gradle properties can still be provided via the `gradle.properties` file in the users home directory.

shubham1g5: Makes sense, happy to stick with current style in that case. Thanks!

shubham1g5 (approved):

snopoke:
Turns out this mechanism isn't supported by dependabot so I'm going to revert it.

PR #1235, opened 2022-09-30
github actions build to build staging artifacts

snopoke:
## Technical Summary
Build formplayer artifacts for staging and upload them to S3.

In order for this to work we need to set the following secrets:

* AWS_ACCESS_KEY_ID
* AWS_SECRET_ACCESS_KEY

We should be able to copy the ones from Jenkins.

This uses the [reggionick/s3-deploy](https://github.com/Reggionick/s3-deploy) action.

### Safety story
Build changes only.

### DIFF
--- a/.github/workflows/staging_build.yml
+++ b/.github/workflows/staging_build.yml
@@ -1,0 +1,42 @@
+name: Formplayer staging build
+
+on:
+  push:
+    branches: [autostaging]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


dannyroberts:
Nice. AWS seems to recommend/support a different method for authentication that allows the build to "assume a role" and use temporary credentials, which is usually preferred over hard coding AWS keys. I found this in https://github.com/aws-actions/configure-aws-credentials. I'll look a bit deeper about how to set it up on the other end, but just wanted to give you a heads up

PR #1245, opened 2022-10-11
disable request retries for requests made from formplayer

snopoke:
## Technical Summary
Currently if a request for Formplayer to HQ fails due to a 'connection error' e.g. timeout, the underlying HTTP library (OkHttp) will retry the request automatically (if the request meets certain criteria e.g. it's a GET request).

We have our timeouts set quite high and none of our requests can be retried exactly since they contain a unique 'Origin' header which is only valid for 60s.

### Safety story
I will put this on staging for basic testing.

### Automated test coverage
None

### QA Plan
Staging smoke testing

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/web/client/FormplayerOkHttp3ClientHttpRequestFactory.java
+++ b/src/main/java/org/commcare/formplayer/web/client/FormplayerOkHttp3ClientHttpRequestFactory.java
@@ -1,0 +1,15 @@
+package org.commcare.formplayer.web.client;
+
+import org.springframework.http.client.OkHttp3ClientHttpRequestFactory;
+
+import okhttp3.OkHttpClient;
+
+/**
+ * Custom OkHttp request factor that disables retries on connection failures.
+ */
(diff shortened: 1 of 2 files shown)


ctsims (approved):
Looks good to me based on what we discussed. We [a similar pattern in cc-core](https://github.com/dimagi/commcare-core/blob/d9066ea561f4ce85717b25263f9ddad695c98b29/src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java#L93), but we [differentiate a bit](https://github.com/dimagi/commcare-core/blob/d9066ea561f4ce85717b25263f9ddad695c98b29/src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java#L101) between the lib used for non-authed requests (primarily for resource fetches for app install) and the auth'd requests. I doubt that'll be an issue on formplayer, since requests are so much more likely to be successful, but it may be worth noting in the code or in a comment in case we find the need to revise in the future

snopoke:
I couldn't actually get this to work. It seems like OkHttp relies on retries to deal with closed connections in the connection pool: https://github.com/square/okhttp/issues/2738

I'm going to drop this for now.

PR #1256, opened 2022-10-20
More optimisations for Entity list performance

shubham1g5:
## Technical Summary

We were calculating the display properties for Entities on the current page twice. Once as part of [buildEntityList](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java#L225) and then again while [processEntity](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java#L287) to build a `EntityBean`. This PR changes the code to directly take use of `Entity` object evaluated fields to build the `EntityBean`

### Safety story

Mostly relying on unit test coverage

### Automated test coverage

A good number of navigation tests touches and tests over `EntityListResponse`

### QA Plan

None

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityDetailResponse.java
@@ -80,7 +80,7 @@ public class EntityDetailResponse {
             boolean isFuzzySearchEnabled) {
         List<Entity<TreeReference>> entityRefs = EntityListResponse.buildEntityList(detail, ec, references, null,
                 0, isFuzzySearchEnabled);
-        List<EntityBean> entityList = EntityListResponse.processEntitiesForCaseList(detail, entityRefs, ec, null);
+        List<EntityBean> entityList = EntityListResponse.processEntitiesForCaseList(entityRefs, ec, null);
         this.entities = new EntityBean[entityList.size()];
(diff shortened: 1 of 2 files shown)


snopoke (approved):
Looking at the `NoteEntityFactory` class it seems like there is quite a bit more cleanup that could be done in the future.

shubham1g5:
> Looking at the NoteEntityFactory class it seems like there is quite a bit more cleanup that could be done in the future.

curious what exactly are you thinking ?

snopoke:
> > Looking at the NoteEntityFactory class it seems like there is quite a bit more cleanup that could be done in the future.
> 
> curious what exactly are you thinking ?

I was mostly thinking that there is a fair amount of duplication between what's being done in EntityListResponse and in NoteEntityFactory.

It also seems like NoteEntityFactory was intended to be subclassed at some point but it isn't now and has some unused methods.

shubham1g5:
> It also seems like NoteEntityFactory was intended to be subclassed at some point but it isn't now and has some unused methods.

It actually gets subclassed on [mobile](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/AsyncNodeEntityFactory.java)

PR #1275, opened 2022-11-25
Minor Fix for return value of new install

shubham1g5:
## Technical Summary

When we fail to use the old ccz installation due to an [exception](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/InstallService.java#L70), we wipe out the old App DB and reinstall the ccz. Though we were not resetting `newInstall` to `true` and it was getting returned as `false` which doesn't seem right. 

The `newInstall` flag is used to do an [incremental sync](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/session/MenuSession.java#L122) later on in the function call chain. So this change will result into a  sync in cases when we we can't use the old App DB which seems like the right thing to happen.

### Safety story

Local testing and it's a code path that is already in use for new app installs. Also putting it on staging for passive testing. 

### Automated test coverage

None. 

### QA Plan

No QA

### Rollback instructions

### Review

### DIFF
--- a/src/test/resources/restores/caseclaim.xml
+++ b/src/test/resources/restores/caseclaim.xml
@@ -225,7 +225,7 @@
         <create>
             <case_type>child_case</case_type>
             <case_name>The Dark Knight</case_name>
-            <owner_id>4bc2b147d2e03808b69351c6373defa0</owner_id>
+            <owner_id>47ef5ea3cc348fb6538702449c855dcd</owner_id>
         </create>
         <update>
             <date_opened>2021-02-03</date_opened>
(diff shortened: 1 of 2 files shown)


shubham1g5:
Looking at test failures ..

src/test/resources/restores/caseclaim.xml
```diff
@@ -225,7 +225,7 @@
         <create>
             <case_type>child_case</case_type>
             <case_name>The Dark Knight</case_name>
-            <owner_id>4bc2b147d2e03808b69351c6373defa0</owner_id>
+            <owner_id>47ef5ea3cc348fb6538702449c855dcd</owner_id>
```
shubham1g5, line 228: This test was getiing purged earlier after restores causing a few tests to fail. I made the `owner_id` here same as the user id in this restore to avoid this getting purged after restores.

snopoke (approved):

shubham1g5:
@snopoke This can cause additional load on our servers when we do breaking App DB deploys as this change will result into a restore after FP refeteches the CC app. Therefore I am thinking of not merging this change even though this might be the right thing to do in long term.

PR #1278, opened 2022-11-28
Avoid clearing App DB when clearing user data

shubham1g5:
## Technical Summary

We were Calling `deleteDatabaseFolder` for clear user data requests which deletes the parent folder for the User DB file. This parent folder also contains the App DB for CC App. This doesn't seem like intentional behaviour to clear App DBs when requesting delete for user data, therefore I think it's more apt to just call `deleteDatabaseFile` instead of `deleteDatabaseFolder`. 

Making this change although results in having no direct way to make users clear App DB in case of any bugs.  Although projects should still be able to release new updates on their app for FP users to refresh their App DBs.

### Safety story

It should not result into any functional change for users except that FP will not request CC app again after clear user data. 

### Automated test coverage

None

### QA Plan

Not planning QA but would leave on staging for some time before merging.

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/UtilController.java
+++ b/src/main/java/org/commcare/formplayer/application/UtilController.java
@@ -98,7 +98,7 @@ public class UtilController {
                 requestBean.getDomain(),
                 requestBean.getUsername(),
                 requestBean.getRestoreAs()
-        ).deleteDatabaseFolder();
+        ).deleteDatabaseFile();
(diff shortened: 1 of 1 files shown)


snopoke (commented):
> This doesn't seem like intentional behaviour to clear App DBs when requesting delete for user data

I don't agree. This action is intended as a heavy hammer used to reset the user's state entirely which includes the app DB.

Since this is not something that should be necessary to use all the time I think it should stay as it is.

PR #1284, opened 2022-11-30
[WIP] Unchain creation and persistence of session objects from ordering

avazirna:
## Product Description
This PR aims to reduce the flakiness of the FormSessionRepoTest.testGetListView_Ordering test. This test involves sorting session objects by their creation date, and it fails when there are matching dates as the order becomes somewhat unpredictable. It is worth noting that the test can still fail but it shouldn't be frequent and if it becomes relevant, another way to solve this could be to add another field to the ordering criteria.

### Safety story
This change consisted in unchaining the creation, persistence and sorting of the objects, which was previously done with a single stream, using a List.

### Rollback instructions

### Review

### DIFF
--- a/src/test/java/org/commcare/formplayer/repo/FormSessionRepoTest.java
+++ b/src/test/java/org/commcare/formplayer/repo/FormSessionRepoTest.java
@@ -25,6 +25,7 @@ import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


snopoke (commented):
I don't see how this changes anything. From my read of the code we've just exchanged a `stream.map` with a for loop.

It's also surprising that 2 sessions could have the same date since the resolution of the date is nanoseconds. 

Where have you seen this test fail?

avazirna:
> 

It failed locally. This surprised me a bit too because it's a sequential stream, but it was still setting coincidental creation dates for objects persisted in sequence. However, it's important to note that this is more relevant to Java 8, I didn't have any problems on Java11 nor 17.

snopoke:
> > 
> 
> @snopoke It failed locally. This surprised me a bit too because it's a sequential stream, but it was still setting coincidental creation dates for objects persisted in sequence. However, it's important to note that this is more relevant to Java 8, I didn't have any problems on Java11 nor 17.

I also haven't seen any failures in the builds.

Re my other comment "From my read of the code we've just exchanged a stream.map with a for loop", did I miss something or is this still essentially the same thing just written in a different way?

avazirna:
> > > 
> > 
> > 
> > @snopoke It failed locally. This surprised me a bit too because it's a sequential stream, but it was still setting coincidental creation dates for objects persisted in sequence. However, it's important to note that this is more relevant to Java 8, I didn't have any problems on Java11 nor 17.
> 
> I also haven't seen any failures in the builds.
> 
> Re my other comment "From my read of the code we've just exchanged a stream.map with a for loop", did I miss something or is this still essentially the same thing just written in a different way?

@snopoke completely right. however, removing it from the stream made it more stable on my env and I was planning to also force it to sleep for 1 ms but I ended up not including that.

shubham1g5:
Ahh I thought `Streams` are `parallel` execution and suggested this change to Ahmad, but reading more it can be both `parallel` and `sequential` and the constructor we are using creates a `sequential`stream. So agree this change should not make any difference.

avazirna:
Holding this off for now

PR #1301, opened 2022-12-20
Rc/media meta data table

Robert-Costello:
## Technical Summary

https://dimagi-dev.atlassian.net/browse/USH-2190

relates to https://github.com/dimagi/formplayer/pull/1198

adds a db table to house metadata for media uploads

### Safety story

new feature not currently being used
tested locally
tests added in this PR

### Migrations

### Rollback instructions

### Review

[USH-2190]: https://dimagi-dev.atlassian.net/browse/USH-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/commcare/formplayer/db/migration/V26__init_media_meta_data.java
+++ b/src/main/java/org/commcare/formplayer/db/migration/V26__init_media_meta_data.java
@@ -4,6 +4,9 @@ import org.commcare.formplayer.util.Constants;
 
 import java.util.Arrays;
 
+/**
+ * Creates table for media metadata
+ */
 public class V26__init_media_meta_data extends BaseFormplayerMigration {
 
     @Override
--- a/src/main/java/org/commcare/formplayer/db/migration/V26__init_media_meta_data.java
+++ b/src/main/java/org/commcare/formplayer/db/migration/V26__init_media_meta_data.java
@@ -12,14 +12,15 @@ public class V26__init_media_meta_data extends BaseFormplayerMigration {
                 "CREATE TABLE " + Constants.POSTGRES_MEDIA_META_DATA_TABLE_NAME + " (\n" +
                         "    id varchar NOT NULL,\n" +
                         "    filePath varchar NOT NULL,\n" +
-                        "    FOREIGN KEY (formSessionId) REFERENCES formplayer_sessions(id) ON DELETE SET NULL (formSessionId),\n" +
+                        "    formSessionId varchar,\n" +
+                        "    CONSTRAINT formSessionId_constraint FOREIGN KEY (formSessionId) REFERENCES formplayer_sessions (id) ON DELETE SET NULL,\n" +
                         "    contentType varchar,\n" +
-                        "    contentLength integer,\n" +
                         "    username varchar NOT NULL,\n" +
                         "    asUser varchar,\n" +
                         "    domain varchar NOT NULL,\n" +
                         "    appid varchar NOT NULL,\n" +
                         "    datecreated timestamptz NOT NULL,\n" +
+                        "    contentLength integer\n" +
                         ")",
                 "CREATE INDEX ON " + Constants.POSTGRES_MEDIA_META_DATA_TABLE_NAME + " (datecreated)"
         );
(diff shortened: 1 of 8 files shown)


src/main/java/org/commcare/formplayer/db/migration/V26__init_media_meta_data.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Arrays;
+
+public class V26__init_media_meta_data extends BaseFormplayerMigration {
+
+    @Override
+    public Iterable<String> getSqlStatements() {
+        return Arrays.asList(
+                "CREATE TABLE " + Constants.POSTGRES_MEDIA_META_DATA_TABLE_NAME + " (\n" +
+                        "    id varchar NOT NULL,\n" +
+                        "    filePath varchar NOT NULL,\n" +
```
snopoke, line 14: These columns names must match the names provided by the `@Column` annotation in Java. So this should be `filepath`. Some of the others also don't match.

src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
```diff
@@ -0,0 +1,87 @@
...
+    private String username;
+
+    @Column(name = "asuser", updatable = false)
+    private String asUser;
+
+    @Column(updatable = false)
+    private String domain;
+
+    @Column(name = "appid", updatable = false)
+    private String appid;
```
snopoke, line 54: nit: consistent naming
```suggestion
    private String appId;
```

src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+@Entity
+@Table(name = Constants.POSTGRES_MEDIA_META_DATA_TABLE_NAME)
+@Getter
+public class MediaMetadataRecord {
+
+    @Id
+    @GeneratedValue(generator = "uuid")
+    @GenericGenerator(name = "uuid", strategy = "org.hibernate.id.UUIDGenerator")
+    private String id;
```
snopoke, line 29: @shubham1g5 is this ID going to be generated in the DB or manually?

src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Column(name = "filepath", updatable = false)
+    private String filePath;
+
+    @Setter
+    @Column(name = "formsessionid")
+    private String formSessionId;
```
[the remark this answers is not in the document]
Robert-Costello: I've been tinkering with this and need to work out how this type change will work in the constructor. I'm assuming the db would still be storing a varchar, is that correct?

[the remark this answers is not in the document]
snopoke: The DB type would be the same but you would pass in the object to the constructor or via a setter. I guess it depends a bit on the usage whether that makes sense of not.

src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
shubham1g5: This gets [generated manually](https://github.com/dimagi/formplayer/blob/saveMediaAttachment/src/main/java/org/commcare/formplayer/services/MediaHandler.kt#L24) , so we should not need to specify generator annotations here.

shubham1g5 (commented):
this looks in good shape to me except there is a test failure in the build for `HqUserDetailsServiceTests`. I think you will need to mock the `MediaMetaDataRepo` in that class for it to be successful

Robert-Costello:
> Also reg. Stickler errors you should be able to format the specific code by selecting the code or entire class you added in Android Studio/IntelliJ with a shortcut (cmd option L on mac).

That's fantastic, thanks! Looking forward to getting the style file/auto formatting incorporated.

Robert-Costello:
I'm assuming this regular expression isn't quite right since none of the other migration file names match it?
https://github.com/dimagi/formplayer/pull/1301/files#annotation_7671451641

shubham1g5 (approved):
The regular expression is correct though the migration classes doesn't follow the right naming convention for a Java class today which is ok to ignore for now.

PR #1305, opened 2023-01-09
Rc/save metadata in request

Robert-Costello:
## Technical Summary

Second part of https://dimagi-dev.atlassian.net/browse/USH-2190
Remove mediaMetaDataRecord if that record is being replaced in request to save a media file and creates new mediaMetaDataRecord for newly saved media files.

### Rollback instructions

### Review


src/main/java/org/commcare/formplayer/services/MediaHandler.kt
```diff
@@ -19,16 +21,37 @@ class MediaHandler(val file: MultipartFile) {
...
+                filePath.toString(),
+                session,
+                fileExtension,
+                file.size.toInt(),
+                username,
+                asUser,
+                domain,
+                appId
+            )
+            mediaMetaDataService.saveMediaMetaData(mediaMetaData)
```
shubham1g5, line 53: we should move out this block out of `try {} catch(){}` block since that's meant to catch the IOExceptions happening explicitly while saving file on filesystem.

src/main/java/org/commcare/formplayer/services/MediaHandler.kt
```diff
@@ -19,16 +21,37 @@ class MediaHandler(val file: MultipartFile) {
...
-        FileUtils.getExtension(file.originalFilename)?.let { fileIdWithExt = "$fileId.$it" }
-        val desintationFile = getMediaFilePath(parentDirPath, fileIdWithExt).toFile()
+        val fileExtension = FileUtils.getExtension(file.originalFilename)
+        fileExtension?.let { fileIdWithExt = "$fileId.$it" }
+        val filePath = getMediaFilePath(parentDirPath, fileIdWithExt)
+        val destinationFile = filePath.toFile()
+
+        val metaDataRecordToDelete = MediaMetaDataService().findByFilePath(filePath.toString())
+        val mediaMetaDataService = MediaMetaDataService()
+        metaDataRecordToDelete?.let{ mediaMetaDataService.deleteMetaDataById(it.id) }
```
shubham1g5, line 38: since this `filePath` is a new path generated above with a new fileId - `PropertyUtils.genUUID()` , we should not have any records at the moment for this in the meta data table making it pointless to perform a deletion here. Instead we wanna do the deletion when we delete a particular file from filesystem as a result of user deleting/replacing any file from a question in web apps. This deletion happens in `cleanMedia` method of `MediaHandler` which is where we should add this code to delete the file metadata record as well.

shubham1g5, line 38: Also curious if there is a reason behind doing a `exists` check before deleting the record here ?  I think it should be fine to call delete directly and probably put this in a try..catch block catching all exceptions as we don't care if delete record fails. 

````
try {
  mediaMetaDataService.deleteMetaDataById(fileId)
}catch(e: Exception){
 // ignore, we don't want to crash even if delete fails
}

**Review by shubham1g5 — CHANGES_REQUESTED** [review]
@Robert-Costello looks like a good start 👍 , I made some comments to address. Let me know if it makes sense. Thanks.

File: src/main/java/org/commcare/formplayer/services/MediaHandler.kt (hunk 1 above)
Comment (Robert-Costello) [reply src/main/java/org/commcare/formplayer/services/MediaHandler.kt:53]: I see, it makes sense to keep the `try {} catch(){}` focused on that specific exception. 
My thinking here was that we should save the metadata after the file is successfully saved, and this was the only place that would work. I can move the metadata save up above the `try`. Is there any concern about what happens with the metadata record if the file save fails, like should it be deleted in the `catch` block or something like that?

File: src/main/java/org/commcare/formplayer/services/MediaHandler.kt (hunk 2 above)
Comment (Robert-Costello) [reply src/main/java/org/commcare/formplayer/services/MediaHandler.kt:38]: Got it, I'll move this logic to `cleanMedia`. What I was thinking here was that if there was already a metadata record with the file path that is being constructed here, we'd find that record with that file path and remove it.

Comment (shubham1g5) [reply src/main/java/org/commcare/formplayer/services/MediaHandler.kt:38]: yeah that makes sense, though my sense is the chances of 2 files having the same `uuid` are practically close to near impossible i.e `PropertyUtils.genUUID()` should give us a near unique number.

File: src/main/java/org/commcare/formplayer/services/MediaHandler.kt (hunk 1 above)
Comment (Robert-Costello) [reply src/main/java/org/commcare/formplayer/services/MediaHandler.kt:53]: Code below won't ever be run, right? The `try-catch` will either return fileIdWithExt or throw an exception.

Comment (shubham1g5) [reply src/main/java/org/commcare/formplayer/services/MediaHandler.kt:53]: no `try` doesn't really `return` without us explicitly writing a `return` statement in the code. So in this case if there is an error the code will exit throwing an exception. Otherewise it should continue executing code written below line by line.

Comment (Robert-Costello) [reply src/main/java/org/commcare/formplayer/services/MediaHandler.kt:53]: There is currently a return statement in the `try` block `return fileIdWithExt` that was part of https://github.com/dimagi/formplayer/pull/1198. Should I remove that?

Comment (shubham1g5) [reply src/main/java/org/commcare/formplayer/services/MediaHandler.kt:53]: Ahh yes I missed it, we should move the return out of try-catch and put it in the end of the method.

File: src/main/java/org/commcare/formplayer/services/MediaHandler.kt [hunk 3]
```diff
@@ -54,6 +75,12 @@ class MediaHandler(val file: MultipartFile) {
     }
 
     fun cleanMedia(parentDirPath: Path, fileId: String): Boolean {
+        try {
```
shubham1g5, line 78: we should delete the meta data record after delete from filesystem is successful.

src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -66,29 +67,58 @@
...
     @Resource(name = "redisVolatilityDict")
     private ValueOperations<String, FormVolatilityRecord> volatilityCache;
-
     @Value("${commcarehq.host}")
     private String host;
 
-    private final Log log = LogFactory.getLog(FormController.class);
-    private final ObjectMapper mapper = new ObjectMapper();
+    // Iterate over all answers and attempt to save them to check for validity.
+    public static HashMap<String, ErrorBean> validateAnswers(FormEntryController formEntryController,
```
shubham1g5, line 90: it looks like the whole class got reformatted. It's fine for now but generally the best practice on formatting is to only re-format the code that you change in the PR to facilititate easier review.

src/main/java/org/commcare/formplayer/beans/AnswerQuestionRequestBean.java
```diff
@@ -23,6 +23,9 @@ public class AnswerQuestionRequestBean extends SessionRequestBean {
     @Nullable
     private Map<String, Object> answersToValidate;
 
+    @Nullable
+    public String metadataId;
```
shubham1g5, line 27: this should not be needed ?

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
```diff
@@ -20,6 +20,14 @@ public int purge() {
         return mediaMetaDataRepo.deleteMetaDataWithoutFormSessionId();
     }
 
+    public MediaMetadataRecord findByFormSessionId(String sessionId) {
+        return mediaMetaDataRepo.findByFormSessionId(sessionId).get();
+    }
+
+    public MediaMetadataRecord findByFilePath(String filePath) {
```
shubham1g5, line 27: this method looks unused at the moment.

src/main/java/org/commcare/formplayer/beans/AnswerQuestionRequestBean.java
Robert-Costello: no, it's not needed. thanks for pointing that out. I meant to remove that.

src/main/java/org/commcare/formplayer/application/FormController.java
Robert-Costello: That makes sense. I'll make sure to only reformat the changes I'm making going forward.

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
Robert-Costello: Using this now in testing that metadata is updated when file is reuploaded.

Robert-Costello:
I ran into an inconsistency where in `saveFile` MediaMetadataRecord was being saved with fileId(without the file extension) and `cleanMedia` was attempting to delete metadata with fileId(with the file extension added on). The path of least resistance was to have `saveFile` use `fileIdWithExt` to save the `metaDataRecord`. This works, but if it's preferable I could strip off the extension in `cleanMedia` and keep using fileId(without the file extension) in `saveFile`.

src/main/java/org/commcare/formplayer/exceptions/MediaMetaDataNotFoundException.java
```diff
@@ -0,0 +1,10 @@
+package org.commcare.formplayer.exceptions;
+
+/**
+ * Throw whenever a user tries to access a media metadata that does not exist.
+ */
+public class MediaMetaDataNotFoundException extends RuntimeException {
```
shubham1g5, line 6: I see this exception is only used in tests currently while mocking. We should have the mocking behaviour same as the main code. So we should also throw this exception from find methods in `MediaMetaDataService` or delete the exception altogether from tests as well.

shubham1g5:
> This works, but if it's preferable I could strip off the extension in cleanMedia and keep using fileId(without the file extension) in saveFile.

yeah I think that's preferrable approach as we already save extension in a different column in the table.

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
shubham1g5: As a best practice we should restrain from adding public methods in main code which are only used in tests. I think we can avoid both `findByFormSessionId` and  `findByFilePath` here by directly querying using `fileID`. You should be able to derive the `fileId` from the path here by getting the file name from the file path and then removing the extension.

shubham1g5 (changes requested):

src/main/java/org/commcare/formplayer/repo/MediaMetaDataRepo.java
```diff
@@ -15,6 +15,7 @@ public interface MediaMetaDataRepo extends JpaRepository<MediaMetadataRecord, St
 
     Optional<MediaMetadataRecord> findByFormSessionId(String formSessionId);
 
+    Optional<MediaMetadataRecord> findByFilePath(String filePath);
```
shubham1g5, line 18: can be removed now.

src/main/java/org/commcare/formplayer/services/MediaHandler.kt
```diff
@@ -55,6 +80,16 @@ class MediaHandler(val file: MultipartFile) {
 
     fun cleanMedia(parentDirPath: Path, fileId: String): Boolean {
```
shubham1g5, line 81: nit: rename `fileId` -> `fileIdWithExt`

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
```diff
@@ -20,6 +20,14 @@ public int purge() {
         return mediaMetaDataRepo.deleteMetaDataWithoutFormSessionId();
     }
 
+    public MediaMetadataRecord findById(String id) {
+        try {
+            return mediaMetaDataRepo.findById(id).get();
+        } catch (MediaMetaDataNotFoundException e) {
```
shubham1g5, line 26: this exception won't get thrown by itself from `mediaMetaDataRepo.findById(id)` but we will need to manually check if we got any result from `mediaMetaDataRepo.findById` and throw an exception. Look [here](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/FormSessionService.java#L58) for a similar example with form session repo.

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -236,7 +237,7 @@ private void setupFunctionContext() {
     private void loadInstanceXml(FormDef formDef, String instanceContent) throws IOException {
         StringReader stringReader = new StringReader(instanceContent);
         XFormParser xFormParser = new XFormParser(stringReader);
-        xFormParser.loadXmlInstance(formDef, stringReader);
+        XFormParser.loadXmlInstance(formDef, stringReader);
```
shubham1g5, line 240: I think it's best to revert this formatting change for the context of this PR. It's a good change but it also makes `xFormParser` in the line above unnecessary. And for this PR we will rather have a smaller change footprint even if it means we are ignoring some good code modifications in non relevant existing code.

src/test/java/org/commcare/formplayer/tests/MediaCaptureTest.kt
```diff
@@ -93,7 +102,12 @@ class MediaCaptureTest {
         var expectedFilePath = getExpectedMediaPath(formResponse.session_id, responseBean)
         val originalSavedFile = expectedFilePath.toFile()
         assertTrue("Could not find saved file on the filesystem", originalSavedFile.exists())
-
+        // check that metadata was created and values match expected
+        val fileName = expectedFilePath.fileName.toString()
+        val metadataId = fileName.substring(0, fileName.indexOf("."))
+        val originalMetadata = mediaMetaDataService.findById(metadataId)
```
shubham1g5, line 108: can we assert for all properties of `originalMetadata` atleast once during this test so that we know all properties are getting saved correctly. It would be ideal to create a new `private` function and do it there for the sake of readability.

shubham1g5 (commented):
Looks great fundamentally. Left some minor comments around styling.

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
```diff
@@ -21,11 +23,11 @@ public int purge() {
     }
 
     public MediaMetadataRecord findById(String id) {
-        try {
-            return mediaMetaDataRepo.findById(id).get();
-        } catch (MediaMetaDataNotFoundException e) {
-            throw e;
+        Optional<MediaMetadataRecord> record = mediaMetaDataRepo.findById(id);
+        if (record.isPresent()) {
```
shubham1g5, line 27: we should inverse the condition here ?

shubham1g5 (changes requested):

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
Robert-Costello: yep, don't know how I missed that. 🤦‍♂️

PR #1316, opened 2023-01-23
create media purge method and task

Robert-Costello:
## Technical Summary

Adds a periodic cleanup task to delete obsolete media and metadata.
https://dimagi-dev.atlassian.net/browse/USH-2194

### QA Plan

Will be QA'd along with larger media upload feature: https://dimagi-dev.atlassian.net/browse/QA-4739

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
+++ b/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
@@ -2,12 +2,14 @@ package org.commcare.formplayer.objects;
 
 import org.commcare.formplayer.util.Constants;
 import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.GenericGenerator;
 
 import java.time.Instant;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.FetchType;
+import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
@@ -25,7 +27,13 @@ import lombok.Setter;
 public class MediaMetadataRecord {
 
     @Id
+    @GeneratedValue(generator = "uuid")
+    @GenericGenerator(name = "uuid", strategy = "org.hibernate.id.UUIDGenerator")
     private String id;
+
+    @Setter
+    @Column(name = "fileid", updatable = false)
+    private String fileId;
 
     @Column(name = "filepath", updatable = false)
     private String filePath;
@@ -61,7 +69,7 @@ public class MediaMetadataRecord {
     }
 
     public MediaMetadataRecord(
-            String id,
+            String fileId,
             String filePath,
             SerializableFormSession formSession,
             String contentType,
@@ -70,7 +78,7 @@ public class MediaMetadataRecord {
             String asUser,
             String domain,
             String appId) {
-        this.id = id;
+        this.fileId = fileId;
         this.filePath = filePath;
         this.formSession = formSession;
         this.contentType = contentType;
@@ -83,24 +91,12 @@ public class MediaMetadataRecord {
 
     @Override
     public String toString() {
-        String formSessionId = "";
-        if (formSession != null) {
-            formSessionId = formSession.getId();
-        }
-        return "MediaMetaData [id=" + id + ", formSessionId=" + formSessionId + ", username=" + username
+        return "MediaMetaData [fileId=" + fileId + ", formSessionId=" + formSession.getId() + ", username=" + username
                 + ", asUser=" + asUser + " domain=" + domain + ", filePath=" + filePath
                 + ", contentType=" + contentType + "]";
     }
 
     public String getId() {
         return id;
-    }
-
-    public String getFilePath() {
-        return filePath;
-    }
-
-    public String getContentType() {
-        return contentType;
     }
 }
--- a/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
+++ b/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
@@ -31,7 +31,6 @@ public class MediaMetadataRecord {
     @GenericGenerator(name = "uuid", strategy = "org.hibernate.id.UUIDGenerator")
     private String id;
 
-    @Setter
     @Column(name = "fileid", updatable = false)
     private String fileId;
 
--- a/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
+++ b/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
@@ -83,7 +83,11 @@ public class MediaMetadataRecord {
 
     @Override
     public String toString() {
-        return "MediaMetaData [id=" + id + ", formSessionId=" + formSession.getId() + ", username=" + username
+        String formSessionId = "";
+        if (formSession != null) {
+            formSessionId = formSession.getId();
+        }
+        return "MediaMetaData [id=" + id + ", formSessionId=" + formSessionId + ", username=" + username
                 + ", asUser=" + asUser + " domain=" + domain + ", filePath=" + filePath
                 + ", contentType=" + contentType + "]";
     }
--- a/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
+++ b/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
@@ -87,4 +87,16 @@ public class MediaMetadataRecord {
                 + ", asUser=" + asUser + " domain=" + domain + ", filePath=" + filePath
                 + ", contentType=" + contentType + "]";
     }
+
+    public String getId() {
+        return id;
+    }
+
+    public String getFilePath() {
+        return filePath;
+    }
+
+    public String getContentType() {
+        return contentType;
+    }
 }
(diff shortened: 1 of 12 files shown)


src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
```diff
@@ -41,6 +45,7 @@ public void purge() {
         log.info("Starting purge scheduled task.");
         doTimedPurge("formSession", formSessionService::purge);
         doTimedPurge("virtualDataInstance", virtualDataInstanceService::purge);
+        doMediaPurge("media", mediaHandler::purge);
```
Robert-Costello, line 48: @shubham1g5   Originally I was using `doTimedPurge` to call `mediaHandler::purge` which required an `Instant` arg to be passed in. This gets rid of that required and unused arg. I'm curious if there are any different approaches that are preferable.

shubham1g5: Ahh I see what you meant earlier. I think in this case for the sake of consistency and avoiding code duplication, it's alright if we have unused `Instant` as an arg for `mediaHandler::purge`

Robert-Costello:
@shubham1g5 It looks like the mock formSession is deleted after this point.
https://github.com/dimagi/formplayer/blob/bca7e2d238f7ee20cba6a5d57046474e1309a54d/src/test/java/org/commcare/formplayer/tests/MediaCaptureTest.kt#L216

To test the media purge would it make sense to have this method (deleteSessionById) also set the `formSession` of the corresponding metadata record to null? Then later when we call mediaHandler.purge() check that the record has been deleted?https://github.com/dimagi/formplayer/blob/bca7e2d238f7ee20cba6a5d57046474e1309a54d/src/test/java/org/commcare/formplayer/junit/FormSessionServiceExtension.kt#L59-L63

Open to any other ideas as well.

shubham1g5:
@Robert-Costello I see, In that case I think we can just have another test to test purge something like - 

````
void testPurge{
 saveImage
 setFormSessionIDAsNull
 purge
 verify meta data record and file is delete
}
````

File: src/test/java/org/commcare/formplayer/tests/MediaCaptureTest.kt [hunk 2]
```diff
@@ -235,6 +237,43 @@ class MediaCaptureTest {
...
+        val fileName = expectedFilePath.fileName.toString()
+        val metadataId = fileName.substring(0, fileName.indexOf("."))
+        val metadata = mediaMetaDataService.findById(metadataId)
+        assertEquals(metadata.formSession.id, formResponse.session_id)
+
+        val fis = FileUtils.getFileStream(this.javaClass, expectedFilePath.toString())
+        val file = MockMultipartFile(PART_FILE, fileName, MediaType.IMAGE_JPEG_VALUE, fis)
+
+        metadata.formSession = null
+        mediaMetaDataService.saveMediaMetaData(metadata)
```
shubham1g5, line 265: we should simplify this by omitting the checks that we are already doing as part of `testImageCapture_fileSaveAndReplace`. We onlly want to test purge as part of this test, so we should only retain the code that's necessary for `purge` testing below and remove the additional checks.

shubham1g5 (changes requested):
Looks great! Left couple minor things to address

src/main/java/org/commcare/formplayer/services/ScheduledTasks.java
```diff
@@ -32,6 +32,9 @@ public class ScheduledTasks {
     @Autowired
     private VirtualDataInstanceService virtualDataInstanceService;
 
+    @Autowired
+    private MediaHandler mediaHandler;
```
shubham1g5, line 36: I don't think this will work as MediaHandler requires a `MultipartFile` attribute to be initialised which we are not supplying anywhere. I think the best approach here is to move `purge` method in `MediaMetaDataService` instead and then call `cleanMedia` from there. 

````
val deletedSuccessfully = MediaHandler.cleanMedia(parentPath, fileIdWithExt)
````

To do this we will also need to make `cleanMedia` static so that we don't need to initialise `MediaHandler` separaterly for this call. I am going to make some small refactors on this branch to be able to do that.

Comment (Robert-Costello) [reply src/main/java/org/commcare/formplayer/services/ScheduledTasks.java:36]: >MediaHandler requires a MultipartFile attribute to be initialised which we are not supplying anywhere

ah, I missed that outside of the test.

Comment (shubham1g5) [reply src/main/java/org/commcare/formplayer/services/ScheduledTasks.java:36]: Ok I made a change  [here](https://github.com/dimagi/formplayer/pull/1316/commits/03930c4f38485307e2598c4e1fdba1eb2674290f). Transitioning to you to make the change to move `purge` to `MediaMetaDataService` instead.

Comment (Robert-Costello) [reply src/main/java/org/commcare/formplayer/services/ScheduledTasks.java]: I'm running into a weird blocker where "`this` is not available" in the debugger. That is making the test fail and making it hard to reason about what is and isn't working in moving the `purge` method to the service. Have you run into this before?

File: src/main/java/org/commcare/formplayer/repo/MediaMetaDataRepo.java [hunk 4]
```diff
@@ -2,22 +2,17 @@
...
 public interface MediaMetaDataRepo extends JpaRepository<MediaMetadataRecord, String> {
 
-    Optional<MediaMetadataRecord> findByFormSessionId(String formSessionId);
+    List<MediaMetadataRecord> findByFormSession(String formSessionId);
 
-    @Modifying
-    @Transactional
-    @Query("DELETE FROM MediaMetadataRecord WHERE formSession IS NULL")
-    int deleteMetaDataWithoutFormSessionId();
+    @Query(value = "SELECT * FROM media_meta_data WHERE formsessionid IS NULL", nativeQuery = true)
```
shubham1g5, line 16: we probabaly don't need to specify manual query here ? (Since the source of our errors was something else)

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
```diff
@@ -30,11 +32,37 @@ public MediaMetadataRecord findById(String id) {
...
 
     public void deleteMetaDataById(String id) {
         mediaMetaDataRepo.deleteById(id);
     }
+
+    public List<MediaMetadataRecord> findAllWithNullFormSession() {
+        return mediaMetaDataRepo.getFormSessionIsNull();
+    }
+
+    public List<MediaMetadataRecord> findAll() {
```
shubham1g5, line 47: looks unused.

src/test/java/org/commcare/formplayer/services/MediaMetaDataServiceTest.java
```diff
@@ -1,83 +1,145 @@
...
+        Optional<MediaMetadataRecord> newlyFetchedMediaMetaData = mediaMetaDataRepo.findById(metadataId);
+        Assertions.assertFalse(newlyFetchedMediaMetaData.isPresent());
+    }
 
-        mediaMetaDataService.deleteMetaDataById(mediaMetaData.getId());
+    @Test
+    public void testPurge() {
+        Assertions.assertTrue(testFile.exists());
+        Integer purgeCount = mediaMetaDataService.purge(Instant.now());
+        Assertions.assertEquals(1, purgeCount);
```
shubham1g5, line 129: also verify `testFile.exisits()` is false now

Robert-Costello: I'll go ahead and add that assert. purgeCount is not passing for some reason. still working on why.

src/main/java/org/commcare/formplayer/repo/MediaMetaDataRepo.java
Robert-Costello: yeah, that makes sense to me.

src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
```diff
@@ -25,8 +27,14 @@
 public class MediaMetadataRecord {
 
     @Id
+    @GeneratedValue(generator = "uuid")
+    @GenericGenerator(name = "uuid", strategy = "org.hibernate.id.UUIDGenerator")
     private String id;
 
+    @Setter
```
shubham1g5, line 34: we should remove `Setter` from here to keep `fileId` unmutable

src/test/java/org/commcare/formplayer/services/MediaMetaDataServiceTest.java
```diff
@@ -1,83 +1,146 @@
...
-        mediaId = UUID.randomUUID().toString();
+    public void setUp() throws IOException {
+        fileId = UUID.randomUUID().toString();
+        testFile = new File("media/" + fileId + ".jpg");
+        testFile.createNewFile();
         mediaMetaData = new MediaMetadataRecord(
-                mediaId,
-                "filePath",
+                fileId,
+                "media/" + fileId + ".jpg",
```
shubham1g5, line 58: nit: use a variable to store filePath in and use it here and above while creating file.

Robert-Costello:
So removing that custom query causes `MediaMetaDataRepoTest` to fail. Digging into why that is.

shubham1g5 (approved):
This looks good and we are good to merge in parent branch to start QA. Although one thing we are missing right now which is probably worth a separate PR is `cache` implementation above the DB layer for the meta data record table. It's quite easy to implement it (look [here](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/MenuSessionService.java#L26) for an example). We also then need to modify the service test accordingly to test for correct caching. In addition we need to add expiration properties for the [cache](https://github.com/dimagi/formplayer/blob/master/src/main/resources/application.properties#L35) in application.properties.

Robert-Costello:
Ok, I'll merge this to let QA get started and work on the caching next.

PR #1320, opened 2023-01-31
Handles noItemsText in HQ if null passed to commcare-core

Jtang-1:
## Product Description
Related to HQ PR: https://github.com/dimagi/commcare-hq/pull/32552
and previous Formplayer PR: https://github.com/dimagi/formplayer/pull/1309

## Technical Summary

Same as described in https://github.com/dimagi/formplayer/pull/1309
New to this PR: [USH-2869](https://dimagi-dev.atlassian.net/browse/USH-2869)
### Safety story

Manual testing on local env and staging. Existing apps and those with Feature Flag off or Commcare version < 2.53 will return null to HQ. HQ then handles null accordingly.

### Automated test coverage

No new automated tests.
### QA Plan

Will go through QA.

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
@@ -526,8 +526,7 @@ public class EntityListResponse extends MenuBean {
         try {
             noItemsTextString = detail.getNoItemsText().evaluate();
         } catch (NoLocalizedTextException | NullPointerException e) {
-            String noItemsTextStringLegacy = "List is empty.";
-            return noItemsTextStringLegacy;
+            return null;
         }
         return noItemsTextString;
     }


src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
```diff
@@ -526,8 +526,7 @@ private String getNoItemsTextLocaleString(Detail detail) {
         try {
             noItemsTextString = detail.getNoItemsText().evaluate();
         } catch (NoLocalizedTextException | NullPointerException e) {
-            String noItemsTextStringLegacy = "List is empty.";
-            return noItemsTextStringLegacy;
+            return null;
```
shubham1g5, line 529: Since the default text right now is `List is empty`, should not we stick to that instead of `null` ?

Jtang-1: Just double checking - changes to this repo (formplayer) will only affect web apps and not impact mobile correct?

If so, this change is connected this HQ [PR](https://github.com/dimagi/commcare-hq/pull/32552). I decided to, on the HQ [template](https://github.com/dimagi/commcare-hq/blob/eb3bff3ba0918163091a7f5528f87e66ebddd7d1/corehq/apps/cloudcare/templates/formplayer/case_list.html#L47-L50) side, handle the null passed back from formplayer and display `{% trans "List is empty." %}`. I don't think how it was done before  (where `List is empty.` is passed from formplayer and displayed through `<%- noItemsText %>`) would properly translate the default text for builds that did not contain the `<no_items_text> node` (such as existing apps, those with the FF off, or commcare version <  2.54)

shubham1g5: >  changes to this repo (formplayer) will only affect web apps and not impact mobile correct?

Right

I don't think this change matters at all. Without this change, even if the apps that don't have `no_items_text`, FP would convey default text `List is empty` to web apps. HQ should not be required to do any `null` handling that way either. So current version where FP is passing the final text to be displayed to HQ in all cases seems more consistent to me.

Jtang-1: Would the default text get translated correctly in the current version? (I made this change because I thought it wouldn't)

shubham1g5: Ahh is HQ doing language tranlation on default text ?

Jtang-1: Yes, with this PR and HQ [PR](https://github.com/dimagi/commcare-hq/pull/32552), when HQ receives null from formplayer, it displays {% trans "List is empty." %}. The `trans` tags this string to be translated.

Previously, when `List is empty` is passed directly to web apps and displayed, there wasn't translation being done.

shubham1g5: got it. though if there was no translation support for default text before, I think it's fine to add it through `no_items_text` instead and make it the case that Web Apps that want to translate this should target CC 2.54 in app settings. This is just so that we keep the code consistent in FP being the primary component responsible for translation.

Jtang-1: I think I miscommunicated. Currently in production, the default text has translation support (through the `trans` tag). 

In the previous PR that was reverted, it did not have translation support for default text on web apps (when FF is off). It directly displays `List is Empty.` from FP.

In this PR, the default text keeps translation support (through the `trans` tag) (when FF is off or CC version < 2.54)

I'm not following this part - does FP handle translation?

> This is just so that we keep the code consistent in FP being the primary component responsible for translation.

shubham1g5: > I think I miscommunicated. Currently in production, the default text has translation support (through the trans tag).

Ahh it makes sense then to keep the original behaviour. 

> does FP handle translation?

yeah for most UI strings translation come from FP/App config files which is what the change for `no_items_text` is now correcting.

PR #1321, opened 2023-02-01
Fix save for media meta data table

shubham1g5:
Since we are pre-setting an id to the media meta data record instead of auto-generating we need to do some custom handling as [specified here](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.entity-persistence.saving-entites.strategies) to persist entities. Without this JPA treats the entities with a pre-set ID as an old one and tries to do an update statement on save instead of insert resulting into a failure

### DIFF
--- a/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
+++ b/src/main/java/org/commcare/formplayer/objects/MediaMetadataRecord.java
@@ -2,6 +2,7 @@ package org.commcare.formplayer.objects;
 
 import org.commcare.formplayer.util.Constants;
 import org.hibernate.annotations.CreationTimestamp;
+import org.springframework.data.domain.Persistable;
 
 import java.time.Instant;
 
@@ -11,7 +12,10 @@ import javax.persistence.FetchType;
 import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
+import javax.persistence.PostLoad;
+import javax.persistence.PrePersist;
 import javax.persistence.Table;
+import javax.persistence.Transient;
 
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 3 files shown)


Robert-Costello (approved):
Nice work. Good to understand why it was treating new uploads as preexisting.

snopoke (commented):
The other alternative is to have separate DB ID from application ID, this is what we did with the `SerializableDataInstance` objects.

shubham1g5:
@snopoke yes that makes sense. I don't have a strong preference between the two approach here. The solution in this PR adds some extra handling in Entity class while the other creates an additional column in the table. Do you think we should go for having a separate DB id  for the sake of consistency ?

snopoke:
> @snopoke yes that makes sense. I don't have a strong preference between the two approach here. The solution in this PR adds some extra handling in Entity class while the other creates an additional column in the table. Do you think we should go for having a separate DB id for the sake of consistency ?

I have a slight preference to the separate DB column for the sake of consistency and I also think it's a bit simpler in code and to reason about. I've found the other method can be tricky.

shubham1g5:
Makes sense. Thanks for the input. Going to close this in that case. 

@Robert-Costello will you be able to take next steps here. We should separate out the DB id `id` with the fileID here. We can call the new column `key` in consistency with how we have done it with `SerializableDataInstance`.  We will also need to do a DB migraton similar to [this one](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/db/migration/V24__virtual_data_instance_key.java)`.  Happy to offline in case it's not clear what is needed here. Also the changes in test classes are still worth here so would appreciate if you can copy/cherry-pick them over in your PR. Thanks!

Robert-Costello:
> @Robert-Costello will you be able to take next steps here.

Sure, I will work on this. (already pulled the tests out)

PR #1327, opened 2023-02-08
[USH-2620] improved 'back' navigation after form submissions

snopoke:
## Product Description
Jira: https://dimagi-dev.atlassian.net/browse/USH-2620

See HQ PR for full description: https://github.com/dimagi/commcare-hq/pull/32575

## Technical Summary
HQ cross request: https://github.com/dimagi/commcare-hq/pull/32575

Detect the presence of a 'form_session_id' during a navigation request and act appropriately. If the session ID is present it implies that the user has performed some non-app based navigation such as going back in the browser history.

When this is detected, attempt to navigate to the previous screen prior to form submission.

### Safety story
The first few commits are cleanups, only the last commit is a functional change.

Assuming the HQ portion of this change is working correct and the form_session_id param is properly cleared after form submission, this should only impact manual navigation.

### Automated test coverage
TODO

### QA Plan
QA on staging

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -124,7 +124,7 @@ public class MenuSessionRunnerService {
     private static final Log log = LogFactory.getLog(MenuSessionRunnerService.class);
 
     public BaseResponseBean getNextMenu(MenuSession menuSession) throws Exception {
-        return getNextMenu(menuSession, null, 0, "", 0, null, 0, null);
+        return getNextMenu(menuSession, null, 0, "", 0, null, 0);
     }
 
     @Trace
@@ -134,8 +134,7 @@ public class MenuSessionRunnerService {
             String searchText,
             int sortIndex,
             QueryData queryData,
-            int casesPerPage,
-            String smartLinkTemplate) throws Exception {
+            int casesPerPage) throws Exception {
         Screen nextScreen = menuSession.getNextScreen(detailSelection != null);
 
         // No next menu screen? Start form entry!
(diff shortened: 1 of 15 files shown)


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -307,8 +316,11 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
                 }
             }
         } catch (CommCareSessionException ccse) {
-            notificationMessage = new NotificationMessage(ccse.getMessage(), true,
-                    NotificationMessage.Tag.menu);
+            // don't show a message in cases where the user is doing 'non-app' navigation
+            if (!nonAppNav) {
```
shubham1g5, line 320: Since we are now showing the screen prior to form entry we don't really expect an error to happen here. So curious why do we want to hide the errors for `nonAppNav` ?

shubham1g5 (commented):
Would be nice to add a navigation test to check for prior to form entry navigation.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
snopoke: There can still be errors if the selected case was closed by the form or if the menu isn't visible anymore.

snopoke:
> Would be nice to add a navigation test to check for prior to form entry navigation.

I am working on some tests but can you elaborate on what you mean.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5: makes sense 👍

shubham1g5 (approved):
@snopoke you have covered what I was thinking in terms of tests already 👍 . This looks great  though a stickler pass would be nice here.

PR #1339, opened 2023-02-28
Update Dockerfile to use JRE 17

avazirna:
## Product Description
This updates the image's Dockerfile to use Java 17 instead of 8.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM adoptopenjdk:8-jre-hotspot as builder
+FROM eclipse-temurin:17-jre as builder
 WORKDIR application
 ARG JAR_FILE=build/libs/*.jar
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM adoptopenjdk:8-jre-hotspot as builder
+FROM eclipse-temurin:17-jre as builder
```
shubham1g5, line 1: can we not use open jdk as well for Java 17 ?

avazirna: I tried to look for an openjdk image but [they are deprecated now](https://github.com/docker-library/openjdk/issues/505), and one of their suggestions is [Eclipse Temurin](https://hub.docker.com/_/eclipse-temurin/).

shubham1g5: makes sense. thanks!

PR #1353, opened 2023-03-28
Rc/cache media metadata

Robert-Costello:
## Technical Summary

this adds caching for media metadata as well as additional tests 

### Automated test coverage

tests added here

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
+++ b/src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
@@ -44,8 +44,8 @@ public class MediaMetaDataService {
         return mediaMetaDataRepo.save(mediaMetadataRecord);
     }
 
-    public void deleteMetaDataById(String id) {
-        mediaMetaDataRepo.deleteById(id);
+    public void deleteByFileId(String fileId) {
+        mediaMetaDataRepo.deleteByFileId(fileId);
     }
 
     public List<MediaMetadataRecord> findAllWithNullFormSession() {
--- a/src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
+++ b/src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
@@ -39,11 +39,21 @@ public class MediaMetaDataService {
         return record.get();
     }
 
+    @Cacheable
+    public MediaMetadataRecord getMetadataById(String id) {
+        Optional<MediaMetadataRecord> record = mediaMetaDataRepo.findById(id);
+        if (!record.isPresent()) {
+            throw new MediaMetaDataNotFoundException(id);
+        }
+        return record.get();
+    }
+
     @CachePut(key = "#mediaMetadataRecord.id")
     public MediaMetadataRecord saveMediaMetaData(MediaMetadataRecord mediaMetadataRecord) {
         return mediaMetaDataRepo.save(mediaMetadataRecord);
     }
 
+    @CacheEvict(key = "#id")
     public void deleteMetaDataById(String id) {
         mediaMetaDataRepo.deleteById(id);
     }
--- a/src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
+++ b/src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
@@ -1,11 +1,16 @@
 package org.commcare.formplayer.services;
 
 import static org.commcare.formplayer.services.MediaHandler.cleanMedia;
+import static org.commcare.formplayer.util.Constants.MEDIA_METADATA_CACHE;
 
 import org.commcare.formplayer.exceptions.MediaMetaDataNotFoundException;
 import org.commcare.formplayer.objects.MediaMetadataRecord;
 import org.commcare.formplayer.repo.MediaMetaDataRepo;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cache.annotation.CacheConfig;
+import org.springframework.cache.annotation.CacheEvict;
+import org.springframework.cache.annotation.CachePut;
+import org.springframework.cache.annotation.Cacheable;
 import org.springframework.stereotype.Service;
 
 import java.nio.file.Path;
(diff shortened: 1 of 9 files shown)


shubham1g5 (changes requested):
One minor nice to have for tests, looks gtm otherwise.

src/test/java/org/commcare/formplayer/services/MediaMetaDataServiceTest.java
```diff
@@ -102,39 +113,68 @@ public List<MediaMetadataRecord> answer(InvocationOnMock invocation) throws Thro
...
+    public void testSaveMetaData_cached() {
+        cacheManager.getCache(MEDIA_METADATA_CACHE).clear();
+
         mediaMetaDataService.saveMediaMetaData(mediaMetaData);
-        Optional<MediaMetadataRecord> fetchedMediaMetaData = mediaMetaDataRepo.findById(mediaMetaData.getId());
-        Assertions.assertTrue(fetchedMediaMetaData.isPresent());
+
+        assertEquals(mediaMetaData, getCachedMetadata(mediaMetaData.getId()).get());
+
+        MediaMetadataRecord fetchedMediaMetaData = mediaMetaDataService.getMetadataById(mediaMetaData.getId());
```
shubham1g5, line 135: Ahh sorry I didn't mean for us to introduce a new function in service layer that we are only using in test. Since we are querying the service in live code by using `findByFileId` we should be using that method to get the record here. And also while writing the record, instead of caching by `#mediaMetadataRecord.id` we should cache by `fileId` since that's what we are using to query the DB later on.

Robert-Costello: That makes total sense. I wasn't thinking about the broader implications. I switched to using `findByFileId` and caching with `fileId`.

snopoke (approved):

src/test/java/org/commcare/formplayer/services/MediaMetaDataServiceTest.java
```diff
@@ -1,9 +1,16 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.when;
 
+import static java.util.Optional.ofNullable;
+
+import org.commcare.formplayer.configuration.CacheConfiguration;
```
Robert-Costello, line 13: Am I missing something here? It looks like the imports are alphabetical and this `CacheConfiguration` import is in the correct position. 
@shubham1g5 @snopoke

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
```diff
@@ -39,21 +39,11 @@ public MediaMetadataRecord findByFileId(String id) {
...
-    }
-
-    @CachePut(key = "#mediaMetadataRecord.id")
+    @CachePut(key = "#mediaMetadataRecord.fileId")
     public MediaMetadataRecord saveMediaMetaData(MediaMetadataRecord mediaMetadataRecord) {
         return mediaMetaDataRepo.save(mediaMetadataRecord);
     }
 
-    @CacheEvict(key = "#id")
     public void deleteMetaDataById(String id) {
```
shubham1g5: Calling any delete method sshould also evict the corresponding record from cache as otherwise read will return the cached record even when a record has been deleted from DB.

Also do we really need to implement delete by id ? Instead can we implement and use `deleteByFileId` since we are doing other if it's simpler to do ?

src/test/java/org/commcare/formplayer/services/MediaMetaDataServiceTest.java
shubham1g5: Hmm yeah stickler sometimes give this issue even when it's in the right order. It's fine to ignore if you had formatted the imports in accordance with IntelliJ code style file.

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
Robert-Costello: Thanks. I should have removed that when I switched over to using `fileId`.

Robert-Costello: > Calling any delete method should also evict the corresponding record from cache

Thanks, that makes sense.

src/main/java/org/commcare/formplayer/services/MediaMetaDataService.java
```diff
@@ -32,12 +39,13 @@ public MediaMetadataRecord findByFileId(String id) {
...
     }
 
+    @CachePut(key = "#mediaMetadataRecord.fileId")
     public MediaMetadataRecord saveMediaMetaData(MediaMetadataRecord mediaMetadataRecord) {
         return mediaMetaDataRepo.save(mediaMetadataRecord);
     }
 
-    public void deleteMetaDataById(String id) {
-        mediaMetaDataRepo.deleteById(id);
+    public void deleteByFileId(String fileId) {
```
shubham1g5, line 47: this should evict the particular record from the cache.

src/test/java/org/commcare/formplayer/services/MediaMetaDataServiceTest.java
```diff
@@ -102,39 +112,66 @@ public List<MediaMetadataRecord> answer(InvocationOnMock invocation) throws Thro
...
         mediaMetaDataService.saveMediaMetaData(mediaMetaData);
-        Optional<MediaMetadataRecord> fetchedMediaMetaData = mediaMetaDataRepo.findById(mediaMetaData.getId());
-        Assertions.assertTrue(fetchedMediaMetaData.isPresent());
         String metadataId = mediaMetaData.getId();
-        mediaMetaDataService.deleteMetaDataById(metadataId);
+        String fileId = mediaMetaData.getFileId();
+        MediaMetadataRecord fetchedMediaMetaData = mediaMetaDataService.findByFileId(fileId);
+        Assertions.assertNotNull(fetchedMediaMetaData);
+        mediaMetaDataService.deleteByFileId(fileId);
         Optional<MediaMetadataRecord> newlyFetchedMediaMetaData = mediaMetaDataRepo.findById(metadataId);
```
shubham1g5, line 146: we should still use service layer here to find the record. We can just do another `findByFileId` call on the service to check for the existing of the record. This will verify that service cache also invalidates when we delete a record.

shubham1g5 (changes requested):

src/test/java/org/commcare/formplayer/services/MediaMetaDataServiceTest.java
Robert-Costello: This was a good change to make. I tested it and the cache was not being updated on delete until `@CacheEvict(key  = "#fileId")` annotation was added to `deleteByFileId`.

shubham1g5 (approved):

PR #1374, opened 2023-05-12
add 'hasDetails' to EntityListResponse

snopoke:
## Product Description
Add `hasDetails` attribute to entity list response to allow Web Apps to skip the details request when there is no 'long detail' configured.

Cross: https://github.com/dimagi/commcare-core/pull/1280

## Technical Summary
Adds `hasDetails` to the entity list response. 

### Safety story
Simple addition of one field.

### QA Plan
Tested locally

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
@@ -61,6 +61,8 @@ public class EntityListResponse extends MenuBean {
     private boolean isMultiSelect = false;
     private int maxSelectValue = -1;
 
+    private boolean hasDetails = true;
+
     public EntityListResponse() {
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
```diff
@@ -267,6 +270,10 @@ public void setSortIndices(int[] sortIndices) {
         this.sortIndices = sortIndices;
     }
 
+    public boolean isHasDetails() {
```
stephherbers, line 273: super nit: awkward name, `doesHaveDetails`?

stephherbers (approved):

src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java
shubham1g5: It does need to be `is$booleanVar` for the Java model to Json conversion as a default conversion mechanism. We can still change it but think it's quite normal in Java to have booleans named like this.

PR #1391, opened 2023-06-02
Add IP address to post requests to HQ

shubham1g5:
## Product Description

https://dimagi-dev.atlassian.net/browse/USH-3093

## Technical Summary

Probably all the complexity here is in to get the IP address from the request. I am using the header "X-FORWARDED-FOR" to get the IP with a fall back to the `request.getRemoteAddr()` . I am assuming that `"X-FORWARDED-FOR"  will always be there when requests are forwarded from Nginx but we probably need to cross-check that assumption on HQ. 

### Safety story

Testing form submission locally and on staging should be a good enough safety check. 

### QA Plan

No QA
### Migrations

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/util/RequestUtils.java
+++ b/src/main/java/org/commcare/formplayer/util/RequestUtils.java
@@ -121,6 +121,22 @@ public class RequestUtils {
         return attribute != null && (Boolean)attribute;
     }
 
+    /**
+     * Gets the IP address by using "X-FORWARDED-FOR" request header falling back to request.getRemoteAddr()
+     * @return IP address from the request
+     */
+    public static String getIpAddress() {
+        HttpServletRequest request = getCurrentRequest();
+        if (request == null) {
+            return null;
+        }
+        String ipAddress = request.getHeader("X-FORWARDED-FOR");
+        if (ipAddress == null) {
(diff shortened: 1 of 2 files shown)


snopoke (approved):
We should be able to test this on staging.

shubham1g5:
> We should be able to test this on staging.

@snopoke I have put this on staging but not really sure how to test it. I don't really see `ip` get added to [form metadata](https://staging.commcarehq.org/a/shubhamgoyaltest/reports/form_data/5fb6bb3a-f94d-4aae-b016-5f827c957405/#form-metadata) on HQ. Is that an incorrect expectation with this work ? Or we only expect to see the `ip` through the form export ?

snopoke:
> > We should be able to test this on staging.
> 
> @snopoke I have put this on staging but not really sure how to test it. I don't really see `ip` get added to [form metadata](https://staging.commcarehq.org/a/shubhamgoyaltest/reports/form_data/5fb6bb3a-f94d-4aae-b016-5f827c957405/#form-metadata) on HQ. Is that an incorrect expectation with this work ? Or we only expect to see the `ip` through the form export ?

It's not part of the displayed metadata. You'll need to use the raw docs view to see it.

snopoke (approved):

PR #1396, opened 2023-06-07
Fix Incorrect Entity Context Attached to Entity Screen

shubham1g5:
## Technical Summary

https://dimagi-dev.atlassian.net/browse/SUPPORT-16768

There were some places in code that were passing the incorrect EntityContext to entity screen. This code pathway was triggered when there in auto computed datum present first before the nodeset datum. I have made the changes to pass in right entityContext to the session navigation code and also removed some methods which were re-initialising Entity context in order to lessen probability of us making a similar mistake again. 

### Safety story

Mostly ensuring safety through tests and local testing

### Automated test coverage

PR adds a regression test to verify the behaviour

### QA Plan

Not planned

### Migrations

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -3,6 +3,8 @@ package org.commcare.formplayer.services;
 import static org.commcare.formplayer.util.Constants.TOGGLE_SESSION_ENDPOINTS;
 import static org.commcare.formplayer.util.Constants.TOGGLE_SPLIT_SCREEN_CASE_SEARCH;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.commcare.core.interfaces.RemoteInstanceFetcher;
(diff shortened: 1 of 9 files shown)


src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -260,7 +241,7 @@ public Screen getNextScreen(boolean needsFullEntityScreen, EntityScreenContext e
             return entityScreen;
         } else if (next.equalsIgnoreCase(SessionFrame.STATE_DATUM_COMPUTED)) {
             computeDatum();
-            return getNextScreen();
+            return getNextScreen(needsFullEntityScreen, entityScreenContext);
```
shubham1g5, line 244: this is the main fix in context of this issue, the entityContext was earlier getting re-initialised instead of us reusing the entityContext.

avazirna (approved):

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -125,6 +127,7 @@ public class MenuSessionRunnerService {
 
     private static final Log log = LogFactory.getLog(MenuSessionRunnerService.class);
 
+    @VisibleForTesting
```
snopoke, line 130: can tests not just use the other method and pass in the entity context?

PR #1402, opened 2023-06-12
increase file upload size limit to 4MB

Robert-Costello:
## Technical Summary

jira ticket: https://dimagi-dev.atlassian.net/browse/USH-3276
HQ PR: https://github.com/dimagi/commcare-hq/pull/33099

this increases the maximum file upload size to 4MB

### Safety story

small change to size limit

### Rollback instructions

if this PR is reverted the corresponding HQ PR should also be reverted. https://github.com/dimagi/commcare-hq/pull/33099

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MediaValidator.kt
+++ b/src/main/java/org/commcare/formplayer/services/MediaValidator.kt
@@ -11,8 +11,8 @@ import java.net.URLConnection
  */
 object MediaValidator {
 
-    // 3 MB size limit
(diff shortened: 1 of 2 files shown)


shubham1g5 (changes requested):
We also need to change the [spring config](https://github.com/dimagi/formplayer/blob/master/src/main/resources/application.properties#L78-79)

Robert-Costello:
> We also need to change the [spring config](https://github.com/dimagi/formplayer/blob/master/src/main/resources/application.properties#L78-79)

the `max-request-size` should be 5MB, right?

shubham1g5 (approved):

PR #1455, opened 2023-09-01
Remove obsolete backward compatibility flag for case search dropdowns

shubham1g5:
## Technical Summary

remove backward compatibility flag introduced as part of the [spec](https://docs.google.com/document/d/1Xtly4YfrdaB1w1Ozwlwy9LQ4JIrVRmhvg-44-Rl-z6A/edit) as all response to FP should have it set to `true` now. 

### Safety story

Removes a obsolete flag
tested on staging. 

### Automated test coverage
Updated tests in the PR to test for new behaviour

### QA Plan

No QA

### Rollback instructions

### Review

cross-request: https://github.com/dimagi/commcare-core/pull/1325

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -109,7 +109,7 @@ dependencies {
     implementation group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-jdk8'
 
     //Unmanaged Dependencies
-    implementation group: 'com.datadoghq', name: 'dd-trace-api', version: '1.19.2'
+    implementation group: 'com.datadoghq', name: 'dd-trace-api', version: '1.19.3'
     implementation group: 'com.datadoghq', name: 'java-dogstatsd-client', version: '4.2.0'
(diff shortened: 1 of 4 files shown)


Jtang-1 (approved):

Jtang-1:
Can you remind me why we needed this flag in the first place? After the HQ changes were released and the remaining FP changes were done, HQ and Formplayer would have both working with keys right? What's the transitory period where FP still needed to support indexes?

Is it the transient period where a user is already in a form and then the "key" support changes of Formplayer gets deployed?

shubham1g5:
> Is it the transient period where a user is already in a form and then the "key" support changes of Formplayer gets deployed?

Yes the transient period was very short only to support requests in close proximity of deploy (When user is already on case search screen before and after the FP deploy).

snopoke (approved):

PR #1458, opened 2023-09-06
use instance_id_ variables For Multi Select Endpoints

shubham1g5:
## Technical Summary
This is a modification on top of the [multi-select session endpoints support](https://docs.google.com/document/d/1nw-MbFvlMR5FQCpmQlHdmmJYQ-3e5BnMwx-epjumP9g/edit#heading=h.u0xbvhmgljhi). Instead of substituting the instance arguments directly, use a prefix `instance_id_argname` to signify that we want to substitue the argument with corresponding instance guid. 

### Migrations

### Rollback instructions

### Review

cross-request: https://github.com/dimagi/commcare-core/pull/1327

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -754,7 +754,7 @@ public class MenuSessionRunnerService {
                 if (instanceSrc.contentEquals(JR_SELECTED_ENTITIES_REFERENCE)) {
                     String[] selectedEntites = argumentValue.split(",");
                     String uuid = storeInSelectedEntitiesInstance(argument, selectedEntites);
-                    argumentValues.put(argument.getId(), uuid);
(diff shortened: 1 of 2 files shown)


shubham1g5:
@snopoke This needs cleanup but I kind of just want to get a OK on the approach here before making further changes here and on HQ. This concern came up whille working on [inline case search changes](https://docs.google.com/document/d/1owt1WE91-8jF7yUnzlbk5jB0YC8Uqeg3I3yTQy5vuVo/edit#heading=h.59613tyr73dz) which might need us to retain both the original argument value along with the substituted instance argument and I would prefer to keep the actual argument name as it is.

snopoke (commented):
I don't really like this approach because it depends on some internal naming convention in the engine.

I made this comment on the spec as well but could not use the instance to get the list of case IDs in the query:

```xml
<data key="case_id" nodeset="instance('selected_cases')/results/value" ref="."/>
```

PR #1459, opened 2023-09-06
clear media answer

Robert-Costello:
## Technical Summary

jira ticket: https://dimagi-dev.atlassian.net/browse/USH-3582
spec doc: https://docs.google.com/document/d/1lRYWe-IRUWN7OTYK9mVvGYhPoy_Wo85NnF0dooSw1rY/edit#heading=h.d7s2sycdfk92
hq PR: https://github.com/dimagi/commcare-hq/pull/33443

### Safety story

changes restricted to feature flag
tested locally and on staging

### QA Plan

no QA planned

### Rollback instructions

If this PR is reverted the related hq PR should also be reverted, as `clear_answer` requests will fail.

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/FormController.java
+++ b/src/main/java/org/commcare/formplayer/application/FormController.java
@@ -204,8 +204,7 @@ public class FormController extends AbstractBaseController {
         String fileId = null;
         Path mediaDirPath = formEntrySession.getMediaDirectoryPath(restoreFactory.getDomain(),
                 restoreFactory.getUsername(), restoreFactory.getAsUsername(), storageFactory.getAppId());
-
-        if (clear) {
+        if (Boolean.TRUE.equals(clear)) {
             formEntrySession.cleanCurrentMedia(mediaDirPath, answerQuestionBean.getFormIndex(),
                     mediaMetaDataService);
         }
--- a/src/main/java/org/commcare/formplayer/application/FormController.java
+++ b/src/main/java/org/commcare/formplayer/application/FormController.java
@@ -185,7 +185,7 @@ public class FormController extends AbstractBaseController {
     }
 
     private FormEntryResponseBean saveAnswer(AnswerQuestionRequestBean answerQuestionBean,
-            @Nullable MultipartFile file, @Nullable Boolean clear) throws Exception {
+            @Nullable MultipartFile file, Boolean clear) throws Exception {
 
         SerializableFormSession serializableFormSession = categoryTimingHelper.timed(
                 Constants.TimingCategories.GET_SESSION,
@@ -204,7 +204,7 @@ public class FormController extends AbstractBaseController {
         String fileId = null;
         Path mediaDirPath = formEntrySession.getMediaDirectoryPath(restoreFactory.getDomain(),
                 restoreFactory.getUsername(), restoreFactory.getAsUsername(), storageFactory.getAppId());
-        if (Boolean.TRUE.equals(clear)) {
+        if (clear) {
             formEntrySession.cleanCurrentMedia(mediaDirPath, answerQuestionBean.getFormIndex(),
                     mediaMetaDataService);
         }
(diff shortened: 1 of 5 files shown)


shubham1g5 (changes requested):
Code looks great, but need tests!

src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -193,6 +204,11 @@ private FormEntryResponseBean saveAnswer(AnswerQuestionRequestBean answerQuestio
         String fileId = null;
         Path mediaDirPath = formEntrySession.getMediaDirectoryPath(restoreFactory.getDomain(),
                 restoreFactory.getUsername(), restoreFactory.getAsUsername(), storageFactory.getAppId());
+
+        if (clear) {
```
MartinRiese, line 208: if `clear` is `@Nullable` this could cause a `NullPointerException`

MartinRiese (changes requested):

src/main/java/org/commcare/formplayer/application/FormController.java
Robert-Costello: handled here https://github.com/dimagi/formplayer/pull/1459/commits/9af5cf36ff32d6e670942af67904d596f29d235e

shubham1g5: booleans doesn't need to `@Nullable`, they have a definite value of `true` or `false`

Robert-Costello: ok, so we'll just keep the `if (cear)` check and take away the `@Nullable' annotation.

MartinRiese: @shubham1g5 `boolean` is `true` or `false` but `Boolean` can be `null`.

shubham1g5: that is right, think we can simply use the primitive `boolean` here. 

Even if we use `Boolean`, they can't be null in this context of code as we are always passing a non null value.  So it seems ok to remove the `@Nullable` annotation and not do the null handling.

src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -170,11 +170,22 @@ public FormEntryResponseBean answerMediaQuestion(
...
             throws Exception {
-        return saveAnswer(answerQuestionBean, file);
+        return saveAnswer(answerQuestionBean, file, false);
+    }
+
+    @RequestMapping(value = Constants.URL_CLEAR_ANSWER, method = RequestMethod.POST)
+    @UserLock
+    @UserRestore
+    @ConfigureStorageFromSession
+    public FormEntryResponseBean removeAnswer(
```
shubham1g5, line 180: nit: rename to `clearAnswer` to be consistent with the api endpoint

shubham1g5 (changes requested):

src/test/java/org/commcare/formplayer/tests/MediaCaptureTest.kt
```diff
@@ -235,6 +236,19 @@ class MediaCaptureTest {
...
+    fun testClearMedia() {
+        val formResponse = startImageCaptureForm()
+        val imageResponse = saveImage(formResponse, "media/valid_image.jpg", "valid_image.jpg")
+
+        var expectedFilePath = getExpectedMediaPath(formResponse.session_id, imageResponse)
+        val originalSavedFile = expectedFilePath.toFile()
+        assertTrue("Could not find saved file on the filesystem", originalSavedFile.exists())
+
+        clearImage(formResponse)
+        assertFalse("Could not remove file from the filesystem", originalSavedFile.exists())
```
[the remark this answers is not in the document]
Robert-Costello: https://github.com/dimagi/formplayer/pull/1459/commits/2baf39615e19de5bd8b81fa6fa00a095b2e04561

shubham1g5 (approved):
Good to go 💯

PR #1468, opened 2023-09-26
Handle Field Action Inputs

shubham1g5:
## Product Description
[Spec](https://docs.google.com/document/d/1coxsFVCYgaQdxE3rgYikRSgEgistrfIUJuoW5KMiM5w/edit)

## Safety Assurance

Covered By tests, small change. 

### Automated test coverage

Covered as part of PR

### QA Plan

No QA 

### Rollback instructions

### Review

cross-request: https://github.com/dimagi/commcare-core/pull/1338/files

### DIFF
--- a/src/test/resources/archives/multi_select_case_list/suite.xml
+++ b/src/test/resources/archives/multi_select_case_list/suite.xml
@@ -46,6 +46,20 @@
           <xpath function="case_name"/>
         </text>
       </sort>
+      <action>
+        <display>
+          <text>
+            <locale id="forms.m0f0"/>
+          </text>
+        </display>
+        <stack>
+          <push>
+            <command value="'m0-f0'"/>
+            <datum id="case_id_new_case_0" value="uuid()"/>
+            <datum id="return_to" value="'m0'"/>
+          </push>
+        </stack>
+      </action>
     </field>
     <action>
       <display>
--- a/src/main/java/org/commcare/formplayer/util/PrototypeUtils.java
+++ b/src/main/java/org/commcare/formplayer/util/PrototypeUtils.java
@@ -17,6 +17,7 @@ import org.commcare.resources.model.installers.MediaInstaller;
 import org.commcare.resources.model.installers.ProfileInstaller;
 import org.commcare.resources.model.installers.SuiteInstaller;
 import org.commcare.resources.model.installers.XFormInstaller;
+import org.commcare.suite.model.Action;
 import org.commcare.suite.model.ComputedDatum;
 import org.commcare.suite.model.EntityDatum;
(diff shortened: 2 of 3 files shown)


src/test/resources/archives/multi_select_case_list/suite.xml
```diff
@@ -46,6 +46,20 @@
...
+      <action>
+        <display>
+          <text>
+            <locale id="forms.m0f0"/>
+          </text>
+        </display>
+        <stack>
+          <push>
+            <command value="'m0-f0'"/>
+            <datum id="case_id_new_case_0" value="uuid()"/>
```
snopoke, line 58: @shubham1g5 how will we get the case ID of the selected case? In the HQ PR I've used `current()/@case_id` which would require setting the 'current' entity in the context. In the tests here I don't see any 'selected case' in the selections so wondering how that's going to work.

shubham1g5: hmm, that defintely looks like a gap in the current feature design. I would think that we need to pass the case id in the selections array along with the action to have the case selected and part of the cc session for ex-  `[m0, caseid, field_action 1]` . Although after the case selection we no longer are on EntityScreen which means we won't be able to handle associated actions. 

Therefore we need some other mechanism to convey case selection here without actually selecting the case on the entity screen. Endpoints handle it by taking case ids as arguments but we don't have argument support in Actions. 

I wonder whether it makes sense for us to use Endpoints here instead of Actions. On clicking the icon, we can fire a session endpoint associated with a auto-submit form with the case id as argument without Web Apps navigating away from UI. 

As an alternative, we can have Web Apps pass the case id in an additional field in request body in the action navigation request and add arguments support to actions simiar to endpoints. I don't think we necessarily need to support a construct like `current()` in Actions though as stack ops only have singular datum values and it might be complicated to construct `current()` if the datum in question is not a case id but for ex. a fixture id.

snopoke: This is one of the reasons we'd thought of using a separate endpoint which would function in a similar way to the 'details' endpoint.

> I wonder whether it makes sense for us to use Endpoints here instead of Actions. On clicking the icon, we can fire a session endpoint associated with a auto-submit form with the case id as argument without Web Apps navigating away from UI.

We did consider this but we'd have to change how endpoints work quite a bit.

shubham1g5: I have put down some thoughts on using session endpoints in [the spec](https://docs.google.com/document/d/1coxsFVCYgaQdxE3rgYikRSgEgistrfIUJuoW5KMiM5w/edit#heading=h.13478xeglcy4)

PR #1484, opened 2023-10-26
Hidden forms support for Smart links

shubham1g5:
## Product Description
[JIRA](https://dimagi-dev.atlassian.net/browse/USH-3742)

Clickabls icons workflow uses session endpoints underhood to launch a form in the background. Although we don’t want these forms to be visible in the normal user navigation of the app but only through the icon click on the case list.

## Technical Summary

Uses endpoint `respect-relevancy` attribute to determine whether to respect menu relevancy conditions for a session endpoint navigation.

### Safety story

- Even though changes touches all session navigation code, they are limited by the `respectRelevancy` flag which is only set to `false` for Session Endpoints. As such the changes should not affect anything outside the smart links. 

- Locally tested

### Automated test coverage

Session navigation is well covered by tests. This PRs adds a test for new bhaviour around session endpoint relevancy. 

### QA Plan

Not planning

### Rollback instructions

### Review

cross-request: https://github.com/dimagi/commcare-core/pull/1365


snopoke:
@shubham1g5 could you give a bit more info about the design / use case for this. There's no info in the ticket.

shubham1g5:
@snopoke Added the usecase in Product description of PR.

snopoke:
> @snopoke Added the usecase in Product description of PR.

Thanks. So basically endpoints being used for click actions must have `respect-relevancy` set to `false` so that they will work even though the form is not 'relevant'.

Is that correct?

shubham1g5:
Yes exactly. We plan to give a UI config for session endpoint to enable this - https://dimagi-dev.atlassian.net/browse/USH-3744 so that app builders can decide whether they want their session endpoint to be able to access hidden forms or not.

snopoke (approved):

PR #1491, opened 2023-11-06
Clickable Icons: Clear cache search cache for get_details request based on a request flag

shubham1g5:
## Product Description
https://dimagi-dev.atlassian.net/browse/QA-5771

Fixes issue where get_details request doesn't give latest data after performing a case change using clickable icon due to cache search cache not refreshing.

## Technical Summary

Adds a flag called `is_refresh_case_search` to request beans and clear cache search cache if it's set to `true`

### Safety story
Should only affect clickable icons feature when we set the `isRefreshCaseSearch` to true for `get_details`

### Automated test coverage
PR adds a test to verify new bheaviour

### QA Plan
None

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/SessionNavigationBean.java
+++ b/src/main/java/org/commcare/formplayer/beans/SessionNavigationBean.java
@@ -27,6 +27,8 @@ public class SessionNavigationBean extends InstallRequestBean {
     private String[] selectedValues;
     private boolean isShortDetail;
 
+    private boolean isRefreshCaseSearch;
+
     /**
      * Form session ID used to prevent attempts to navigate into a form session
      * e.g. pressing the back button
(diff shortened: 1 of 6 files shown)


src/main/java/org/commcare/formplayer/beans/SessionNavigationBean.java
```diff
@@ -118,6 +120,9 @@ public void setIsPersistent(boolean persistent) {
     public boolean getIsShortDetail() {
         return isShortDetail;
     }
+    public boolean getIsRefreshCaseSearch() {
```
MartinRiese, line 123: I thought standard naming leaves the get off the name if the field is a boolean.

MartinRiese (approved):
Just a naming nit

src/main/java/org/commcare/formplayer/beans/SessionNavigationBean.java
shubham1g5: yea, although the naming for response beans is a little different as  this gets used by the Json coverter to convert a json for this class and the standand naming convention there is `get_{$var}` and `set_{$var}`

Robert-Costello: Does this need a setter or does it function differently from `isShortDetail`?

shubham1g5: it does, corrected

PR #1502, opened 2023-11-22
notify hq that case is not in case list

Robert-Costello:
### Migrations

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/MenuController.java
+++ b/src/main/java/org/commcare/formplayer/application/MenuController.java
@@ -130,7 +130,12 @@ public class MenuController extends AbstractBaseController {
         TreeReference reference = entityScreen.resolveTreeReference(detailSelection);
 
         if (reference == null) {
-            throw new RuntimeException("Could not find case with ID " + detailSelection);
+            if (sessionNavigationBean.getIsRefreshCaseSearch()) {
+                EntityDetailListResponse removeCaseRowResponse = new EntityDetailListResponse(true);
+                return removeCaseRowResponse;
+            } else {
+                throw new RuntimeException("Could not find case with ID " + detailSelection);
+            }
         }
 
         restoreFactory.cacheSessionSelections(selections);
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/formplayer/application/MenuController.java
```diff
@@ -130,7 +130,12 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
         TreeReference reference = entityScreen.resolveTreeReference(detailSelection);
 
         if (reference == null) {
-            throw new RuntimeException("Could not find case with ID " + detailSelection);
+            if (sessionNavigationBean.getIsRefreshCaseSearch()) {
```
shubham1g5, line 133: 1. I don't like that the api behaviour depends on an irrelevant flag here. We should be able to use same response anytime we can't find the case. 
2. Probably the simplest change to make here is to throw a 404 from FP instead of a flag to denote that case does not exisit and handle it on Web Apps accordingly. 

Instead of throwing a RuntimeException we can simply use this - 
`throw new ResponseStatusException(NOT_FOUND, "Could not find case with ID " + detailSelection);`

@MartinRiese @snopoke FYI in case if you have any opposing thoughts on this approach.

Robert-Costello: My thinking for using `getIsRefreshCaseSearch` here was to localize changes to this error response to clickable icons requests, and possible replace `getIsRefreshCaseSearch` with a separate flag like `isIconRequest`. Throwing a 404 whenever the case isn't found and dealing with that on HQ side makes sense, though. 
We'd only remove the case row for clickable icon requests, correct?

shubham1g5: > We'd only remove the case row for clickable icon requests, correct?
right.

Robert-Costello:
closing out in favor of https://github.com/dimagi/formplayer/pull/1507

PR #1507, opened 2023-11-27
Rc/clickable icons remove case from UI

Robert-Costello:
## Product Description

This improves on error for not found cases and marks requests as coming from clickable icons to allow for filtering case list by clickable icon properties.
## Technical Summary

jira ticket: https://dimagi-dev.atlassian.net/browse/USH-3875
mini spec: https://docs.google.com/document/d/11oJjcL5LnP0gwHOBuOZMGqcjkaRnekukbLN3K6vWxPo/edit

### Safety story

tested locally and on staging

### Rollback instructions

### Review

### DIFF
--- a/src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
+++ b/src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
@@ -703,8 +703,7 @@ public class BaseTestClass {
     }
 
     <T> T getDetails(String[] selections, String testName, String locale, QueryData queryData, Class<T> clazz,
-            boolean inline, boolean isShortDetail, boolean isRefreshCaseSearch,
-            boolean isClickableIcon) throws Exception {
+            boolean inline, boolean isShortDetail, boolean isRefreshCaseSearch) throws Exception {
         SessionNavigationBean sessionNavigationBean = new SessionNavigationBean();
         sessionNavigationBean.setDomain(testName + "domain");
         sessionNavigationBean.setAppId(testName + "appid");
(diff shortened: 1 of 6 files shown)


src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -713,6 +714,7 @@ <T> T getDetails(String[] selections, String testName, String locale, QueryData
         sessionNavigationBean.setQueryData(queryData);
         sessionNavigationBean.setIsShortDetail(isShortDetail);
         sessionNavigationBean.setIsRefreshCaseSearch(isRefreshCaseSearch);
+        sessionNavigationBean.setIsClickableIcon(isClickableIcon);
```
shubham1g5, line 717: what is the purpose of this ? I don't think FP cares whether a request is clickable icon or not.

Robert-Costello: The purpose is to distinguish between get_details requests from clickable icons and other sources. That way we are only removing cases from the ui when clickable icons remove them from the case list. In other cases the 404 is displayed.

Robert-Costello: >I think that it can be handled on Web Apps instead without needing to make a FP change for it.

There may be a way to do this. @MartinRiese any thoughts?

>Doesn't web apps know whether the response it's handling is for clickable icons request ?

As far as I can tell, it can't tell that from the variables it currently has access to.

Robert-Costello: I removed the flag and set something up on the HQ side that serves the same function.

PR #1512, opened 2023-12-15
Submit Button Translation

stephherbers:
## Product Description

This goes with this [HQ pr](https://github.com/dimagi/commcare-hq/pull/33868/files)

## Technical Summary

This was a workaround for not modifying the suite or Xforms

### Safety story

The HQ side that works with that code fail gracefully if that key is not available

### QA Plan

I am not requesting QA

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -55,6 +55,7 @@ import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.ExternalDataInstance;
 import org.javarosa.core.model.instance.ExternalDataInstanceSource;
 import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.services.locale.Localization;
 import org.javarosa.xml.util.InvalidStructureException;
 import org.javarosa.xml.util.UnfullfilledRequirementsException;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -670,7 +671,12 @@ public class MenuSessionRunnerService {
         NewFormResponse response = newFormResponseFactory.getResponse(formEntrySession);
 
         String commandId = menuSession.getSessionWrapper().getCommand();
-        newFormResponseFactory.setTranslations(response, commandId);
+        String modifiedCommandId = commandId.replace("-", "");
+        String submit_label_key = "forms." + modifiedCommandId + ".submit_label";
+        String translation = Localization.getWithDefault(submit_label_key, null);
+        if (translation != null) {
+            response.addToTranslation(key, translation);
+        }
 
         response.setNotification(establishVolatility(formEntrySession));
         response.setShouldAutoSubmit(formEntrySession.getAutoSubmitFlag());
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -668,6 +668,10 @@ private NewFormResponse generateFormEntrySession(MenuSession menuSession) throws
                 formDefinitionService);
 
         NewFormResponse response = newFormResponseFactory.getResponse(formEntrySession);
+
+        String commandId = menuSession.getSessionWrapper().getCommand();
+        newFormResponseFactory.setTranslations(response, commandId);
+
```
shubham1g5, line 674: Why not init translations in ` newFormResponseFactory.getResponse` as before ?

shubham1g5 (approved):
I am a bit mixed on this approach because we are hard-coding the app translations string on FP side, generally strings like these are defined on HQ in the app ccz itself. 

I am also curious if you have had a chance to talk to @ctsims  on it as I see some thoughts from Clayton [on this ticket](https://dimagi-dev.atlassian.net/browse/USH-3885)

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
stephherbers: I thought this was cleaner to pass in `commandId` just where it's needed rather than pass into `NewFormResponse`

stephherbers:
> hard-coding the app translations string on FP side

I dug some more I found it unnecessary with some HQ changes so I modified to return null like the others if not translation found

stephherbers:
> I am also curious if you have had a chance to talk to @ctsims on it as I see some thoughts from Clayton [on this ticket](https://dimagi-dev.atlassian.net/browse/USH-3885)

@shubham1g5 I chatted with Jenny about that ticket and it seems we'll have to go with a totally different approach for the ticket with a higher LOE so opted for this route for the Jan 1st deadline

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5: I would think otherwise here as the response **_factory_** should be responsible to form complete response without cluttering the calling service with additional calls.

orangejenny:
(I meant to put those thoughts from Clayton on [this ticket](https://dimagi-dev.atlassian.net/browse/USH-3632) instead. Copied them over.)

Robert-Costello (approved):
Other than the branch conflicts this looks good to me.

Jtang-1 (approved):

Jtang-1:
Discussed with Steph beforehand that would merge for her behalf if review looks good

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -668,6 +669,15 @@ private NewFormResponse generateFormEntrySession(MenuSession menuSession) throws
...
 
         NewFormResponse response = newFormResponseFactory.getResponse(formEntrySession);
+
+        String commandId = menuSession.getSessionWrapper().getCommand();
+        String modifiedCommandId = commandId.replace("-", "");
+        String submit_label_key = "forms." + modifiedCommandId + ".submit_label";
+        String translation = Localization.getWithDefault(submit_label_key, null);
+        if (translation != null) {
+            response.addToTranslation(submit_label_key, translation);
+        }
```
shubham1g5, line 679: this should be abstracted into a new method for clarity.

PR #1519, opened 2024-01-03
Do not skip cache for page changes

shubham1g5:
## Technical Summary

Prompted by [the issue here](https://docs.google.com/document/d/1y-G9PH3c3FFupRawIYTeaX_cehc-hGHQomu1BwnB3gY/edit#heading=h.7vsnlyupdpde)

When shifting from page 1 to page 2 of a case list backed by case search, we currently redo case search request because of `skipCache` set to true. (This is a side effect of [the fix](https://github.com/dimagi/formplayer/pull/1172) that was implemented for https://dimagi-dev.atlassian.net/browse/USH-2171.)  This PR makes a change so that if we are not requesting the first page of the entity list, we keep `skipCache` set to `false` and not redo the case search. 

Caveating that a better solution here can be for Web Apps to send a flag in the request say 'pageChange` based on which FP can know that it was only a page change from user end and it should not skip the case search cache.  

## Safety Assurance

- Should only affect entity list pagination 
- Workflow should be unaffected as the only impact the change have is whether we do the case search request again or not.

### Automated test coverage
None around the specific change but the workflow is covered by a number of tests. Possibly a test here can be to verify whether we used cache or not when doing a request for `offset!=0` ?

### QA Plan
NA

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -377,7 +377,8 @@ public class MenuSessionRunnerService {
                 }
             } else if (nextScreen instanceof FormplayerQueryScreen) {
                 boolean replay = !nextInput.equals(NO_SELECTION);
-                boolean skipCache = !(replay || entityScreenContext.getDetailSelection() != null);
+                boolean skipCache = !(replay || entityScreenContext.getDetailSelection() != null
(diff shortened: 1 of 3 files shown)


Robert-Costello (approved):

shubham1g5:
Merging this as it is but we might wanna revisit this to have a better indication from Web Apps around pagination changes rather than FP trying to deduce it with the help of `offset!=0`.

PR #1526, opened 2024-01-08
Adds metadata in request response 

shubham1g5:
## Technical Summary

Adds metadata in request response  to indicate if the request went through a restore or app install

Screenshot 2024-01-08 at 12 31 55 PM

This is to help in adding visibility to requests that takes more than normal time once in a while. 

### Safety story

Locally tested

### QA Plan
No

### Rollback instructions

### Review


Jtang-1 (approved):
makes sense to me, but also may want input from someone more familiar with formplayer.

Robert-Costello (approved):

PR #1527, opened 2024-01-08
No details if only a popup field is defined

MartinRiese:
## Product Description

The details should not be shown if only an address popup field is defined

## Technical Summary

https://dimagi-dev.atlassian.net/browse/USH-4008

### Rollback instructions

### Review


shubham1g5 (approved):

shubham1g5:
One question - What if a case list has other detail fields along with `address-popup` field ? We would want to remove the address-popup data from the normal detail dialog in that case or does Web Apps take care of filtering that data ?

PR #1531, opened 2024-01-16
Adds searchOnClear to queryResponse

Jtang-1:
## Product Description

See description in [HQ PR 
](https://github.com/dimagi/commcare-hq/pull/33983)

## Technical Summary

[Jira Ticket](https://dimagi-dev.atlassian.net/browse/USH-4046)
[Commcare Core PR](https://github.com/dimagi/commcare-core/pull/1397)
[HQ PR 
](https://github.com/dimagi/commcare-hq/pull/33983)

### Automated test coverage

added test that attribute is correctly parsed from suite

### QA Plan

no QA 

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
+++ b/src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
@@ -9,8 +9,6 @@ import java.util.HashMap;
 
 import org.javarosa.core.services.locale.Localization;
 
-import io.micrometer.core.lang.Nullable;
-
 /**
  * Base class for responses being sent to the front end. Params are: title - self explanatory
  * notification - A message String and error boolean to be displayed by frontend sholdAutoSubmit - A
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/formplayer/beans/menus/BaseResponseBean.java
```diff
@@ -9,6 +9,8 @@
 
 import org.javarosa.core.services.locale.Localization;
 
+import io.micrometer.core.lang.Nullable;
```
Robert-Costello, line 12: should this be `import org.springframework.lang.Nullable;`?

Jtang-1: good catch, I can't even remember where i got that from.

It actually can't be null. When parsed, if the attribute isn't present, the value is saved as `false`. I removed it in [64d31a8](https://github.com/dimagi/formplayer/pull/1531/commits/64d31a866dd2699c7b8b1ddd00dd2f23f9ed0b59). I also changed it so it's only in the queryResponse since it's not applicable elsewhere.

Robert-Costello (approved):

Jtang-1:
duplicate this PR 04b97fe 402dca4

PR #1533, opened 2024-01-22
Adds DD tag for Case Search Input

Jtang-1:
## Product Description

Adds a `module_initiated_by` tag to Datadog. This value is passed by HQ to signify what initiated the query. Specifically in this case, we are interested in queries triggered via user search input for case search. These will be tagged as `field_change`.

Screen Shot 2024-01-22 at 9 24 54 AM

This additionally includes `module_name` tags for requests tagged as `case_search` module 

## Technical Summary

[Commcare HQ ](https://github.com/dimagi/commcare-hq/pull/34006)
[Jira Ticket](https://dimagi-dev.atlassian.net/browse/USH-4051)

### Safety story

Locally tested and tested on staging

### QA Plan

no QA 

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/objects/QueryData.java
+++ b/src/main/java/org/commcare/formplayer/objects/QueryData.java
@@ -52,6 +52,11 @@ public class QueryData extends Hashtable<String, Object> {
         return null;
     }
 
+    public void setInitiatedBy(String key, String value) {
+        this.initKey(key);
+        ((Map<String, Object>) this.get(key)).put(INITIATED_BY, value);
+    }
+
     public Hashtable<String, String> getInputs(String key) {
         Map<String, Object> value = (Map<String, Object>) this.get(key);
         if (value != null) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/formplayer/objects/QueryData.java
```diff
@@ -40,6 +41,17 @@ public void setForceManualSearch(String key, Boolean value) {
         setProperty(key, value, KEY_FORCE_MANUAL_SEARCH);
     }
 
+    public String getInitiatedBy(String key) {
+        Map<String, Object> value = (Map<String, Object>) this.get(key);
+        if (value != null) {
+            String initiatedBy = (String) value.get(INITIATED_BY);
+            if (initiatedBy != null) {
```
Robert-Costello, line 48: nit: This null check seems unnecessary to me since you will return `null` anyway if `initiatedBy` is `null`.  maybe it's more readable with the check?

Jtang-1: yes you're right [e943ab5](https://github.com/dimagi/formplayer/pull/1533/commits/e943ab577db1c703cfdd8471442e9fb98937e7cc)

Robert-Costello (approved):

PR #1538, opened 2024-02-06
Adds datadog tags for restore and app installs

Jtang-1:
## Product Description

No visual change.

## Technical Summary

Adds a `metadata` tag to Datadog where the value can include `app_install` or `attempt_restore`. These processes can lead to slow response times so the tags are added so those requests can be filtered and analyzed independently. 

Screen Shot 2024-02-07 at 10 15 57 AM

Related [PR](https://github.com/dimagi/formplayer/pull/1526) which added this metadata to the response.
[USH-4069](https://dimagi.atlassian.net/browse/USH-4069)

### Safety story

Tested on staging. No change to product behavior.

### QA Plan

no QA

### Rollback instructions

### Review

[USH-4069]: https://dimagi.atlassian.net/browse/USH-4069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/ResponseMetaDataTracker.java
+++ b/src/main/java/org/commcare/formplayer/services/ResponseMetaDataTracker.java
@@ -23,7 +23,7 @@ public class ResponseMetaDataTracker {
     }
 
     public void setAttemptRestore(boolean attemptRestore) {
-        datadog.addRequestScopedTag(Constants.METADATA_TAG, "attempt_restore");
+        datadog.addRequestScopedTag(Constants.CATEGORY_TAG, Constants.TimingCategories.COMPLETE_RESTORE);
         this.attemptRestore = attemptRestore;
     }
 
@@ -32,7 +32,7 @@ public class ResponseMetaDataTracker {
     }
 
     public void setNewInstall(boolean newInstall) {
-        datadog.addRequestScopedTag(Constants.METADATA_TAG, "app_install");
+        datadog.addRequestScopedTag(Constants.CATEGORY_TAG, Constants.TimingCategories.APP_INSTALL);
         this.newInstall = newInstall;
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/formplayer/services/ResponseMetaDataTracker.java
```diff
@@ -12,11 +14,15 @@ public class ResponseMetaDataTracker {
...
 
+    @Autowired
+    private FormplayerDatadog datadog;
+
     public boolean isAttemptRestore() {
         return attemptRestore;
     }
 
     public void setAttemptRestore(boolean attemptRestore) {
+        datadog.addRequestScopedTag("metadata", "attempt_restore");
```
nospame, line 25: I don't think this needs a test like Codecov insists, but seems like it probably should use/add a tag as a constant for "metadata" here: https://github.com/dimagi/formplayer/blob/42cea06991728fa896198f31602a62e3850a70c2/src/main/java/org/commcare/formplayer/util/Constants.java#L124

Jtang-1: Good call [d2d609f](https://github.com/dimagi/formplayer/pull/1538/commits/d2d609f429396cde84cfae1a4740fb8aa0a79c31)

I'm also open to any tag name suggestion. I'm not sure if "metadata" is the best since all tags are in a sense metadata

nospame: Is this usage necessarily/distinctly different from the `CATEGORY_TAG`? That's the only other place I would maybe see it fitting.

Robert-Costello (approved):

PR #1539, opened 2024-02-07
Clear case search db when clearing user data

stephherbers:
## Product Description

Written by @Robert-Costello 
P2 ticket: https://dimagi.atlassian.net/browse/USH-4181

## Technical Summary

Adds to what is cleared when syncing user data as a workaround for stale data in case search. (now if stale data is shown they can clear user data to see the updated results)

### Safety story

This has been tested on staging by support, tech, and delivery members

### QA Plan

No QA has been requested

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/UtilController.java
+++ b/src/main/java/org/commcare/formplayer/application/UtilController.java
@@ -13,16 +13,20 @@ import org.commcare.formplayer.beans.NotificationMessage;
 import org.commcare.formplayer.beans.ServerUpBean;
 import org.commcare.formplayer.beans.SyncDbRequestBean;
 import org.commcare.formplayer.beans.SyncDbResponseBean;
+import org.commcare.formplayer.services.CaseSearchHelper;
 import org.commcare.formplayer.services.CategoryTimingHelper;
 import org.commcare.formplayer.services.FormSessionService;
 import org.commcare.formplayer.services.FormplayerLockRegistry;
 import org.commcare.formplayer.services.RestoreFactory;
+import org.commcare.formplayer.sqlitedb.CaseSearchDB;
 import org.commcare.formplayer.sqlitedb.UserDB;
 import org.commcare.formplayer.util.Constants;
 import org.commcare.formplayer.util.NotificationLogger;
+import org.javarosa.core.model.instance.ExternalDataInstanceSource;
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/application/UtilController.java
```diff
@@ -102,7 +106,12 @@ public NotificationMessage deleteApplicationDbs(
     @UserLock
     public NotificationMessage clearUserData(
             @RequestBody AuthenticatedRequestBean requestBean,
-            HttpServletRequest request) {
+            HttpServletRequest request) throws InvalidStructureException {
+
+        // Test clearing case db
```
gherceg, line 111: nit: can this comment be removed? Or is this still in the testing phase?

gherceg (approved):
No need to address the nit now or ever. Looks good.

src/main/java/org/commcare/formplayer/application/UtilController.java
Robert-Costello: i'll remove it

src/main/java/org/commcare/formplayer/application/UtilController.java
```diff
@@ -102,7 +106,12 @@ public NotificationMessage deleteApplicationDbs(
...
+        CaseSearchDB caseSearchDB = new CaseSearchDB(requestBean.getDomain(), requestBean.getUsername(),
+                requestBean.getRestoreAs());
+        caseSearchDB.deleteDatabaseFile();
```
ctsims, line 114: This introduces a new error handling path which will invert control out of the method since it's added to the method signature. That means that an issue with the database clearing will theoretically block the existing functionality. 

I'm fine with the return path of the InvalidStructureException causing a hard fault (that's a very unexpected result from the code), but I'd prefer for the new lines to be placed after the old ones, so a failure in the new functionality is an at-worst-neutral change for users

ctsims (changes requested):

src/main/java/org/commcare/formplayer/application/UtilController.java
Robert-Costello: moved new lines here https://github.com/dimagi/formplayer/pull/1539/commits/02fcbc457d72b791746fe55ab720a044feb4fba5

ctsims (approved):

PR #1541, opened 2024-02-12
sort params to keep cache key consistent

Robert-Costello:
## Technical Summary

As per discussion about better cache handling, this sorts the query params key set values that are used to create keys for case search caches. There is no ordering guarantee with `MultiMap` key sets. This will avoid possible issues with cache key inconsistency.
### Safety story

This doesn't change the fundamental structure of the cache keys, just the ordering of the params. I believe any pre-existing cache key will be removed by a Formplayer deploy, meaning that there should be no issue where FP attempts to reference a cache key with the old ordering. 

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
+++ b/src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
@@ -40,6 +40,9 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
 
 @CacheConfig(cacheNames = "case_search")
 @Component
@@ -209,7 +212,9 @@ public class CaseSearchHelper {
             builder.append("_").append(restoreFactory.getAsUsername());
         }
         builder.append("_").append(uri);
-        for (String key : queryParams.keySet()) {
+        List<String> keyList = queryParams.keySet().stream().collect(Collectors.toList());
+        Collections.sort(keyList);
+        for (String key : keyList) {
             builder.append("_").append(key);
             for (String value : queryParams.get(key)) {
                 builder.append("=").append(value);


src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -209,7 +212,9 @@ private String getCacheKey(String url, Multimap<String, String> queryParams) thr
             builder.append("_").append(restoreFactory.getAsUsername());
         }
         builder.append("_").append(uri);
-        for (String key : queryParams.keySet()) {
+        List<String> keyList = queryParams.keySet().stream().collect(Collectors.toList());
+        Collections.sort(keyList);
+        for (String key : keyList) {
             builder.append("_").append(key);
             for (String value : queryParams.get(key)) {
```
snopoke, line 219: The values should also be sorted

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -209,9 +213,14 @@ private String getCacheKey(String url, Multimap<String, String> queryParams) thr
             builder.append("_").append(restoreFactory.getAsUsername());
         }
         builder.append("_").append(uri);
-        for (String key : queryParams.keySet()) {
+        List<String> keyList = queryParams.keySet().stream().collect(Collectors.toList());
+        Collections.sort(keyList);
+        for (String key : keyList) {
```
snopoke, line 218: Looking at this again now I think you can simplify this a bit by using a TreeMap which stores it's data in sorted order:
```suggestion
        // Convert Multimap to TreeMap to sort keys
        Map<String, Collection<String>> sortedQueryParams = new TreeMap<>(queryParams.asMap());
        for (String key : sortedQueryParams.keySet()) {
```

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -209,9 +213,14 @@ private String getCacheKey(String url, Multimap<String, String> queryParams) thr
...
             builder.append("_").append(key);
-            for (String value : queryParams.get(key)) {
+            Collection<String> values = queryParams.get(key);
+            List<String> valuesList = values.stream().collect(Collectors.toList());
+            Collections.sort(valuesList);
```
snopoke, line 222: You can combine the sorting with the previous operation:
```suggestion
            List<String> valuesList = values.stream().sorted().toList();
```

snopoke (commented):
Some small style suggestions. Also it would be great to add a unit test. Here's an example of how you can call this function in a test:

https://github.com/dimagi/formplayer/blob/95cab483628416356b1c16899d5b45c96ec199d6/src/test/java/org/commcare/formplayer/tests/CaseClaimNavigationTests.java#L91-L97

Robert-Costello:
> Also it would be great to add a unit test.
> 
> 

added a test here https://github.com/dimagi/formplayer/pull/1541/commits/e332caa8456cb59ff410c627e819fef5b8fe2e4c
I looked around for where this test should live. maybe `CaseClaimTests` would be better?

snopoke (approved):

ctsims:
@Robert-Costello I think there are a few other important things to add to this test. Specifically I think it's important to add some evidence of negative key matches, as these tests will pass any time the blackbox implementation just returns the first ordered value for each key (or just returns `True` for that matter). 

Since one of the nuances of the multi-map is that it can have multiple values for the same key, for instance, it could be the case that the test will falsely succeed because both key orderings are **excluding** the same value from the key set. I'd add a separate case for the scenario where "a" has two values and the second multimap doesn't include the second value according to the comparator.

ctsims:
@Robert-Costello Whoops, looks like the test changed since I wrote that yesterday and remembered to post it today. I'm a little hesitant / skeptical of the current blackbox testing methodology, though, where the test is comparing to a hand-constructed hash, rather than generating multiple hashes and ensuring they compare properly. It locks the test into the implementation, versus the properties of the function. 

Not strong enough to push back on merge, but I think it'd be more future facing to implement the checks for each aspect of when the key should / shouldn't match based on input condition rather than a hand constructed comparison.

Robert-Costello:
@ctsims I think these changes address your concerns, but let me know if there is anything I missed. https://github.com/dimagi/formplayer/pull/1541/commits/221af4176c230170a3aabb14bb04622ef50fc7c9

Robert-Costello:
pinging for re-approval after changes
@snopoke @ctsims

snopoke (approved):
I'm not exactly sure what @ctsims meant by his testing comment but I don't think it warrants waiting to merge the PR.

From my point of view it would be nice to see a suite of tests that test different aspects of the function like:
* no keys
* key order
* keys with no values
* keys with multiple values in random order

shubham1g5:
> As per discussion about better cache handling, this sorts the query params key set values that are used to create keys for case search caches.

Looking it ahead in time I can't tell exactly why this change was made. Are there any links we can add to point to the discussion you mentioned ? Or maybe add more info on the issue we are trying to solve here.

Robert-Costello:
> Are there any links we can add to point to the discussion you mentioned ?

@shubham1g5  Added a link to the discussion in the description.

PR #1542, opened 2024-02-13
clear search data before parsing

snopoke:
This is the minimal fix and test for the issue described in https://dimagi.atlassian.net/browse/USH-4181

Note that there is scope for future changes to improve this further:

* move the init code to the storage classes and only clear the data if the tables already exist
* review whether skipping the mark rewind for auto-launch case lists is desirable. I believe we could adjust the 'selections' (by adding the action selection) after the first auto-launch so that future replays would be forced to follow the case search path.
* consider per session IDs stored in the stack which could be included in the cache key. The life-cycle of these IDs would need to be carefully considered

## Product Description
Ensure case search storage is cleared before loading new data.

## Technical Summary
Clear the case search tables before parsing new data into them to prevent leaving stale data in the tables.

### Safety story
Testing in unit tests and on staging.

### Automated test coverage
Adding in the PR

### QA Plan
Testing on staging by delivery team.

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
+++ b/src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
@@ -129,10 +129,12 @@ public class CaseSearchHelper {
             XmlPullParserException, IOException {
         try {
             DbUtils.setAutoCommit(caseSearchDb, false);
+            caseSearchIndexTable.delete();
             caseSearchIndexTable.createTable();
             CaseInstanceXmlTransactionParserFactory factory = new CaseInstanceXmlTransactionParserFactory(
                     caseSearchSandbox, caseSearchIndexTable);
             caseSearchStorage.initStorage();
+            caseSearchStorage.removeAll();
             ParseUtils.parseIntoSandbox(responeStream, factory, true, true);
             DbUtils.commit(caseSearchDb);
         } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -129,10 +129,12 @@ private void parseIntoCaseSearchStorage(SQLiteDB caseSearchDb, UserSqlSandbox ca
             XmlPullParserException, IOException {
         try {
             DbUtils.setAutoCommit(caseSearchDb, false);
+            caseSearchIndexTable.delete();
             caseSearchIndexTable.createTable();
             CaseInstanceXmlTransactionParserFactory factory = new CaseInstanceXmlTransactionParserFactory(
                     caseSearchSandbox, caseSearchIndexTable);
             caseSearchStorage.initStorage();
+            caseSearchStorage.removeAll();
```
Robert-Costello, line 137: This implies that there may be data in the table immediately after `initStorage`. Is this because case data may persist in an `ExternalDataInstance`?

Robert-Costello (approved):
Left one question. Assuming testing is successful this looks good to me.

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
snopoke: no, `initStorage` is called optimistically to ensure that the tables are created. Ideally it would return a boolean indicating whether the tables were created on not which would allow us to skip the `removeAll` call for fresh tables.

ctsims:
@snopoke This fix is interesting to me in where it's implemented in the code, can you articulate in the PR description (or point to a ticket) what the actual defect / failure path is?

ctsims (approved):
Definitely looks like a safe change to me for now while we dig in.

Robert-Costello:
going to go ahead and merge to get change on prod

snopoke:
> @snopoke This fix is interesting to me in where it's implemented in the code, can you articulate in the PR description (or point to a ticket) what the actual defect / failure path is?

@ctsims I've updated the description with more details which previously I had in the P2 doc.

PR #1551, opened 2024-03-11
Datadog Metric Slicing

Jtang-1:
## Product Description

For the metric `formplayer.metrics.timings`
-  Datadog tag `module_initiated_by` is renamed to `request_initiated_by`.
- `app_name` tag is added
- `module_name` tag is included when `navigate_menu` enters a menu

TODO: Add screenshot

Screen Shot 2024-03-18 at 4 46 21 PM

## Technical Summary

[Jira Ticket ](https://dimagi.atlassian.net/browse/USH-4182)
[Related HQ Changes](https://github.com/dimagi/commcare-hq/pull/34247)

This PR consists of two changes:
1. The `module_initiated_by` tag was introduced previously in this [PR](https://github.com/dimagi/formplayer/pull/1533). This tag will be renamed to a more general `request_initiated_by`. This name better reflects that the tag reflects the user action that resulted in the request. 
2. The `request_initiated_by` value is moved from `QueryData` to `SessionNavigationBean` as the values are no longer specific to  just `Query`

### Safety story

tested on staging
The blast zone would include endpoints where HQ includes a `requestInitiatedByTag` value. However, the logic added only involves adding a datadog tag which is already extensively used.

### Automated test coverage

tests already exists for `navigate_menu` and `navigate_menu_start` endpoints.

### QA Plan

[QA Ticket](https://dimagi.atlassian.net/browse/QA-6188?focusedCommentId=316724&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-316724)

### Special deploy instructions

- [X] This PR can be deployed after merge with no further considerations.

### Rollback instructions

- [X] This PR can be reverted after deploy with no further considerations.

### Review

- [X] The set of people pinged as reviewers is appropriate for the level of risk of the change.

PR #1553, opened 2024-03-15
Datadog Improve Metric Fidelity for Autoselect

Jtang-1:
## Product Description

For the metric `formplayer.metrics.timings`
- Datadog tag `category` is separated into multiple tags that are in the form `request_includes_<action>` with the value "true" if applicable. For example, `category` with the value `app_install` will now be the tag `request_includes_app_install` with the value `true`.
- Adds `request_includes_autoselect` tag

Screen Shot 2024-03-18 at 4 56 00 PM

For the metric `formplayer.metrics.granular.timings`
- for `category` `complete_restore`: adds associated tag for `module_name` and `request_includes_autoselect`
- adds `category` `form_entry` with associated tag for `form_name` and `request_includes_autoselect`

Screen Shot 2024-03-18 at 4 56 53 PM

Screen Shot 2024-03-18 at 4 57 32 PM

## Technical Summary

[Jira Ticket](https://dimagi.atlassian.net/browse/USH-4243)

The `category` tag was introduced previously to metric `formplayer.metrics.timings` from this [PR](https://github.com/dimagi/formplayer/pull/1538). However, I now want to also include "autoselect" as a value. However, `category` with the value `autoselect` doesn't make sense in the context of metric `formplayer.metrics.granular.timings` because category is conveys that the timing measures the specific category. For this reason, I am changing the tag key to `request_includes_<action>`. 

Separate tag keys are created for each action instead of having one tag key with multiple values. This is because with the implementation of [addRequestScopedTag](https://github.com/dimagi/formplayer/blob/1bdce0757d05e9a1325f12b7d24055184deabc21/src/main/java/org/commcare/formplayer/util/FormplayerDatadog.java#L90), only the last value saved to the tag would be recorded. However, multiple actions of interest could occur in the same request. See this [PR](https://github.com/dimagi/formplayer/pull/1551) for a similar change.

### Safety story

tested on staging.  The logic added only involves adding datadog tags which is already extensively used.

### Automated test coverage

No explicit testing for datadog tags, but testing exists for menu navigation.

### QA Plan

no QA

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -24,6 +24,7 @@ import org.commcare.formplayer.exceptions.ApplicationConfigException;
 import org.commcare.formplayer.exceptions.SyncRestoreException;
 import org.commcare.formplayer.objects.FormVolatilityRecord;
 import org.commcare.formplayer.objects.QueryData;
+import org.commcare.formplayer.services.CategoryTimingHelper;
 import org.commcare.formplayer.screens.FormplayerQueryScreen;
 import org.commcare.formplayer.screens.FormplayerSyncScreen;
 import org.commcare.formplayer.session.FormSession;
@@ -31,6 +32,7 @@ import org.commcare.formplayer.session.MenuSession;
 import org.commcare.formplayer.util.Constants;
 import org.commcare.formplayer.util.FormplayerDatadog;
 import org.commcare.formplayer.util.FormplayerHereFunctionHandler;
+import org.commcare.formplayer.util.SimpleTimer;
 import org.commcare.formplayer.web.client.WebClient;
 import org.commcare.modern.session.SessionWrapper;
 import org.commcare.modern.util.Pair;
@@ -97,6 +99,9 @@ public class MenuSessionRunnerService {
 
     @Autowired
     private CaseSearchHelper caseSearchHelper;
+
+    @Autowired
+    private CategoryTimingHelper categoryTimingHelper;
 
     @Autowired
     private WebClient webClient;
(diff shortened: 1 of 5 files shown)


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -475,8 +475,10 @@ private void doPostAndSync(MenuSession menuSession, FormplayerSyncScreen screen)
         if (shouldSync) {
             String moduleName = ScreenUtils.getBestTitle(menuSession.getSessionWrapper());
             Map<String, String> extraTags = new HashMap<>();
-            extraTags.put(Constants.MODULE_NAME_TAG, moduleName);
+            Map<String, String> requestScopedTagNameAndValueMap = datadog.getRequestScopedTagNameAndValue();
 
+            requestScopedTagNameAndValueMap.computeIfPresent(Constants.REQUEST_INCLUDES_TAG, extraTags::put);
```
shubham1g5: we can probably make this logic to consolidate tags part of `recordCategoryTiming` method itself ?

shubham1g5 (approved):
Looks good, a minor suggestion to ponder.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
Jtang-1: I think to reduce the number of tag combinations (thinking about cost) we wouldn't always want to include the requestScopedTags into the category timing metric. So in that case, we would decide which of the requestScopedTags we're interested in on the higher level and pass it into `recordCategoryTiming`. But let me know if i'm misunderstanding.

Jtang-1: mmm that could work. But the `requestScopedTag` would contain more than just the `REQUEST_INCLUDES_TAG` so are you imagining a flag for each tag? Like a parameter to `recordCategoryTiming ` for `add_request_includes_tag`, `add_module_name_tag`?

shubham1g5: no I was just imagining a single flag like ` includeRequestScopedTags` for `recordCategoryTiming`  that adds the request scoped tags to `extraTags` .. I think the main thing I am trying to avoid is duplication of the tags merging logic. there are 2 places in the PR where we are doing this. Another easier method might just be to have a method like `datadog.consolidateTags(extraTags)` that returns you the complete set of tags including the scoped tags.

Jtang-1: Ah that would relate back to the initial concern with tag combinations. `requestScopedTagNameAndValueMap` contains more than just the `REQUEST_INCLUDES_TAG` tag. And for costs, we won't want to include all the tags.

I am merging this PR for to make into the deploy, but happy to continue discussing this.

PR #1560, opened 2024-03-22
End of Form Error Handling

Jtang-1:
## Product Description

This PR does two things: 
- Displays errors while rebuilding session from frame that were previously caught but not displayed. Namely, errors resulting from end of form navigation but that is not the only feature where rebuilding session is done.
- Adds error handling for when no matches are found between steps and menu options.
Screen Shot 2024-03-21 at 10 26 19 PM

## Technical Summary

[Jira Ticket](https://dimagi.atlassian.net/browse/USH-4176)

### Safety story

tested locally, will test on staging

### Automated test coverage

### QA Plan

no QA

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -125,6 +125,9 @@ public class MenuSessionFactory {
                             currentStep = step.getValue();
                             processedSteps.add(step);
                             needsFullInit = ++processedStepsCount == steps.size();
+                        } else {
+                            throw new CommCareSessionException("Value " + step.getValue() +" is not found in " +
+                                entityScreen.toString());
                         }
                         break;
                     }
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -151,7 +151,6 @@ public class MenuSessionFactory {
                         try {
                             uri = new URI(step.getValue());
                             processedSteps.add(step);
-                            processedStepsCount++;
                         } catch (URISyntaxException e) {
                             e.printStackTrace();
                             throw new CommCareSessionException("Query URL format error: " + e.getMessage(), e);
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -34,6 +34,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collection;
+import java.util.StringJoiner;
 import java.util.Vector;
 
 import datadog.trace.api.Trace;
@@ -105,6 +106,17 @@ public class MenuSessionFactory {
                             needsFullInit = ++processedStepsCount == steps.size();
                         }
                     }
+                }
+                if (currentStep == null) {
+                    StringJoiner optionsIDJoiner = new StringJoiner(", ", "[", "]");
+                    StringJoiner stepIDJoiner = new StringJoiner(", ", "[", "]");
+                    for (MenuDisplayable option : options) {
+                        optionsIDJoiner.add(option.getCommandID());
+                    }
+                    for (StackFrameStep step : steps) {
+                        stepIDJoiner.add(step.getId());
+                    }
+                    throw new CommCareSessionException("Match Error: Steps " + stepIDJoiner.toString() + " do not contain a valid option " + optionsIDJoiner.toString());
                 }
             } else if (screen instanceof EntityScreen) {
                 EntityScreen entityScreen = (EntityScreen)screen;
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -121,7 +121,8 @@ public class MenuSessionFactory {
                 }
                 if (currentStep == null && processedStepsCount != steps.size()) {
                     for (StackFrameStep unprocessedStep : unprocessedSteps) {
-                        if (unprocessedStep.getType().equals(SessionFrame.STATE_COMMAND_ID)) {
+                        if (unprocessedStep.getType().equals(SessionFrame.STATE_COMMAND_ID) &&
+                            !unprocessedStep.getId().startsWith("claim_command")) {
                             StringJoiner optionsIDJoiner = new StringJoiner(", ", "[", "]");
                             StringJoiner stepIDJoiner = new StringJoiner(", ", "[", "]");
                             for (MenuDisplayable option : options) {
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -128,16 +128,7 @@ public class MenuSessionFactory {
                             processedSteps.add(step);
                             needsFullInit = ++processedStepsCount == steps.size();
                         } else {
-                            // This block constructs the message to display then throws the exception
-                            List<String> refsList = entityScreen.getReferences().stream()
-                            .map(ref -> EntityScreen.getReturnValueFromSelection(ref, (EntityDatum) neededDatum, entityScreen.getEvalContext()))
-                            .toList();
-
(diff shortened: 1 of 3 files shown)


snopoke (approved):

Jtang-1:
@snopoke There's an edge case causing test failures and I can't figure out how to best handle it. Could I get your thoughts? From what I understand, the situation is:
Steps `[m1, case_id, case_id_new_child_case_0]`. We just processed `case_id` (via EntityScreen) and move to the next screen which is a MenuScreen['Case Claim' with choices [Entry with id m1-f0, Menu with id m2]].

`case_id_new_child_case_0` is a datum for a new case id (`<datum id="case_id_new_child_case_0" function="uuid()"/>`). The step is `(CASE_ID case_id_new_child_case_0 : 81e0aa64-7aea-4134-b000-43f04fd80f2d)` (where the case_id is a generated uuid). The logic before this PR is that this step is ignored (if this [loop](https://github.com/dimagi/formplayer/blob/767f3becad2563e9ab181fe06389fd819365f223/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L101-L109) has no matches, `currentStep` stays `null` and [we break out of the `while` loop.](https://github.com/dimagi/formplayer/blob/767f3becad2563e9ab181fe06389fd819365f223/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L190-L191))

But with this PR, the non-match throws an exception. A solution for this is: if we encounter a match "miss", we check if the remaining steps are datums for new case ids. If true, we don't throw an exception. However, I don't see a way to know if the step is associated with creating a new case_id. 

But similarly, if the last step is `(SELECTED_ENTITIES selected_cases : ...)`, it causes the same issue. So maybe an additional condition before throwing the error should be that the step we're processing is a `COMMAND_ID`.

snopoke:
I did some debugging and it took a while to figure out what was going on but I think your suggestion to check that the current step we're processing is a `COMMAND_ID` makes sense.

This might be a challenge to implement without completely refactoring that method since we don't actually track which step is the current one. Personally I've wanted to refactor that function many times but never managed to push it all the way through.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -106,6 +107,17 @@ public void rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper ca
                         }
                     }
                 }
+                if (currentStep == null && processedStepsCount != steps.size()) {
+                    StringJoiner optionsIDJoiner = new StringJoiner(", ", "[", "]");
```
MartinRiese, line 111: I cannot believe `StringJoiner` exists since 1.8 and I never heard of it. Pretty handy.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -106,6 +107,17 @@ public void rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper ca
                         }
                     }
                 }
+                if (currentStep == null && processedStepsCount != steps.size()) {
```
MartinRiese, line 110: Could you push that in its own method? `rebuildSessionFromFrame` is quite long already

Jtang-1: [bf6e37e](https://github.com/dimagi/formplayer/pull/1560/commits/bf6e37e2230b0c66ff3b542a85d89a6219b4242b)

Jtang-1:
Thank you for taking the time to look into this for me @snopoke. I ended up tracking what steps were processed, and checking that the remaining unprocessed steps contained a `COMMAND_ID` type (minus the special value `claim_command...`) that was not matched to an available option. This seems to be more conservative to reduce chances of falsely throwing the exception.
From testing, I found tracking the step we're trying to process is tricky (as you mentioned). It seems like there's steps that we just skip and don't handle within `rebuildSessionFromFrame`, such as the `claim_command...` and the "new case id` datums. Unfortunately, I'll also be passing this along and not refactor it at this moment.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -101,11 +106,15 @@ public void rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper ca
                     for (StackFrameStep step : steps) {
                         if (step.getId().equals(options[i].getCommandID())) {
                             currentStep = String.valueOf(i);
+                            processedSteps.add(step);
```
snopoke, line 109: I notice this loop has no break statement in it. I think it should since I can't think of a scenario where more than one step would match.

Jtang-1: I was thinking that too. I will make that change in this separate [PR](https://github.com/dimagi/formplayer/pull/1562)

Aside note: I was speaking to Shubham about commit [4ef0364](https://github.com/dimagi/formplayer/pull/1560/commits/4ef0364b1a911e25fd90f45a6a4615e3d15f7d96) and he suggested doing this in a separate PR with more scrutiny (doing QA). So may as well have this extra check done on the `break` statement too.

Jtang-1:
Bumping for review

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
snopoke: In that case, should this line be moved out of the for loop since only the last matching step would be 'processed'. I don't think it should be possible for multiple steps to match but I can't prove that.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -114,7 +125,19 @@ public void rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper ca
                     if (step.getId().equals(neededDatum.getDataId())) {
                         if (entityScreen.referencesContainStep(step.getValue())) {
                             currentStep = step.getValue();
+                            processedSteps.add(step);
                             needsFullInit = ++processedStepsCount == steps.size();
+                        } else {
+                            // This block constructs the message to display then throws the exception
+                            List<String> refsList = entityScreen.getReferences().stream()
+                            .map(ref -> EntityScreen.getReturnValueFromSelection(ref, (EntityDatum) neededDatum, entityScreen.getEvalContext()))
+                            .collect(Collectors.toList());
```
snopoke, line 134: instead of `.collect(Collectors.toList())` you should be able to just call `.toList()`

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -114,7 +125,19 @@ public void rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper ca
                     if (step.getId().equals(neededDatum.getDataId())) {
                         if (entityScreen.referencesContainStep(step.getValue())) {
                             currentStep = step.getValue();
+                            processedSteps.add(step);
                             needsFullInit = ++processedStepsCount == steps.size();
+                        } else {
+                            // This block constructs the message to display then throws the exception
```
snopoke, line 131: similar to the comment above, it would be good to extract this to a method

snopoke (approved):
Nothing blocking.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
Jtang-1: [6754f61](https://github.com/dimagi/formplayer/pull/1560/commits/6754f61adb48a77d3581fcce4be805c8385918e6)

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
Jtang-1: [8b47055](https://github.com/dimagi/formplayer/pull/1560/commits/8b47055ee35d51dd9e5d5aad89eb30a9c9c60a81)

Jtang-1:
@snopoke, I couldn't reply to that comment thread
> In that case, should this line be moved out of the for loop since only the last matching step would be 'processed'. I don't think it should be possible for multiple steps to match but I can't prove that.

I'm not entirely sure if it's possible either but I see `needsFullInit = ++processedStepsCount == steps.size();` is within the for loop so I kept it within the same level to be on the safe side.

snopoke (approved):

PR #1576, opened 2024-04-26
End of Form Error Handling

Jtang-1:
## Product Description

Context:
This [PR](https://github.com/dimagi/formplayer/pull/1560) introduces validation during session rebuilding. If there's a discrepancy during this process, it will now raise an exception and show an error to the user. Before this PR, the session would simply revert to the last valid step without raising any exceptions. In terms of End of Form (EoF), this means the user would be directed to the last valid screen without any error prompts.

Issue:
However, some apps have been configured to rely on this EoF behavior, where the user is directed to the last valid screen, even if it's not explicitly defined in the app configuration. Take for example this app set up/workflow:

`Home → Application → My Cases Case List→ <casefirst_name case_last_name> → My Cases Case List (Form selection) -> Case Investigation Form`

- "My Cases Case List" is default filtered to only show open cases
- "Case Investigation Form" has EoF set to "Previous Screen"
If "Case Investigation Form" closes the case, then the path from `My Cases Case List→ <casefirst_name case_last_name> → My Cases Case List (Form selection)` is broken and would throw an exception. However, the prior behavior would navigate the user to `My Cases Case List` since it can't reach "My Cases Case List (Form Selection).

Additional note: This was configured to go to the "Previous Screen" instead of specifying the specific case list because there are various shadow modules from which the Case Investigation Form is accessible and we want to take the user back to whatever case list they came from.

Solution from this PR:
For now, instead of throwing the error, we will just log it so we can continue to use this information for debugging any real EoF bugs.

## Technical Summary

[Jira Ticket](https://dimagi.atlassian.net/browse/USH-4505?atlOrigin=eyJpIjoiYTU2YjE3YzgyMWUwNGZmYTkyOTJmYTA4NDE1ZTY4M2QiLCJwIjoiaiJ9)
[Sentry Error
](https://dimagi.sentry.io/issues/5176121153/events/c83aac6cc1e74083b628ce8a8da05b2d/?project=136860)

### Safety story

locally tested. This change only changes raising an exception to logging.

### Automated test coverage

### QA Plan

no QA

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -128,7 +128,7 @@ public class MenuSessionFactory {
                             processedSteps.add(step);
                             needsFullInit = ++processedStepsCount == steps.size();
                         } else {
-                            LogStepNotInEntityScreenError(entityScreen, neededDatum, step);
+                            logStepNotInEntityScreenError(entityScreen, neededDatum, step);
                         }
                         break;
                     }
@@ -296,7 +296,7 @@ public class MenuSessionFactory {
         }
     }
 
-    private void LogStepNotInEntityScreenError(EntityScreen entityScreen, SessionDatum neededDatum, StackFrameStep step) throws CommCareSessionException {
+    private void logStepNotInEntityScreenError(EntityScreen entityScreen, SessionDatum neededDatum, StackFrameStep step) throws CommCareSessionException {
         // This block constructs the message to display then throws the exception
         List<String> refsList = entityScreen.getReferences().stream()
         .map(ref -> EntityScreen.getReturnValueFromSelection(ref, (EntityDatum) neededDatum, entityScreen.getEvalContext()))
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -288,15 +288,15 @@ private void checkAndThrowCaseIDMatchError(Vector<StackFrameStep> steps, List<St
...
+                log.error(
                     "Match Error: Steps " + stepIDJoiner.toString() +
                     " do not contain a valid datum ID " + neededDatumID
                 );
             }
         }
     }
 
-    private void throwStepNotInEntityScreenException(EntityScreen entityScreen, SessionDatum neededDatum, StackFrameStep step) throws CommCareSessionException {
+    private void LogStepNotInEntityScreenError(EntityScreen entityScreen, SessionDatum neededDatum, StackFrameStep step) throws CommCareSessionException {
```
MartinRiese, line 299: methods start with a lower case

MartinRiese (changes requested):
looks good otherwise

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
Jtang-1: [44cf45e](https://github.com/dimagi/formplayer/pull/1576/commits/44cf45e48a979244b92a3621ac6e1c877a10ec24)

MartinRiese (approved):

PR #1578, opened 2024-04-29
Log instead of blocking app for EoF errors

Jtang-1:
## Product Description

Context:
This https://github.com/dimagi/formplayer/pull/1560 introduces validation during session rebuilding. If there's a discrepancy during this process, it will now raise an exception and show an error to the user. Before this PR, the session would simply revert to the last valid step without raising any exceptions. In terms of End of Form (EoF), this means the user would be directed to the last valid screen without any error prompts.

Issue:
There were previously working EoF navigation configurations that are now displaying an error and blocking the app workflow.

It is currently unknown what specifically about those app configurations are conflicting with this error check. However, due to the urgency, this change logs the error instead of blocking the app. The logging will aid in investigation.

## Technical Summary

[Jira Ticket ](https://dimagi.atlassian.net/browse/QA-6438)

### Safety story

locally tested. This change only changes raising an exception to logging. This change removes blocking errors that were recently introduced.

### Automated test coverage

### QA Plan

no QA

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -115,7 +115,7 @@ public class MenuSessionFactory {
                     }
                 }
                 if (currentStep == null && processedStepsCount != steps.size()) {
-                    checkAndThrowCommandIDMatchError(steps, processedSteps, options);
+                    checkAndLogCommandIDMatchError(steps, processedSteps, options);
                 }
             } else if (screen instanceof EntityScreen) {
                 EntityScreen entityScreen = (EntityScreen)screen;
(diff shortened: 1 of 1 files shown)


MartinRiese (approved):

shubham1g5 (approved):

shubham1g5:
Think as a post-step on these issues, it's important to figure out if we are going to support these workflows as it is and if yes, we should add unit tests on these workflows so that we don't make the same mistake again.

Jtang-1:
Following up - created this ticket to have that discussion. https://dimagi.atlassian.net/browse/USH-4511. 
The ticket is more so related to another [PR](https://github.com/dimagi/formplayer/pull/1576). This PR did not end up being an "unsupported workflow" issue. The app just didn't seem to be set up to work correctly to begin with.

PR #1579, opened 2024-05-02
USH-4497: Do not append domain to username with @

MartinRiese:
## Technical Summary

https://dimagi.atlassian.net/browse/USH-4497

Users now can login as web users which already have the complete username used for couch. Do not append domain and host it that case. Without this change the claim-case request has a non-existing username in the body and will fail silently.

We do the same in hq: https://github.com/dimagi/commcare-hq/pull/34351/files#diff-b5520648dffd5a4f57530e5e1c36f90cd022db2cbf59bfef9bb3d65997d6f316R64-R74

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/util/StringUtils.java
+++ b/src/main/java/org/commcare/formplayer/util/StringUtils.java
@@ -5,6 +5,9 @@ package org.commcare.formplayer.util;
  */
 public class StringUtils {
     public static String getFullUsername(String user, String domain, String host) {
+        if (user.contains("@")) {
+            return user;
+        }
         return user + "@" + domain + "." + host;
     }
 


MartinRiese:
@shubham1g5 the function I changed is called twice. Once in the menu session, which is where the username for the claim case request comes from. And the second time in the [RestoreFactory](https://github.com/dimagi/formplayer/blob/4e9b0d5089c6ad73cdac3d8dc2d837184e305c05/src/main/java/org/commcare/formplayer/services/RestoreFactory.java#L368). I could not figure out when that code gets called and if the change would be appropriate there or if I should move the check.

src/main/java/org/commcare/formplayer/util/StringUtils.java
```diff
@@ -9,6 +9,9 @@ public static String getFullUsername(String user, String domain, String host) {
     }
 
     public static String getFullUsername(String user, String domain) {
+        if (user.contains("@")) {
```
shubham1g5, line 12: since this method is only overloading the method before this, this change should actually be made in the previous method.

shubham1g5:
> And the second time in the [RestoreFactory](https://github.com/dimagi/formplayer/blob/4e9b0d5089c6ad73cdac3d8dc2d837184e305c05/src/main/java/org/commcare/formplayer/services/RestoreFactory.java#L368).

That usage looks harmless as the code that calls it is only getting used for logging without any affect on business logic.

shubham1g5 (changes requested):

shubham1g5:
Q: Can't a Mobile worker name be like - `user@1` on HQ ?

src/main/java/org/commcare/formplayer/util/StringUtils.java
MartinRiese: You mean in MenuSession.getNextScreen correct?

MartinRiese:
A: No they cannot. If you try using an @ during creation you get a validation error.

src/main/java/org/commcare/formplayer/util/StringUtils.java
shubham1g5: no there is another `getFullUsername` method that this method calls.

PR #1590, opened 2024-06-04
Avoid Re-calculating next screen again if we can

shubham1g5:
## Technical Summary

Minor performance optimisation in session eval code. Earlier we were calculating the final screen in session navigation twice, once while processing the selection as part of [session advance](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L369) and then again while calling [getNextMenu](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L153). 

This PR makes a simple change to avoid this re-calculation if we can utilise the existing screen again. 

### Safety story

- Good test coverage arounf this code
- Did basic testing on staging, will let it go through a few round of QA automated tests as well.

### QA Plan

No

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/MenuController.java
+++ b/src/main/java/org/commcare/formplayer/application/MenuController.java
@@ -168,8 +168,10 @@ public class MenuController extends AbstractBaseController {
             @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken,
             HttpServletRequest request) throws Exception {
         String[] selections = sessionNavigationBean.getSelections();
-        MenuSession menuSession;
-        menuSession = menuSessionFactory.getMenuSessionFromBean(sessionNavigationBean);
+        if (selections == null) {
+            selections = new String[0];
+        }
+        MenuSession menuSession = menuSessionFactory.getMenuSessionFromBean(sessionNavigationBean);
         EntityScreenContext entityScreenContext = new EntityScreenContext(sessionNavigationBean.getOffset(),
                 sessionNavigationBean.getSearchText(),
                 sessionNavigationBean.getSortIndex(),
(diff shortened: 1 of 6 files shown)


src/main/java/org/commcare/formplayer/application/MenuController.java
```diff
@@ -168,8 +168,10 @@ public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBe
             @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken,
             HttpServletRequest request) throws Exception {
         String[] selections = sessionNavigationBean.getSelections();
-        MenuSession menuSession;
-        menuSession = menuSessionFactory.getMenuSessionFromBean(sessionNavigationBean);
+        if (selections == null) {
+            selections = new String[0];
```
Robert-Costello, line 172: This makes sense to me, but could this cause any problems in the case of `nonAppNav` in `advanceSessionWithSelections`? I'm not super familiar with when/how `nonAppNav` would be true, but if `nonAppNav` was true the length of `selections` was `0`,  `selections = Arrays.copyOf(selections, selections.length - 1)` might throw an error.

shubham1g5: That's a good point, I think `null` selections are only sent as part of `navigate_menu_start` but nevertheless added [a empty selections check](https://github.com/dimagi/formplayer/pull/1590/commits/5136e3e0f1000a26e4b47256b31aaf0da3f7b2da) for safety here.

Robert-Costello (approved):

PR #1596, opened 2024-06-28
make widow width available in session

Robert-Costello:
## Technical Summary

[jira ticket](https://dimagi.atlassian.net/browse/USH-4671)
[Related CC-Core PR](https://github.com/dimagi/commcare-core/pull/1414)
[Related Web Apps PR  ](https://github.com/dimagi/commcare-hq/pull/34832)

### QA Plan

[QA Ticket](https://dimagi.atlassian.net/browse/QA-6751)

### Rollback instructions

### Review

### Duplicate PR
Automatically duplicate this PR as defined in [contributing.md](https://github.com/dimagi/commcare-core/blob/master/.github/contributing.md).


src/main/java/org/commcare/formplayer/beans/AuthenticatedRequestBean.java
```diff
@@ -100,4 +101,14 @@ public String getRestoreAsCaseId() {
     public void setRestoreAsCaseId(String restoreAsCaseId) {
         this.restoreAsCaseId = restoreAsCaseId;
     }
+
+    @JsonGetter(value = "windowWidth")
+    public String getWindowWidth() {
```
MartinRiese, line 106: Not having much expertise here. But `AuthenticatedRequestBean` to me seems like the wrong place to put this. What is the reasoning for this location?

MartinRiese (commented):
The rest looks like straight forward passing through of the parameter.

src/main/java/org/commcare/formplayer/beans/AuthenticatedRequestBean.java
Robert-Costello: Also not an expert here. Originally this was in SessionNavigationBean, but I shifted it up the inheritance tree so it would be available [here](https://github.com/dimagi/formplayer/blob/26fcd44ff99a07225b0e5153b54295e2668b4c3b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L241) where it creates a new `MenuSession`.

shubham1g5: Probably better to add it in `InstallRequestBean` in that case.

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -361,7 +364,6 @@ private long getVolatilityKeyTimeout() {
                 int timeOutInput = Integer.parseInt(timeOut);
                 timeOutWindow = timeOutInput;
             } catch (NumberFormatException nfe) {
-                System.out.println("Invalid timeout window: " + timeOut);
```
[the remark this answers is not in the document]
Robert-Costello: nope, that was entirely accidental. I must have removed this when cleaning up my changes. I'll put it back.

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -146,7 +147,8 @@ public FormSession(UserSqlSandbox sandbox,
             boolean inPromptMode,
             String caseId,
             @Nullable SessionFrame sessionFrame,
-            RemoteInstanceFetcher instanceFetcher) throws Exception {
+            RemoteInstanceFetcher instanceFetcher,
+            @Nullable String windowWidth) throws Exception {
```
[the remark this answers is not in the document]
Robert-Costello: I actually don't anticipate there ever being a `null` value with the way this is now implemented. It wasn't clear during the speccing phase, but I think this can be removed.

src/main/java/org/commcare/formplayer/beans/AuthenticatedRequestBean.java
shubham1g5: makes sense, fine to be in `AuthenticatedRequestBean` in that case.

Robert-Costello:
Hi @shubham1g5, I added a test that follows the template of testFormEntryWithQuery [here](https://github.com/dimagi/formplayer/pull/1596/commits/1d76f9bad11923367afe4114cb9a94f1b58cdc5e). I've tried a bunch of other things with the various session navigate methods in BaseTestClass but keep seeing the same result, where the windowWidth is being set when I step through, but when I grab the formSession to check, the windowWidth is null.

shubham1g5:
> I added a test that follows the template of testFormEntryWithQuery [here](https://github.com/dimagi/formplayer/pull/1596/commits/1d76f9bad11923367afe4114cb9a94f1b58cdc5e). I've tried a bunch of other things with the various session navigate methods in BaseTestClass but keep seeing the same result, where the windowWidth is being set when I step through, but when I grab the formSession to check, the windowWidth is null.

@Robert-Costello Seems like we are not passing the `windowWidth` to all constructors of `FormSession` ([class](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/application/FormSessionFactory.java)) which is the pathway being utilised in your test.

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -364,6 +364,7 @@ private long getVolatilityKeyTimeout() {
                 int timeOutInput = Integer.parseInt(timeOut);
                 timeOutWindow = timeOutInput;
             } catch (NumberFormatException nfe) {
+                System.out.println("Invalid timeout window: " + timeOut);
```
MartinRiese, line 367: is this the correct way to log the error?

MartinRiese (approved):

minhaminha (approved):

minhaminha:
meant to add a comment to the approval: looks good except the failing tests!

Robert-Costello:
> meant to add a comment to the approval: looks good except the failing tests! edit: actually the FormplayerInstanceInitializer not having access to `getWindowWidth` is definitely a problem

`FormplayerInstanceInitializer` will have access to `getWindowWidth` once [this cc-core PR](https://github.com/dimagi/commcare-core/pull/1414) is merged. [This comment](https://github.com/dimagi/commcare-core/pull/1414#issuecomment-2269405228) adds more context.

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
```diff
@@ -156,7 +157,8 @@ public FormSession getFormSession(SerializableFormSession serializableFormSessio
                 storageFactory,
                 commCareSession,
                 formplayerRemoteInstanceFetcher,
-                formDefinitionService
+                formDefinitionService,
+                null
```
shubham1g5, line 161: we need to pass `window_width` from beans to all constructors otherwise specific request will not work. For ex. this will cause all `/incomplete-form` requests to have the param as `null`

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -89,14 +89,16 @@ public class FormSession {
...
 
     public FormSession(SerializableFormSession session,
             RestoreFactory restoreFactory,
             FormSendCalloutHandler formSendCalloutHandler,
             FormplayerStorageFactory storageFactory,
             @Nullable CommCareSession commCareSession,
             RemoteInstanceFetcher instanceFetcher,
-            FormDefinitionService formDefinitionService) throws Exception {
+            FormDefinitionService formDefinitionService,
+            String windowWidth) throws Exception {
```
shubham1g5, line 101: Instead of passing `windowWidth` directly can we pass a Object say `MetaSessionContext` that can host all different session variables, otherwise adding each parameter to session meta data will cause addition of new variable in method signatures for all new params.

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -1035,7 +1041,8 @@ protected FormSession getFormSession(SerializableFormSession serializableFormSes
                 storageFactoryMock,
                 getCommCareSession(serializableFormSession.getMenuSessionId()),
                 remoteInstanceFetcher,
-                formDefinitionService
+                formDefinitionService,
+                "1796"
```
shubham1g5, line 1045: It's better to have the individual test pass it instead of harcoding it in a parent test class.

src/test/java/org/commcare/formplayer/tests/ScreenSizeTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.cache.CacheManager;
+
+import java.util.Hashtable;
+
+@WebMvcTest
+public class ScreenSizeTest extends BaseTestClass {
```
shubham1g5, line 27: Don't think we need this test anymore as it test the same thing as `testWindowWidthParam` with unnecessary setup.

shubham1g5 (changes requested):

src/main/java/org/commcare/formplayer/session/FormSession.java
Robert-Costello: I'm getting all of the constructors accepting `windowWidth` from beans before I shift every thing to this new format.

shubham1g5 (approved):
Looks good to be merged, think we can have the refactor to include `MetaSession` object as a followup and it doesn't need to block merge on this if needed.

MartinRiese (approved):

PR #1601, opened 2024-07-15
bug: populates instance so that it is available when getting the next…

Jtang-1:
… screen

### Migrations

### Rollback instructions

### Review


Jtang-1:
Resolved with this PR https://github.com/dimagi/formplayer/pull/1608

PR #1608, opened 2024-08-06
Smartlink bug with Display Condition

Jtang-1:
## Product Description

App Config:
Module that uses inline case search and a smartlink. Inside that module, a form contains a "Display condition" that refers to `instance('results:inline')...`. 

Bug:
Using the smartlink to the module will result in an error. `Instance referenced by instance(results:inline)... has not been loaded`

## Technical Summary

Jira Ticket
https://dimagi.atlassian.net/browse/USH-4714

Syncing, if required, needs to happen between stack operations. However, the `results:inline` instance is not loaded until after `rebuildSessionFromFrame`. If a form contains a display condition that references the `results:inline` instance, then `menuScreen` can not be initialized and will error. We only need the `syncScreen` while executing operations so I pulled out the logic that will check for and initialize only a `syncScreen`. 

### Safety story

Tested locally. Also  The impact is limited to smartlinking.

### Automated test coverage

Test exists for core smartlinking feature. I also added a test that reproduces the bug that this PR fixes. 

### QA Plan

no QA

### Rollback instructions

### Review


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -786,7 +786,7 @@ public BaseResponseBean advanceSessionWithEndpoint(MenuSession menuSession, Stri
         // Sync requests aren't run when executing operations, so stop and check for them after each operation
         for (StackOperation op : endpoint.getStackOperations()) {
             sessionWrapper.executeStackOperations(new Vector<>(Arrays.asList(op)), evalContext);
-            Screen screen = menuSession.getNextScreen(false, new EntityScreenContext());
+            Screen screen = menuSession.getNextScreenIfSyncScreen(false, new EntityScreenContext());
             if (screen instanceof FormplayerSyncScreen) {
```
snopoke, line 790: seems like this if statement isn't necessary now

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -258,15 +258,35 @@ public Screen getNextScreen(boolean needsFullEntityScreen, EntityScreenContext e
...
-                    StringUtils.getFullUsername(session.getAsUser(), session.getDomain()) : null;
-            FormplayerSyncScreen syncScreen = new FormplayerSyncScreen(username);
-            syncScreen.init(sessionWrapper);
-            return syncScreen;
+            return getSyncScreen();
         }
         throw new RuntimeException("Unexpected Frame Request: " + sessionWrapper.getNeededData());
     }
 
+    public Screen getNextScreenIfSyncScreen(boolean needsFullEntityScreen, EntityScreenContext entityScreenContext) throws CommCareSessionException {
```
snopoke, line 266: This could do with a doc string giving more context. I don't quite follow what it's trying to do. It seems like it's looking for a sync screen anywhere in the stack up to that point. Is that correct?

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -258,15 +258,35 @@ public Screen getNextScreen(boolean needsFullEntityScreen, EntityScreenContext e
...
+        String next = sessionWrapper.getNeededData(sessionWrapper.getEvaluationContext());
+        if (next == null) {
+            if (sessionWrapper.isViewCommand(sessionWrapper.getCommand())) {
+                sessionWrapper.stepBack();
```
snopoke, line 270: Can this ever cause an infinite loop? "step back", next is null, "step back"

snopoke (commented):
Needs some tests

src/main/java/org/commcare/formplayer/session/MenuSession.java
Jtang-1: I don't think so, this logic is the same as what's done for `getNextScreen`. And from what i'm understanding, `stepBack` pops steps until we reach a step that requires user input, at which point `isViewCommand` will be false. 

But actually, I think this can be removed since `syncScreen` does not requires user input so this will not result in a syncScreen.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
Jtang-1: an if statement is still needed, but it can be updated to `if (screen != null)`

src/main/java/org/commcare/formplayer/session/MenuSession.java
Jtang-1: It's what `getNextScreen` does but only initializes and returns the screen if it's a `syncScreen`. 

What do you think of this docstring?
```
Get next screen for the current request, based on the current state of the session, 
but only initialize and return the screen if it is of type `FormplayerSyncScreen`.     ```

Comment (Jtang-1) [reply src/main/java/org/commcare/formplayer/session/MenuSession.java]: [9580cb2](https://github.com/dimagi/formplayer/pull/1608/commits/9580cb245cc2008d1284d99b0ac32386263d3a83)

File: src/main/java/org/commcare/formplayer/session/MenuSession.java (hunk 3 above)
Comment (Jtang-1) [reply src/main/java/org/commcare/formplayer/session/MenuSession.java]: [f6c4d59](https://github.com/dimagi/formplayer/pull/1608/commits/f6c4d59baa7ceb5f19a6088c28e6abb1f3c8239e)

File: src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java (hunk 1 above)
Comment (Jtang-1) [reply src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java]: [6d17f14](https://github.com/dimagi/formplayer/pull/1608/commits/6d17f14eac50f83d418be65ec0e56222744e01de)

File: src/main/java/org/commcare/formplayer/session/MenuSession.java [hunk 4]
```diff
@@ -258,15 +258,34 @@ public Screen getNextScreen(boolean needsFullEntityScreen, EntityScreenContext e
...
+            return getNextScreenIfSyncScreen(needsFullEntityScreen, entityScreenContext);
+        } else if (next.equalsIgnoreCase(SessionFrame.STATE_SYNC_REQUEST)) {
+            return getSyncScreen();
+        }
+        return null;
+    }
+
+    private FormplayerSyncScreen getSyncScreen() throws CommCareSessionException{
+        String username = session.getAsUser() != null ?
+        StringUtils.getFullUsername(session.getAsUser(), session.getDomain()) : null;
```
shubham1g5, line 283: nit: indentation seems off

shubham1g5 (approved):

src/main/java/org/commcare/formplayer/session/MenuSession.java
```diff
@@ -258,15 +258,34 @@ public Screen getNextScreen(boolean needsFullEntityScreen, EntityScreenContext e
...
         throw new RuntimeException("Unexpected Frame Request: " + sessionWrapper.getNeededData());
     }
 
+    /**
+     * Get next screen for the current request, based on the current state of the session,
+     * but only initialize and return the screen if it is of type `FormplayerSyncScreen`.
+     */
+    public FormplayerSyncScreen getNextScreenIfSyncScreen(boolean needsFullEntityScreen, EntityScreenContext entityScreenContext) throws CommCareSessionException {
+        String next = sessionWrapper.getNeededData(sessionWrapper.getEvaluationContext());
+        if (next.equalsIgnoreCase(SessionFrame.STATE_DATUM_COMPUTED)) {
```
shubham1g5, line 272: We need to check for `next == null`

Jtang-1: I had it at first but i ended up removing it, see this [comment](https://github.com/dimagi/formplayer/pull/1608#discussion_r1705980081). But let me know if you disagree

shubham1g5: It seems like the tests are failing currently due to `next` being `null`

Jtang-1: ah yes you're right, thank you [0aa1810](https://github.com/dimagi/formplayer/pull/1608/commits/0aa1810556afd85e075e58f8493d8259d2590939)

src/main/java/org/commcare/formplayer/session/MenuSession.java
Jtang-1: [593c77e](https://github.com/dimagi/formplayer/pull/1608/commits/593c77ede4be17ad921e53e3fde55a4c67e115d6)

src/test/java/org/commcare/formplayer/tests/EndpointLaunchTest.java
```diff
@@ -123,6 +129,21 @@ public void testEndpoints() throws Exception {
...
+    @WithHqUser(enabledToggles = {TOGGLE_SESSION_ENDPOINTS})
+    public void testEndpointsWithInlineCaseSearch() throws Exception {
+        configureQueryMock();
+        HashMap<String, String> endpointArgs = new HashMap<>();
+        endpointArgs.put("case_id", "0156fa3e-093e-4136-b95c-01b13dae66c6");
+        CommandListResponseBean commandListResponse = sessionNavigateWithEndpoint(APP_NAME,
+                "inline_w_display_cond_case_list",
+                endpointArgs,
+                CommandListResponseBean.class);
+        assert commandListResponse.getCommands().length == 2;
```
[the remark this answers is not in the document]
Jtang-1: Good call. A case is needed so it can be selected and move onto the next menu. Of the two forms, I made one display condition false and the other true. Now the test will auto select the visible one and auto advance into the visible form.

[b92cbdc](https://github.com/dimagi/formplayer/pull/1608/commits/b92cbdc583df07c96d3d16832f07f93471390657)

shubham1g5 (approved):

PR #1614, opened 2024-08-16
Jt/support primary app id on install

Jtang-1:
### Migrations

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
+++ b/src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
@@ -42,6 +42,9 @@ public class SerializableFormSession implements Serializable {
     @Column(name = "appid", updatable = false)
     private String appId;
 
+    @Column(name = "appversion", updatable = false)
+    private String appVersion;
+
     @Column(name = "caseid", updatable = false)
     private String restoreAsCaseId;
 
@@ -108,7 +111,8 @@ public class SerializableFormSession implements Serializable {
             String initLang,
             boolean inPromptMode,
             Map<String, String> sessionData,
-            Map<String, FunctionHandler[]> functionContext) {
+            Map<String, FunctionHandler[]> functionContext,
+            String appVersion) {
         this.domain = domain;
         this.asUser = asUser;
         this.appId = appId;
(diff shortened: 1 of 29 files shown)


src/main/java/org/commcare/formplayer/sqlitedb/ApplicationDBPath.java
```diff
@@ -11,17 +11,23 @@ class ApplicationDBPath extends DBPath {
...
         this.appId = appId;
+        this.appVersion = appVersion;
     }
 
     @Override
     public String getDatabasePath() {
-        return DbUtils.getDbPathForUser(domain, username, asUsername) + File.separator + appId;
+        String path = DbUtils.getDbPathForUser(domain, username, asUsername) + File.separator + appId;
+        if (appVersion != null) {
+            path = path + "_" + appVersion;
```
shubham1g5, line 30: optional: Seems like we are doing this concatenation of `appId +  "_" + appVersion` at multiple places in the PR and think would be nice to have this as a util method somewhere and reuse it.

src/test/resources/requests/delete_db/delete_db.json
```diff
@@ -1,5 +1,6 @@
 {
   "app_id": "casetestappid",
   "username": "casetestuser",
-  "domain": "casetestdomain"
+  "domain": "casetestdomain",
+  "appVersion": null
```
shubham1g5, line 5: Confirming that tests passes even if we omit `"appVersion": null` from here ?

src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
```diff
@@ -42,6 +42,9 @@ public enum SubmitStatus {
     @Column(name = "appid", updatable = false)
     private String appId;
 
+    @Column(name = "appversion", updatable = false)
+    private String appVersion;
```
shubham1g5, line 46: Think it should not but think we should verify that it doesn't break Incomplete forms on resuming after deploy (can be part of QA)

shubham1g5 (commented):
Looks quite solid to me overall. Just some questions.  Also can we link spec in the PR description. Thanks!

Jtang-1:
Pivoted to a more simple solution that does not require FP changes ([HQ PR](https://github.com/dimagi/commcare-hq/pull/35013/files/6514a2b493270d989d3904e6c06a58f878f66633..f88b6a7ceb5ce33192c93231635cbd3959db5938)) so closing this.

PR #1620, opened 2024-08-24
Fix persistent menu for eof nav

shubham1g5:
## Product Description
https://dimagi.atlassian.net/browse/QA-7001

## Technical Summary
Fix an issue for persistent menu with eof navigation where duplicate menus get added to the persistent menu due to `getNextScreen` getting called more than once for a single menu screen. 

### Safety story

Relying on test coverage, we have good coverage  around navigation paths on FP including EOF navigation. 

### Automated test coverage

PR adds additional testing around the bug. 

### QA Plan

None

### Rollback instructions

### Review


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -371,7 +371,10 @@ private Screen autoAdvanceSession(
             previousScreen = nextScreen;
             iterationCount += 1;
 
-            nextScreen = menuSession.getNextScreen(needsFullEntityScreen, entityScreenContext);
+            // skips the call to to getNextScreen for first iteration if we already have nextScreen
+            if (iterationCount > 1 || nextScreen == null) {
+                nextScreen = menuSession.getNextScreen(needsFullEntityScreen, entityScreenContext);
+            }
```
shubham1g5, line 377: This is the crux of the issue where we are no longer calling getNextScreen for the first iteration if we already have it evaluated.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -356,13 +356,13 @@ public BaseResponseBean advanceSessionWithSelections(MenuSession menuSession,
      * @throws CommCareSessionException
      */
     private Screen autoAdvanceSession(
+            @Nullable Screen nextScreen,
```
Jtang-1, line 359: nit: this naming seems confusing, isn't it actually the current screen we want to pass in?

This is also only relevant for EOF right? I would be a bit confused on when this should or shouldn't be null when using the function. Maybe a docstring would help?

Jtang-1 (approved):

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
shubham1g5: Added [here](https://github.com/dimagi/formplayer/pull/1620/commits/6e5ce3d44e9002877201ccbaa8e84c72a981a624).

> This is also only relevant for EOF right? 

Not really, it can be used in another navigation pathways as well if the screen is already evaluated for the session. The other pathway where it's passed as `null` in the PR doesn't have the screen evaluated already and therefore we need it to re-calculate as part of this method. 

> I would be a bit confused on when this should or shouldn't be null when using the function.

Agree it's a bit tricky as one will need to check whether we have already evaluated the screen as per the current session state before calling this.

Jtang-1 (approved):

PR #1622, opened 2024-08-29
Fix Smartlink with End of Form Navigation Compatibility

Jtang-1:
## Product Description

Navigating to form’s smartlink lands the user all the way to the form’s defined End of Form instead of to the form. This PR fixes that and lands the user at the defined smartlink form.

## Technical Summary

[USH-3901 ](https://dimagi.atlassian.net/browse/USH-3901)
The problem is that on advanceSessionWithEndpoint → executeAndRebuildSession, [ops](https://github.com/dimagi/commcare-core/blob/d09e72adeab33cff86a324da2fa0d542a62bf2c3/src/main/java/org/commcare/session/CommCareSession.java#L880) contain the stack operation of EoF nav and this is being executed.

The fix here checks if the screen after a Formplayer Sync is a form. If it is a form, then that should be the end of the navigation and it skips adding post entry stack operations to the stack.

### Safety story

Locally tested and the change is restricted to smartlinks that require a formplayer sync.

### QA Plan

Will QA different smartlinks

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -813,7 +813,10 @@ public class MenuSessionRunnerService {
                 try {
                     screen.init(sessionWrapper);
                     doPostAndSync(menuSession, (FormplayerSyncScreen)screen);
-                    executeAndRebuildSession(menuSession);
+                    Screen nextScreen = menuSession.getNextScreen(false, new EntityScreenContext());
+                    if (nextScreen == null && menuSession.getSessionWrapper().getForm() == null) {
+                        executeAndRebuildSession(menuSession);
+                    }
                 } catch (CommCareSessionException ccse) {
                     throw new RuntimeException("Unable to claim case.");
                 }


src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -813,7 +813,10 @@ public BaseResponseBean advanceSessionWithEndpoint(MenuSession menuSession, Stri
                 try {
                     screen.init(sessionWrapper);
                     doPostAndSync(menuSession, (FormplayerSyncScreen)screen);
-                    executeAndRebuildSession(menuSession);
+                    Screen nextScreen = menuSession.getNextScreen(false, new EntityScreenContext());
+                    if (nextScreen == null && menuSession.getSessionWrapper().getForm() == null) {
```
Jtang-1, line 817: Maybe it should just check that the op being processed is not the last op in the loop? Because should that be where we want to land up no matter what?

shubham1g5 (approved):
@Jtang-1 No real concerns with the change but think we are heading in a wrong direction by putting session running code outside the `rebuildSessionFromFrame` method. I think the call to `getNextScreen` should only really happen inside `rebuildSessionFromFrame` method and as such we should move all code related to `FormplayerSyncScreen` as a if condition in [the session navigation loop here](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L97). I would assume that should elimitate need of doing special handling like this.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
```diff
@@ -808,16 +782,6 @@ public BaseResponseBean advanceSessionWithEndpoint(MenuSession menuSession, Stri
         // Sync requests aren't run when executing operations, so stop and check for them after each operation
         for (StackOperation op : endpoint.getStackOperations()) {
             sessionWrapper.executeStackOperations(new Vector<>(Arrays.asList(op)), evalContext);
-            FormplayerSyncScreen screen = menuSession.getNextScreenIfSyncScreen(false, new EntityScreenContext());
-            if (screen != null) {
-                try {
-                    screen.init(sessionWrapper);
-                    doPostAndSync(menuSession, (FormplayerSyncScreen)screen);
-                    executeAndRebuildSession(menuSession);
```
Jtang-1, line 816: I don't think `executeAndRebuildSession` is needed anymore since the sync screen is now handled while the session is being rebuilt. But i'm not completely confident, @shubham1g5 do you know?

shubham1g5: Agree that it should not be needed anymore.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -315,4 +337,30 @@ private void logStepNotInEntityScreenError(EntityScreen entityScreen, SessionDat
         neededDatum.getDataId(), step.getValue(), nodeSetString, referencesString));
 
     }
+
+    /**
+     * Execute the post request associated with the sync screen and perform a sync if necessary.
+     */
+    public void doPostAndSync(MenuSession menuSession, FormplayerSyncScreen screen) throws SyncRestoreException {
```
shubham1g5, line 344: nit: A Factory class doesn't seem like a great place for this public method, think it would be better to put it in it's own helper class if possible.

shubham1g5, line 344: nit: A Factory class doesn't seem like a great place for this public method, think it would be better to put it in it's own helper class if possible.

shubham1g5 (approved):
Thanks! looks great. Left a comment around re-organisation but nothing blocking. 

Also seems like we have quite extensive coverage with unit tests around smart linking behaviour, so think would be fine if we want to skip QA here.

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
Jtang-1: that makes sense, [f817923](https://github.com/dimagi/formplayer/pull/1622/commits/f817923ccd4b5aa1fa100636d70b2035571aaa2a)

src/test/java/org/commcare/formplayer/tests/CaseClaimNavigationTests.java
```diff
@@ -218,6 +219,40 @@ public void testPostInEntryWithQuery_RelevantFalse() throws Exception {
...
+                    EntityListResponse.class);
+        }
+        assertEquals(1, entityListResponse.getEntities().length);
+        assertEquals("0156fa3e-093e-4136-b95c-01b13dae66c6",
+                entityListResponse.getEntities()[0].getId());
+
+        selections.add("0156fa3e-093e-4136-b95c-01b13dae66c6");
+        FormEntryResponseBean formEntryResponseBean;
+        try (
+                MockRequestUtils.VerifiedMock ignoredPostMock = mockRequest.mockPost(true, 2);
```
shubham1g5, line 246: why does post happens twice here ?

src/test/java/org/commcare/formplayer/tests/CaseClaimNavigationTests.java
```diff
@@ -218,6 +219,40 @@ public void testPostInEntryWithQuery_RelevantFalse() throws Exception {
         verifyNoInteractions(webClientMock);
     }
 
+    @Test
+    public void testNormalClaimWithPostInEntry() throws Exception {
```
shubham1g5, line 223: thanks for being thorough here by writing tests for encountered regressions.

src/test/java/org/commcare/formplayer/tests/CaseClaimNavigationTests.java
Jtang-1: It's because with the configuration
- Menu with case search
- Menu containing both followup and registration form
- sync upon form entry

Then the app work flow would be Select Menu -> select form -> case search query -> select entity -> form entry

Autoadvance session always upon navigate menu. But after case search select entity, `rebuildSessionFromFrame` also runs. So both `autoAdvanceSession` and `rebuildSessionFromFrame` will try to post. Although logic is now added to `doPostAndSync` such that the post does not actually happen twice. But maybe that conditional check should be done by the caller so that `doPostAndSync` does not get called twice?

shubham1g5: Makes sense, I don't think that will solve the issue since while rebuilding the session we reset the session state which I think will remove the `EXTRA_POST_SUCCESS` as well.  I was thinking that the [relevant condition](https://github.com/dimagi/formplayer/blob/master/src/test/resources/archives/case_claim_post_in_entry/suite.xml#L355) on the post itself should stop the second post from happening again as after the first post, we perform a restore which we are mocking here to include the selected case.

Jtang-1: By "while rebuilding the session we reset the session state" Do you mean the reset done [here](https://github.com/dimagi/formplayer/blob/0b20bf3c6e87f40ddb331cb8b3ea2f2b9053cc3b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L92)?

The two calls to `doPostAndSync` happens after that reset so `EXTRA_POST_SUCCESS ` is not removed between the calls. First during the `rebuildSessionFromFrame` and second during `autoAdvanceSession`, I think both are the post related to form entry. This is the logging output
```
   autoadvancing session. screen is: MenuScreen['Case Claim' with choices [Entry with id m3-f0, Entry with id m3- 
    f.1]]
    autoadvancing session. screen is: EntityScreen [id=case_id, selection=null]
    autoadvancing session. screen is: QueryScreen[Case Claim]
    autoadvancing session. screen is: EntityScreen [id=case_id, selection=null]
    autoadvancing session. screen is: org.commcare.formplayer.screens.FormplayerSyncScreen@7a5073f8
    rebuilding session from frame, screen isMenuScreen['Case Claim' with choices [Entry with id m0-f0, Menu with id m1, Menu with id m2, Menu with id m3]]
    rebuilding session from frame, screen isMenuScreen['Case Claim' with choices [Entry with id m3-f0, Entry with id m3-f1]]
    rebuilding session from frame, screen isEntityScreen [id=case_id, selection=null]
    rebuilding session from frame, screen isorg.commcare.formplayer.screens.FormplayerSyncScreen@108eb810
    rebuilding sesssion from frame: screen.getSessionSuccessStatus() is null
    autoadvancing session. screen is: org.commcare.formplayer.screens.FormplayerSyncScreen@108eb810
    autoadvancingsession screen.getSessionSuccessStatus() is true
```

I think the user is being restored since the [mock is called twice](https://github.com/dimagi/formplayer/blob/45e146346aa3060d6a94b9aee2ff6a861ca9a5cb/src/test/java/org/commcare/formplayer/tests/CaseClaimNavigationTests.java#L247)? Also, I think the post we're interested in is the [entry's](https://github.com/dimagi/formplayer/blob/45e146346aa3060d6a94b9aee2ff6a861ca9a5cb/src/test/resources/archives/case_claim_post_in_entry/suite.xml#L345-L347), not the remote request's. If I changed that `relevant` condition to `count(instance('casedb')/casedb/case[@case_id=instance('commcaresession')/session/data/case_id]) = 0`, then the tests calls `doPostAndSync` only once.

shubham1g5: > I think the post we're interested in is the [entry's](https://github.com/dimagi/formplayer/blob/45e146346aa3060d6a94b9aee2ff6a861ca9a5cb/src/test/resources/archives/case_claim_post_in_entry/suite.xml#L345-L347), not the remote request's

Ahh yes, I didn't realise we have a post in entry as well. Given that, My sense is the 2 posts that are happening here belongs to 2 different code areas with first one being the remote request one and later being the entry post. I say this because we are adding `action 1` in the selections which points to the remote request with command id `case_search.m1`. This should trigger the first post on case selection and then when the user entes the form we trigger the entry->post as well. This also will explain why you only see 1 post on changing the relevancy condition for entry post. 

Confirming that overall this seems expected behaviour in context of this test.

src/main/java/org/commcare/formplayer/services/MenuSessionRunnerHelper.java
```diff
@@ -34,7 +34,7 @@ public class MenuSessionRunnerHelper {
     public void doPostAndSync(MenuSession menuSession, FormplayerSyncScreen screen) throws SyncRestoreException {
         boolean shouldSync = false;
         try {
-            if (screen.getSessionSuccessStatus() == null) {
+            if (screen.isPostAlreadyDone() == null) {
```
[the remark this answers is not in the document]
Jtang-1: oops, thanks for catching that [b1b7590](https://github.com/dimagi/formplayer/pull/1622/commits/b1b7590afb95c41cb8669128f71ffba2e39419b0)

I also changed the name to `isPostAlreadyAttempted`

shubham1g5 (approved):

shubham1g5 (approved):

PR #1627, opened 2024-10-01
Revert "Fix Smartlink with End of Form Navigation Compatibility"

Jtang-1:
Reverts dimagi/formplayer#1622

To revert bug seen here: https://dimagi.atlassian.net/browse/USH-4975. A revert is done first from this [plan](https://dimagi.atlassian.net/browse/USH-4975?focusedCommentId=348420)


shubham1g5 (approved):
No concerns as such, but we should add a regression test for the bug probably as part of a separate PR.

Jtang-1:
To close the loop, I've added a regression test for this bug in this commit https://github.com/dimagi/formplayer/pull/1628/commits/9a8c91f72d70a5cef1f715e5e146ac16392981e2

PR #1628, opened 2024-10-02
Fix Smartlink with End of Form Navigation Compatibility

Jtang-1:
## Product Description

Navigating to form’s smartlink lands the user all the way to the form’s defined End of Form instead of to the form. This PR fixes that and lands the user at the defined smartlink form.

## Technical Summary

[USH-3901](https://dimagi.atlassian.net/browse/USH-3901) and [USH-5001](https://dimagi.atlassian.net/browse/USH-5001)
The problem is that on advanceSessionWithEndpoint → executeAndRebuildSession, [ops](https://github.com/dimagi/commcare-core/blob/d09e72adeab33cff86a324da2fa0d542a62bf2c3/src/main/java/org/commcare/session/CommCareSession.java#L880) contain the stack operation of EoF nav and this is being executed. 

The solution here depends on the fact that only the last stack defined in the `<endpoint>` stack definition defines the navigation. While the stacks prior define only case claim. Meaning that when processing the last stack, we do not want to pop new frames into the stack (meaning that we skip `executeAndRebuildSession`). 

Additional context: An initial attempt to resolving this issue (https://github.com/dimagi/formplayer/pull/1622) resulted in this [bug](https://dimagi.atlassian.net/browse/USH-4975) and subsequently a [revert](https://github.com/dimagi/formplayer/pull/1627). The 

### Safety story

Locally tested and the change is restricted to smartlinks that require a Formplayer sync.

### Automated test coverage

There exists tests surrounding smartlinking. I added tests that would catch
-  The bug involving smartlinking to a form that defines an EOF nav
-  Scenarios where both rebuildSessionFromFrame and autoAdvanceSession post https://github.com/dimagi/formplayer/pull/1628/commits/3e129e754e5f4460a7949baf21c690ce83072aef. This occurs with changes in this reverted PR https://github.com/dimagi/formplayer/pull/1622 where `FormplayerSyncScreen` is processed in `rebuildSessionFromFrame`. Consequently, it is not an existing logic flow but I kept the test here to future proof this regression.
- Smartlinking to a case that is not yet claimed in user's `casedb` (which occured in this [ticket](https://dimagi.atlassian.net/browse/USH-4975))

### QA Plan

will QA

### Rollback instructions

### Review

[USH-3901]: https://dimagi.atlassian.net/browse/USH-3901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[USH-5001]: https://dimagi.atlassian.net/browse/USH-5001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java
@@ -805,15 +805,27 @@ public class MenuSessionRunnerService {
             throw new RuntimeException(
                     String.format("Invalid arguments supplied for link.%s%s", missingMessage, unexpectedMessage));
         }
+
+        Vector<StackOperation> stackOperations = endpoint.getStackOperations();
+        int totalOps = stackOperations.size();
+
         // Sync requests aren't run when executing operations, so stop and check for them after each operation
-        for (StackOperation op : endpoint.getStackOperations()) {
+        for (int i = 0; i < totalOps; i++) {
+            StackOperation op = stackOperations.get(i);
             sessionWrapper.executeStackOperations(new Vector<>(Arrays.asList(op)), evalContext);
+
             FormplayerSyncScreen screen = menuSession.getNextScreenIfSyncScreen(false, new EntityScreenContext());
             if (screen != null) {
                 try {
                     screen.init(sessionWrapper);
-                    doPostAndSync(menuSession, (FormplayerSyncScreen)screen);
-                    executeAndRebuildSession(menuSession);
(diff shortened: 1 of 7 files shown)


shubham1g5 (approved):
Looks great!

PR #1630, opened 2024-10-08
Make language slug available via session metadata

minhaminha:
## Product Description

Users will now be able to reference the current language slug (i.e. 'en', 'es', etc.) on the screen by referencing `instance('commcaresession')/session/context/applanguage`

## Technical Summary

[Jira Ticket](https://dimagi.atlassian.net/browse/USH-4726)

Associated [commcare-core PR](https://github.com/dimagi/commcare-core/pull/1441)

This is the formplayer side of the changes. All this code does is ensures that `EntityScreenContext` has the right locale (a string) value provided on initialization and a null value for any instance that's not directly related to the workflow kicked off by a `navigateSessionWithAuth` request. Because this process continually recreates the context object, its the ideal entry point to insert this value which will be propagated to the eventual `generateRoot` function (actual code surrounding this part and further explanation will be in the commcare-core PR).

To be complete this PR also covers an instance of `ExternalDataInstance` initialization that will now optionally accept a locale argument. 

### Safety story

This shouldn't impact existing data since the `EntityScreenContext` object is constantly being recreated with updated information and most of the time the locale value will be null. 

### QA Plan

No plans for QA but will have delivery (who request this feature) do some testing to ensure it's doing what they want. 

- [x] This PR can be deployed after merge with no further considerations. (Of course the commcare-core PR will be merged first and this PR updated with those changes before merging/deploying).

### Rollback instructions

- [x] This PR can be reverted after deploy with no further considerations. (Ideally should be rolled back with attached commcare-core PR?)

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/MenuController.java
+++ b/src/main/java/org/commcare/formplayer/application/MenuController.java
@@ -79,7 +79,6 @@ public class MenuController extends AbstractBaseController {
                     sessionNavigationBean.getCasesPerPage(),
                     sessionNavigationBean.getSelectedValues(),
                     null,
-                    null,
                     isFuzzySearch);
             BaseResponseBean baseResponseBean = runnerService.advanceSessionWithSelections(menuSession,
                     sessionNavigationBean.getSelections(),
@@ -111,7 +110,6 @@ public class MenuController extends AbstractBaseController {
                 sessionNavigationBean.getSortIndex(),
                 sessionNavigationBean.getCasesPerPage(),
                 sessionNavigationBean.getSelectedValues(),
-                null,
                 detailSelection,
                 isFuzzySearch);
         BaseResponseBean baseResponseBean = runnerService.advanceSessionWithSelections(
@@ -167,7 +165,6 @@ public class MenuController extends AbstractBaseController {
     public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBean sessionNavigationBean,
             @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken,
             HttpServletRequest request) throws Exception {
-        String locale = sessionNavigationBean.getLocale();
         String[] selections = sessionNavigationBean.getSelections();
         if (selections == null) {
             selections = new String[0];
@@ -178,7 +175,6 @@ public class MenuController extends AbstractBaseController {
                 sessionNavigationBean.getSortIndex(),
                 sessionNavigationBean.getCasesPerPage(),
                 sessionNavigationBean.getSelectedValues(),
-                locale,
                 null,
                 storageFactory.getPropertyManager().isFuzzySearchEnabled());
         BaseResponseBean response = runnerService.advanceSessionWithSelections(
--- a/src/main/java/org/commcare/formplayer/application/MenuController.java
+++ b/src/main/java/org/commcare/formplayer/application/MenuController.java
@@ -79,6 +79,7 @@ public class MenuController extends AbstractBaseController {
                     sessionNavigationBean.getCasesPerPage(),
                     sessionNavigationBean.getSelectedValues(),
                     null,
+                    null,
                     isFuzzySearch);
             BaseResponseBean baseResponseBean = runnerService.advanceSessionWithSelections(menuSession,
                     sessionNavigationBean.getSelections(),
@@ -110,6 +111,7 @@ public class MenuController extends AbstractBaseController {
                 sessionNavigationBean.getSortIndex(),
                 sessionNavigationBean.getCasesPerPage(),
                 sessionNavigationBean.getSelectedValues(),
+                null,
                 detailSelection,
                 isFuzzySearch);
         BaseResponseBean baseResponseBean = runnerService.advanceSessionWithSelections(
@@ -165,6 +167,7 @@ public class MenuController extends AbstractBaseController {
     public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBean sessionNavigationBean,
             @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken,
             HttpServletRequest request) throws Exception {
+        String locale = sessionNavigationBean.getLocale();
         String[] selections = sessionNavigationBean.getSelections();
         if (selections == null) {
             selections = new String[0];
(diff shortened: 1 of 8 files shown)


src/main/java/org/commcare/formplayer/session/FormplayerInstanceInitializer.java
```diff
@@ -61,10 +62,16 @@ protected InstanceRoot setupSessionData(ExternalDataInstance instance) {
 
         Hashtable<String, String> userProperties = u.getProperties();
 
+        String appLang = locale;
+        if (locale == null) {
+            String[] locales = Localization.getGlobalLocalizerAdvanced().getAvailableLocales();
+            appLang =  locales[locales.length-1];
```
Robert-Costello, line 68: this grabs the last locale created? is that always right one?

minhaminha: thanks so much for pointing this out, totally forgot to go back to this. The locales list is always as follows: ['default' (like literally just a string called default), <the actual language slug user set as default>, <all other slugs in order the were added]. The `locales.length-1` was previously a placeholder and wouldn't have worked if there were more than one languages. Added a similar comment in the code too to avoid confusion.

Robert-Costello (approved):

src/main/java/org/commcare/formplayer/application/MenuController.java
```diff
@@ -165,6 +167,7 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
     public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBean sessionNavigationBean,
             @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken,
             HttpServletRequest request) throws Exception {
+        String locale = sessionNavigationBean.getLocale();
```
shubham1g5, line 170: How does HQ get to know of current locale to send here ?

src/main/java/org/commcare/formplayer/session/FormplayerInstanceInitializer.java
```diff
@@ -61,10 +63,22 @@ protected InstanceRoot setupSessionData(ExternalDataInstance instance) {
...
+        String[] locales = Localization.getGlobalLocalizerAdvanced().getAvailableLocales();
+        if (appLang == null || !Arrays.asList(locales).contains(appLang)) {
+            // the ordering is always ['default', <true default slug>, <all other slugs in order they were added>]
+            if (locales.length >= 2) {
+                appLang =  locales[1];
+            } else {
+                // to pass tests
+                appLang = locales[0];
+            }
+        }
```
shubham1g5, line 76: My read of the code here is you are trying to get the current selected locale ? If so, think we can use `Localization.getCurrentLocale()` directly.

shubham1g5 (changes requested):
This will need test coverage before the merge, you can look at the tests added in https://github.com/dimagi/formplayer/pull/1596 for an example.

src/main/java/org/commcare/formplayer/application/MenuController.java
minhaminha: Honestly not too sure! I just assumed all navigation request included the locale in the request body which gets set when initializing the sessionNavigationBean. Happy to be told otherwise but it's been accurately giving me the correct locale so far.

shubham1g5: did you mean HQ was already sending the current locale in the request body without us having to make a corresponding HQ change here ?

minhaminha: yes I believe so!

src/main/java/org/commcare/formplayer/session/FormplayerInstanceInitializer.java
minhaminha: I wish I had known about this function sooner because this is doing exactly what I was looking to do from the start, without having to do all the unnecessary steps of passing the locale to the root generator function... I might end up reverting the commcare-core PR :(

minhaminha: [partial revert PR](https://github.com/dimagi/commcare-core/pull/1447)

src/main/java/org/commcare/formplayer/beans/AuthenticatedRequestBean.java
```diff
@@ -22,6 +22,7 @@ public class AuthenticatedRequestBean {
     private int timezoneOffsetMillis = -1;
     private String timezoneFromBrowser = null;
     private String windowWidth;
+    private String locale;
```
shubham1g5, line 25: This doesn't seem like it's getting used anymore, think we can remove all bean model changes with the updated implementation.

shubham1g5 (changes requested):

src/main/java/org/commcare/formplayer/beans/AuthenticatedRequestBean.java
minhaminha: this is solely so that the test can pass!

minhaminha: nevermind! this was removed

Robert-Costello (approved):

src/test/java/org/commcare/formplayer/tests/BaseTestClass.java
```diff
@@ -828,6 +828,7 @@ <T> T sessionNavigateWithSelectedValues(String[] selections, String testName, St
         sessionNavigationBean.setSelections(selections);
         sessionNavigationBean.setSelectedValues(selectedValues);
         sessionNavigationBean.setWindowWidth(windowWidth);
+        sessionNavigationBean.setLocale(locale);
```
shubham1g5, line 831: How is this working given `sessionNavigationBean` does no longer have `setLocale` method ?

shubham1g5 (changes requested):
I don't think the tests are set up correctly. Since we are no longer dependent on `locale` being passed from the request body or request bean , `locale` should no longer be assigned to the `bean` for testing. Also we should add additional testing that incorporates changing of locale in between a test and check whether the xpath expression returns new changed locale correctly. 

Also double confirming that we have tested the new implementation on an app somewhere ?

shubham1g5 (approved):

minhaminha:
After some discussion, because delivery wants this feature to be usable ASAP, Shubham and I agreed to add (proper) test coverage for this change in a separate PR. I'll link it back to this one once it's up

PR #1637, opened 2024-11-01
Add test for language slug in metadata feature

minhaminha:
## Technical Summary

This is a followup PR to [this one](https://github.com/dimagi/formplayer/pull/1630), which allows for the current screen's language slug to be accessible through the session metadata. This just adds a test to cover that area. 

### Safety story

This only adds a test and test-specific code blocks to existing functions. `evaluateXpath` is currently written to initialize the session locales every time it's called (at least test-specific scenarios) which makes it necessary to manually globalLocalizer's available and default locales, because the test is looking to verify that that value can be accessed through the `applanguage` xpath,

### Automated test coverage

The whole PR is the test coverage. 

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/DebuggerController.java
+++ b/src/main/java/org/commcare/formplayer/application/DebuggerController.java
@@ -147,7 +147,15 @@ public class DebuggerController extends AbstractBaseController {
     public EvaluateXPathResponseBean evaluateXpath(@RequestBody EvaluateXPathRequestBean evaluateXPathRequestBean,
                                                    @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken) throws Exception {
         SerializableFormSession serializableFormSession = formSessionService.getSessionById(evaluateXPathRequestBean.getSessionId());
-        FormSession formEntrySession = formSessionFactory.getFormSession(serializableFormSession, evaluateXPathRequestBean.getWindowWidth());
+        FormSession formEntrySession;
+        String locale = evaluateXPathRequestBean.getLocale();
+        if (locale != null && locale.equals("test-locale")) {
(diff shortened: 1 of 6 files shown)


minhaminha:
HI @shubham1g5, apologies for the delay in getting this up! I made the mistake of trying to model it after the WindowWidth test which led me astray. I'm still not sure about how valid this test is given that it feels a bit heavy handed (having to manually set the GlobalLocalizer's available locales and current locale). I wrote it in the description too but this does indeed test the ability to access the localizer's locale through that specific xpath which I think is enough but wanted to get your thoughts on it.

shubham1g5 (changes requested):
@minhaminha I think the way we want to write these tests is by using the `changeLanguage` calls like [here](https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/LocalizationTests.java#L66). We should not really need to make a change in any bean models. The big difference between `windowWidth` and `locale` here is that HQ conveys FP the `windowWitdth` in the request payload while `locale` is already stored and maintained on FP side with HQ only needing to convey the local changes on browser side by using the `/change_locale` endpoint on FP.  Let me know if I should provide any more details here and thanks for following up with tests!

minhaminha:
@shubham1g5 all the bean changes are so that `evaluateXpathRequest` can be simulated properly. I've repeatedly tried to make use of changeLanguage but calling `evaluateXpath` in the test context forces a re-initialization of the session, which resets the available locales back to 'default' and 'en' and sets the global locale back to 'default'. Do you have any other ideas of how to preserve the specified locale and while also getting the xpath value?

shubham1g5:
@minhaminha Is there an app I can see this feature working ? I want to cross check the behaviour you are seeing with the actual feature behaviour.

PR #1638, opened 2024-11-05
Persistent menu auto select fix

shubham1g5:
## Product Description

https://dimagi.atlassian.net/browse/USH-5045

Fixes an issue where a reference to case list auto-selection will get added to the persistent menu and breadcrumbs. This only happens in case search and claim workflows. 

## Technical Summary

With case search and claim, we were triggering `auto-selection` on the case list with the `handleInput` method when `autoselectEntities()` is a better method to do so as it skips un-needed handling like adding a reference to persistent menu/ breadcrumbs. 

### Safety story

- Added tests  for persistent menus around various auto-select and auto-advance workflows
- Will test the exact workflow on staging before deploy. 

### QA Plan
NA

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -142,9 +142,10 @@ public class MenuSessionFactory {
                 }
 
                 if (currentStep != null && currentStep != NEXT_SCREEN && entityScreen.shouldBeSkipped()) {
-                    menuSession.handleInput(screen, currentStep, needsFullInit, true, false, entityScreenContext);
-                    screen = menuSession.getNextScreen(needsFullInit, entityScreenContext);
-                    continue;
+                    if (((EntityScreen)screen).autoSelectEntities(menuSession.getSessionWrapper())) {
+                        screen = menuSession.getNextScreen(needsFullInit, entityScreenContext);
+                        continue;
+                    }
                 }
                 if (currentStep == null && processedStepsCount != steps.size()) {
                     checkAndLogCaseIDMatchError(steps, processedSteps, neededDatum.getDataId());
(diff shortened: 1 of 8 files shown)


src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -142,9 +142,10 @@ public Screen rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper
                 }
 
                 if (currentStep != null && currentStep != NEXT_SCREEN && entityScreen.shouldBeSkipped()) {
-                    menuSession.handleInput(screen, currentStep, needsFullInit, true, false, entityScreenContext);
-                    screen = menuSession.getNextScreen(needsFullInit, entityScreenContext);
-                    continue;
+                    if (((EntityScreen)screen).autoSelectEntities(menuSession.getSessionWrapper())) {
```
MartinRiese, line 145: Is there no case where we get to this line but we should not auto select but handle the input?

MartinRiese (commented):
Otherwise, this looks good to me

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
Jtang-1: I have the same question, if the conditional is false, shouldn't we do `handleInput` and then `getNextScreen`?

shubham1g5: Yes, that was my confusion as well here. But it seems like an entity screen without auto-select will get handled [as part of the handleInput later below](https://github.com/dimagi/formplayer/blob/persistentMenuAutoSelectFix/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L200)

Jtang-1: ah, that makes sense

Jtang-1 (approved):
Thank you for the speedy fix and adding tests for this

PR #1640, opened 2024-11-06
Make form title translatable

avazirna:
## Product Description
This PR makes it possible to set the form title according to the user's preferred language or by changing the language in the middle of a form session. Currently, the form title is derived from the form XML, which is defined by the active language when the form is saved, after a change or form creation.
In the video at the bottom of the description, it is possible to see that:
- While the language of the user is `Portuguese` and most of the app content is shown in that language, when entering the form, the form title is in `English`;
- Changing the language in app builder to `Portuguese`, making a change to the app and then creating a new build, results in the form title be presented in that language

Additionally, switching to another language in the middle of a form session doesn't affect the form title in any way.

## Technical Summary
In terms of design decisions, two options were entertained during this work:
1. Setting the title in the appropriate language when preparing response objects to be sent to the front-end;
2. Maintaining the title in `SerializingFormSession` objects in the appropriate language;

Considering that the title in response objects is set based on `SerializedFormSession` objects, number 2 was the chosen option. This not only ensured fewer changes to the codebase, but it will also improve maintainability since changes to response classes are unlikely to affect this capability. 

HQ PR: https://github.com/dimagi/commcare-hq/pull/35444

### Safety story
Local tests were successful but one resource had to be update as with this change, the form title is retrieved from the translations.

### Rollback instructions

### Review

<video width="200" height="200" src="https://github.com/user-attachments/assets/dd610b6f-d153-48fe-89d7-ab72760b6695" />


src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -183,6 +190,30 @@ public FormSession(UserSqlSandbox sandbox,
...
+        }
+    }
+
+    private String getFormTitleLocaleKey() {
+        if (commandId != null && !commandId.equals("")) {
+            String preparedCommandId = commandId.replace("-", "");
+            return "forms." + preparedCommandId;
+        }
+        return "";
+    }
```
shubham1g5, line 215: there is no guarantee  in Xform spec that forms locale key will follow this format (it can be `dianasour.m0f0` for exmaple). This should be using the same logic as we use to load the display text for menus instead.

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
```diff
@@ -108,7 +108,8 @@ public NewFormResponse getResponse(NewSessionRequestBean bean, String postUrl) t
                 bean.getRestoreAsCaseId(),
                 null,
                 formplayerRemoteInstanceFetcher,
-                bean.getWindowWidth()
+                bean.getWindowWidth(),
+                null
```
shubham1g5, line 112: this doesn't need to be translated ?

src/main/java/org/commcare/formplayer/application/FormController.java
```diff
@@ -141,6 +141,7 @@ public FormEntryResponseBean changeLocale(@RequestBody ChangeLocaleRequestBean c
                 changeLocaleBean.getSessionId());
         FormSession formEntrySession = formSessionFactory.getFormSession(serializableFormSession, changeLocaleBean.getWindowWidth());
         formEntrySession.changeLocale(changeLocaleBean.getLocale());
+        formEntrySession.updateFormTitle(changeLocaleBean.getLocale());
```
shubham1g5, line 144: this can reside inside `formEntrySession.changeLocale` itself.

shubham1g5 (changes requested):
We should add some unit tests that can demonstrate the title changing appropriately with the change of language.

src/main/java/org/commcare/formplayer/objects/SerializableFormSession.java
```diff
@@ -51,7 +51,7 @@ public enum SubmitStatus {
     @Column(name = "menu_session_id", updatable = false)
     private String menuSessionId;
 
-    @Column(updatable = false)
```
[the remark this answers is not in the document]
avazirna: This should only impact DML statements, not the schema itself, but happy to test to make sure it doesn't break anything

src/main/java/org/commcare/formplayer/session/FormSession.java
avazirna: Noted

src/main/java/org/commcare/formplayer/services/NewFormResponseFactory.java
avazirna: because `SessionFrame` is null, I believe there is no `commandId`. But I'm not too familiar with this workflow, so happy to follow your lead.

avazirna:
> We should add some unit tests that can demonstrate the title changing appropriately with the change of language.

@shubham1g5 I have incorporated this into existing localization tests: [2f1f467](https://github.com/dimagi/formplayer/pull/1640/commits/2f1f467af3a7df17ea6f04ad4a283b8cde89da34)

src/main/java/org/commcare/formplayer/session/FormSession.java
```diff
@@ -108,7 +111,9 @@ public FormSession(SerializableFormSession session,
 
         this.sandbox = restoreFactory.getSandbox();
         this.windowWidth = windowWidth;
-
+        if (commCareSession != null && commCareSession.getCurrentEntry() != null) {
+            this.formTitleLocaleKey = commCareSession.getCurrentEntry().getText().getArgument();
```
shubham1g5, line 115: think we need to use`getText().evaluate()` which already seems to have [locale support](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/suite/model/Text.java#L213), so you might be able to use it's value directly without doing any further locale processing.

shubham1g5 (changes requested):

src/main/java/org/commcare/formplayer/session/FormSession.java
avazirna: HI @shubham1g5 I explored this approach before and revisited after seeing your comment. The issue is that it seems that `getText().evaluate()` [fetches the values using the default locale](https://github.com/dimagi/commcare-core/blob/25a931722f4441ae01c7d4531ba19a99ce9be96f/src/main/java/org/commcare/suite/model/Text.java#L221C24-L221C36) and not the current language

shubham1g5: I think that takes into account whatever locale is set to the Localiser, saying as we use ` Localization.get()` at other several places in the app to get the translation.

avazirna: Yes, but when we change language in the middle of a form session, it doesn't change the global locale, just the [form-level locale](https://github.com/dimagi/formplayer/blob/bb7cdbb2db7418834938e94222167be2aff98561/src/main/java/org/commcare/formplayer/session/FormSession.java#L636). So calling `getText().evaluate() `continues to return the value in the default language, right?

shubham1g5 (approved):

PR #1648, opened 2024-11-25
Set session cookie for user's preferred language

avazirna:
## Product Description
This PR sets a session cookie to store the user's preferred language for the current session. This cookie is to be used to override the user's favourite language in subsequent HTTP requests, making sure that the chosen language persists throughout the session.
This change also requires changes to the _frontend_, so another PR will be linked to this one.

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/MenuController.java
+++ b/src/main/java/org/commcare/formplayer/application/MenuController.java
@@ -1,6 +1,6 @@
 package org.commcare.formplayer.application;
 
-import static org.commcare.formplayer.util.Constants.SESSION_PREFERRED_LANGUAGE;
+import static org.commcare.formplayer.util.Constants.FORMPLAYER_SESSION_LANGUAGE;
 import static org.springframework.http.HttpStatus.NOT_FOUND;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/formplayer/application/MenuController.java
```diff
@@ -163,8 +166,14 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
     @UserRestore
     @AppInstall
     public BaseResponseBean navigateSessionWithAuth(@RequestBody SessionNavigationBean sessionNavigationBean,
-            @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken,
-            HttpServletRequest request) throws Exception {
+                        @CookieValue(Constants.POSTGRES_DJANGO_SESSION_ID) String authToken,
+                        @CookieValue(value = FORMPLAYER_SESSION_LANGUAGE, required = false) String sessionPrefLang,
```
snopoke, line 170: nit: Constants is already imported:
```suggestion
                        @CookieValue(Constants.FORMPLAYER_SESSION_LANGUAGE, required = false) String sessionPrefLang,
```

src/main/java/org/commcare/formplayer/application/MenuController.java
```diff
@@ -163,8 +166,14 @@ public EntityDetailListResponse getDetails(@RequestBody SessionNavigationBean se
...
+                        HttpServletRequest request,
+                        HttpServletResponse response) throws Exception {
+
+        if (sessionPrefLang == null || !sessionPrefLang.equals(sessionNavigationBean.getLocale())) {
+            response.addCookie(new Cookie(FORMPLAYER_SESSION_LANGUAGE, sessionNavigationBean.getLocale()));
+        }
```
snopoke, line 176: I don't follow what's going on here. We're resetting the cookie value if it doesn't match what's being sent in the request body. But the value in the request body is coming from the cookie right (unless the cookie is empty in which case the first part of the condition is met)?

avazirna:
The decision was to use the _localStorage_ instead of cookies, so this is no longer needed.

PR #1649, opened 2024-11-27
Reinstate auto select revert

shubham1g5:
## Product Description

Alternate implementation for skipping auto-select menu from contextual menus

[Original PR](https://github.com/dimagi/formplayer/pull/1638/files)
[Revert PR](https://github.com/dimagi/formplayer/pull/1642)
[Jira](https://dimagi.atlassian.net/browse/USH-5121)

## Technical Summary

Reverts [original solution](https://github.com/dimagi/formplayer/pull/1638) implemented in favor of a simple check before adding the entity list menu to contextual menus. 

## Safety Assurance

In addition to the unit test, I tested the correspondix fix on staging as per the steps outlined in [this comment](https://dimagi.atlassian.net/browse/USH-5100?focusedCommentId=354489). 

### Automated test coverage

PR adds a regression test for the issue that caused the revert. 

### QA Plan

None

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -142,11 +142,11 @@ public class MenuSessionFactory {
                 }
 
                 if (currentStep != null && currentStep != NEXT_SCREEN && entityScreen.shouldBeSkipped()) {
-                    if (((EntityScreen)screen).autoSelectEntities(menuSession.getSessionWrapper())) {
-                        screen = menuSession.getNextScreen(needsFullInit, entityScreenContext);
-                        continue;
-                    }
+                    menuSession.handleInput(screen, currentStep, needsFullInit, true, false, entityScreenContext);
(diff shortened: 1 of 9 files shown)


Jtang-1 (approved):
The changes look fine and seem to exclude the autoselect entity screen from the persistent menu. But, does this address the root cause of the issue? I think that selecting a form or any screen after the autoselect process would still result in an error?  Demonstrated by the workflow shown in this [screen recording](https://dimagi.atlassian.net/browse/USH-5045?focusedCommentId=351672) (the screen recording shows the original behavior - it didn't pull in this PR)

shubham1g5:
@Jtang-1 That issue was only present with persistent menu navigations (And not the main in-screen navigation) due to us trying to make a case selection on a auto-select screen with a guid. We solved it by removing the menu item from persistent menu entirely as that's indeed the behaviour we want, although you are right that if we try to make a selection on auto-selection screen with selections[] containing the guid of the selection, that issue may rise again but that'll be us supplying incorrect input to FP apis and something we should never intent to do

PR #1661, opened 2025-01-14
Purge old cases more often

minhaminha:
## Product Description

Resolves a bug in which users were being shown cases that belonged to locations the user was not assigned to. In the specific reported scenario, users were initially assigned to a parent location (Organization) and then assigned to a child location (Facility) but still being shown all cases belonging to children of the parent location. This turned out to be happening because when a user data sync is triggered every five minutes (or at least five minutes after the last sync - they are currently using both the `cc-skip-fixtures-after-submit` and `cc-sync-after-form` custom properties), `shouldPurge` is always set to false, so even scenarios in which users switched between child locations were also accumulating cases belonging to locations they were previously assigned to. 

All this does it change that value to true to ensure old cases are purged, similar to what happens when you click the manual sync button in the web apps home page (which this specific domain has disabled... for some reason). 

## Technical Summary

[Jira Ticket](https://dimagi.atlassian.net/browse/USH-5143)
Thanks Ethan for inspiring me to have another look at what was going on in formplayer after I've been stuck looking at the fixture syncing side of things for a little while!!

### Safety story

Tested locally. Should be safe as all it does is increase the scope of the sync to be similar to that of a manual sync. 

I'm not sure what the performance implication of this is, since this is part of a workflow that'll happen at most every 5 minutes, though since this is limited to specifically looking at a user's cases, I think it should be fine. 

### QA Plan

N/a

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -273,7 +273,7 @@ public class RestoreFactory {
             return getSqlSandbox();
         } else {
             getSQLiteDB().createDatabaseFolder();
-            return performTimedSync(false, false, false);
+            return performTimedSync(true, false, false);
         }
     }
 
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -64,6 +64,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.sql.SQLException;
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -316,9 +317,13 @@ public class RestoreFactory {
                 );
                 sandbox.writeSyncToken();
 
-                if (!shouldPurge) {
+                if (!shouldPurge && !oldSandboxLocations.isEmpty()) {
                     String newSandboxLocations = UserUtils.getUserLocationsByDomain(domain, sandbox);
-                    hasLocationChanged = !oldSandboxLocations.isEmpty() && !oldSandboxLocations.equals(newSandboxLocations);
+                    String[] oldArray = oldSandboxLocations.split(" ");
+                    String[] newArray = newSandboxLocations.split(" ");
+                    Arrays.sort(oldArray);
+                    Arrays.sort(newArray);
+                    hasLocationChanged = !Arrays.equals(oldArray, newArray);
                 }
                 return sandbox;
             } catch (InvalidStructureException | SQLiteRuntimeException e) {
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -338,7 +338,7 @@ public class RestoreFactory {
                             break;
                         }
                     }
-                    if (!oldSandboxLocations.equals(newSandboxLocations)) {
+                    if (!oldSandboxLocations.isEmpty() && !oldSandboxLocations.equals(newSandboxLocations)) {
                         hasLocationChanged = true;
                     }
                 }
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -318,11 +318,7 @@ public class RestoreFactory {
 
                 if (!shouldPurge) {
                     String newSandboxLocations = UserUtils.getUserLocationsByDomain(domain, sandbox);
-                    if (!oldSandboxLocations.isEmpty() && !oldSandboxLocations.equals(newSandboxLocations)) {
-                        hasLocationChanged = true;
-                    } else {
-                        hasLocationChanged = false;
-                    }
+                    hasLocationChanged = !oldSandboxLocations.isEmpty() && !oldSandboxLocations.equals(newSandboxLocations);
                 }
                 return sandbox;
             } catch (InvalidStructureException | SQLiteRuntimeException e) {
--- a/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/RestoreFactory.java
@@ -237,7 +237,6 @@ public class RestoreFactory {
                         null,
                         extraTags
                 );
-                hasLocationChanged = false;
             } catch (InvalidCaseGraphException e) {
                 FormplayerSentry.captureException(e, SentryLevel.WARNING);
                 // if we have not already, do a fresh sync to try and resolve state
@@ -321,6 +320,8 @@ public class RestoreFactory {
                     String newSandboxLocations = UserUtils.getUserLocationsByDomain(domain, sandbox);
                     if (!oldSandboxLocations.isEmpty() && !oldSandboxLocations.equals(newSandboxLocations)) {
                         hasLocationChanged = true;
+                    } else {
+                        hasLocationChanged = false;
                     }
                 }
                 return sandbox;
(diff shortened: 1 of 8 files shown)


shubham1g5 (commented):
@minhaminha do we have a clear repro to support this hypothesis ?

Also I do think we will need to test performance here as purging is an intensive process and may take long time depending on # of cases in user db which will defeat purpose of settings like `cc-skip-fixtures-after-submit` as the goal there is to make sync faster.

minhaminha:
hi @shubham1g5

I wrote up the reproduction steps I used [here](https://dimagi.atlassian.net/browse/USH-5143?focusedCommentId=361461).

I agree with the performance aspect and am open to test it (though I'm not exactly sure how), but just so you know, we're currently looking at [getting rid of](https://dimagi.atlassian.net/browse/USH-5141) `cc-skip-fixtures-after-submit` in favor of "per-fixture sync logic" such as configuring each fixture to sync at different intervals (Woody proposes at least 8hrs since the last sync for UCRs, for example). 

With that in mind, would there be less of a performance concern if I also added a similar interval for case purging? say every hour or so? I'm also looking to add a conditional to check whether user location assignment has changed (really comparing location_ids in the user sandbox to the extracted value from the restoreStream... which is a little harder to do than I imagined). Though even with the conditional, I'll need to at least purge cases for affected users at least once to get them out of the buggy state they're in.

shubham1g5:
Thanks for the repro steps @minhaminha. Curious if this issue resolves itself after a form submission as we do have the [case purge ON](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/application/FormSubmissionHelper.java#L389) for sync after form submissions ? 


> I agree with the performance aspect and am open to test it (though I'm not exactly sure how)

We seems to be already sending [case purge timings](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/RestoreFactory.java#L233) to DD, so looking at BHA data should give us some indications on the potential slowness it may cause to turn this on.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -292,6 +300,42 @@ private UserSqlSandbox restoreUser(boolean skipFixtures) throws SyncRestoreExcep
                 UserSqlSandbox sandbox = getSqlSandbox();
                 FormplayerTransactionParserFactory factory = new FormplayerTransactionParserFactory(sandbox, true);
                 InputStream restoreStream = getRestoreXml(skipFixtures);
+                if (!shouldPurge) {
+                    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
shubham1g5, line 304: Don't think we need to do any additional parsing to determine location has changed or not. Instead we can monitor the # of locations in user sandbox before and after the restore has happened to determine it.

shubham1g5 (changes requested):
@minhaminha The approach looks reasonable to me but will need unit tests to go with it. Left a comment using which I think the code to check change in location can be simplified. 

Also curious if were able to look into [my comment above](https://github.com/dimagi/formplayer/pull/1661#issuecomment-2594563774) ?

minhaminha:
@shubham1g5 whoops sorry I forgot to answer your question but yes, it probably would resolve the issue but the app we're seeing this issue in has syncing after form submissions disabled, which is why we need a secondary trigger for case purging.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
minhaminha: ahh, that's true. Though, instead of number of locations I'll still be comparing the contents of `uUser.getProperty("commcare_location_ids")` since it's probably common to change from one location to another (so you're still assigned to just one location).

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
```diff
@@ -310,6 +315,15 @@ private UserSqlSandbox restoreUser(boolean skipFixtures) throws SyncRestoreExcep
                         extras
                 );
                 sandbox.writeSyncToken();
+
+                if (!shouldPurge) {
+                    String newSandboxLocations = UserUtils.getUserLocationsByDomain(domain, sandbox);
+                    if (!oldSandboxLocations.isEmpty() && !oldSandboxLocations.equals(newSandboxLocations)) {
```
shubham1g5, line 321: nit: you can directly assign `hasLocationChanged` to this statement and remove if/else

src/test/java/org/commcare/formplayer/tests/UserUtilsTests.java
```diff
@@ -16,4 +21,18 @@ public void testGetShortUsername() {
         assertEquals("dedicatedemployee@example.com",
                 UserUtils.getShortUsername("dedicatedemployee@example.com", "example"));
     }
+
+    @Test
+    public void testGetUserLocations() {
+        String test_loc_id = "test_location_id";
```
shubham1g5, line 27: variables need to be camelcased as per our coding standards for Java

src/test/java/org/commcare/formplayer/tests/RestoreFactoryTest.java
```diff
@@ -311,6 +315,23 @@ public void testGetRequestHeaders_UseHmacAuthEvenIfHqAuthPresent() throws Except
         );
     }
 
+    @Test
+    public void testChecksForLocationChanges() throws Exception {
```
shubham1g5, line 319: can we also do a check for multiple locations when unchanged, I want to verify hasLocationChanged remains `false` when same multiple locations are assigned to user before and after but in a a different order of array items.

shubham1g5 (changes requested):
Starting to look quite solid. I am unsure about the array equality operation we are performing here and hence asked for another test for gaining more confidence into the PR.

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
minhaminha: I'm not sure what you meant by this - are you suggesting I reset `hasLocationChanged` to false first and then set it to true if the locations have changed?

src/test/java/org/commcare/formplayer/tests/UserUtilsTests.java
minhaminha: whoops you're right - going back and forth between python and java code had me confused

src/main/java/org/commcare/formplayer/services/RestoreFactory.java
shubham1g5: you can do `hasLocationChanged = !oldSandboxLocations.isEmpty() && !oldSandboxLocations.equals(newSandboxLocations)` instead of if/else.

src/test/java/org/commcare/formplayer/tests/RestoreFactoryTest.java
minhaminha: Actually, it turns out when you assign a user multiple locations, the order you choose them in the location selector really doesn't matter, so this test is unnecessary. This is something I've noticed before (which is why I had it check the whole string) but it's apparently because django's multiple selector widgets (we specifically use our own `LocationSelectWidget` but I believe it applies here, since I'm seeing the same behavior) will always order user selected options in the order they appear in the initial choice dict.

shubham1g5: @minhaminha I don't think Formplayer should rely on another system to send it locations in a particular order. The question here is more -  if the locations are indeed in a random order, will Formplayer do the right thing here or not ?

src/test/java/org/commcare/formplayer/tests/UserUtilsTests.java
```diff
@@ -24,8 +24,8 @@ public void testGetShortUsername() {
 
     @Test
     public void testGetUserLocations() {
-        String test_loc_id = "test_location_id";
-        String test_domain = "test_domain";
+        String test_loc_id = "testLocationId";
```
shubham1g5, line 27: @minhaminha we need to camel case the variables, not their values.

shubham1g5 (approved):
Thanks for working through all the nitbits and think it's overall a great enhancement to the codebase.

minhaminha:
@shubham1g5 thank you for being so thorough and for the great pointers!!

PR #1662, opened 2025-01-17
Simplify default error message

gherceg:
## Product Description

This results in the following difference:

image

vs
image

## Technical Summary

The first example doesn't seem ideal to display to users. I believe `getResponseBodyAsString` should be sufficient to display the error to the user with enough information to take action or get help. 

### Safety story

This only impacts general errors on form submissions. I would assume that if a typical user comes across an error like this, they would be overwhelmed and potentially not even see the useful part of the message.

The risk is that we return errors with empty messages, in which case it will render like this:
image

Perhaps including the http status code at the very least would be helpful in the event the user receives an empty error message and reports it to us.

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/FormSubmissionHelper.java
+++ b/src/main/java/org/commcare/formplayer/application/FormSubmissionHelper.java
@@ -268,8 +268,7 @@ public class FormSubmissionHelper {
         } catch (HttpClientErrorException e) {
             return getErrorResponse(
                     context.getHttpRequest(), Constants.SUBMIT_RESPONSE_ERROR,
-                    String.format("Form submission failed with error response: %s, %s, %s",
-                            e.getMessage(), e.getResponseBodyAsString(), e.getResponseHeaders()),
+                    String.format("Form submission failed with error response: %s", e.getResponseBodyAsString()),
                     e);
         } finally {
             // If autoCommit hasn't been reset to `true` by the commit() call then an error occurred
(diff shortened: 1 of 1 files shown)


gherceg:
I certainly don't have all of the context here, so @shubham1g5 if you or you know of anyone else who has thoughts on this, feel free to ping them. It seems ideal to make a change _like_ this.

src/main/java/org/commcare/formplayer/application/FormSubmissionHelper.java
```diff
@@ -268,8 +268,7 @@ private SubmitResponseBean processFormXml(FormSubmissionContext context) throws
         } catch (HttpClientErrorException e) {
             return getErrorResponse(
                     context.getHttpRequest(), Constants.SUBMIT_RESPONSE_ERROR,
-                    String.format("Form submission failed with error response: %s, %s, %s",
-                            e.getMessage(), e.getResponseBodyAsString(), e.getResponseHeaders()),
+                    String.format("Form submission failed with error response: %s", e.getResponseBodyAsString()),
```
shubham1g5, line 271: agree that it will be helpful to have response code here when the reponse body is empty.

shubham1g5 (approved):

src/main/java/org/commcare/formplayer/application/FormSubmissionHelper.java
gherceg: Cool updated to use `getMessage` which appears to include the status code in testing.

shubham1g5 (approved):

PR #1677, opened 2025-02-25
domain specific apm traces

Robert-Costello:
## Technical Summary

[Jira Ticket ](https://dimagi.atlassian.net/browse/USH-5218)

[Related HQ PR](https://github.com/dimagi/commcare-hq/pull/35825)

[relevant Datadog docs ](https://docs.datadoghq.com/tracing/trace_pipeline/ingestion_mechanisms/?tab=java#force-keep-and-drop)

### Migrations

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
@@ -15,6 +15,8 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Component;
 
+import java.util.HashMap;
+
 @Component
 public class FormSessionFactory {
 
@@ -39,17 +41,20 @@ public class FormSessionFactory {
     @Autowired
     private CommCareSessionFactory commCareSessionFactory;
 
-    public FormSession getFormSession(SerializableFormSession serializableFormSession, String windowWidth) throws Exception {
+    public FormSession getFormSession(SerializableFormSession serializableFormSession, String windowWidth, boolean keepAPMTraces) throws Exception {
         CommCareSession commCareSession = commCareSessionFactory.getCommCareSession(serializableFormSession.getMenuSessionId());
-        return getFormSession(serializableFormSession, commCareSession, windowWidth);
+        return getFormSession(serializableFormSession, commCareSession, windowWidth, keepAPMTraces);
     }
 
     @NotNull
     public FormSession getFormSession(SerializableFormSession serializableFormSession,
-            @Nullable CommCareSession commCareSession, @Nullable String windowWidth) throws Exception {
+            @Nullable CommCareSession commCareSession, @Nullable String windowWidth, @Nullable boolean keepAPMTraces) throws Exception {
         FormplayerRemoteInstanceFetcher formplayerRemoteInstanceFetcher = new FormplayerRemoteInstanceFetcher(
                 runnerService.getCaseSearchHelper(),
                 virtualDataInstanceService);
+        HashMap metaSessionContext = new HashMap();
+        metaSessionContext.put("windowWidth", windowWidth);
+        metaSessionContext.put("keepAPMTraces", keepAPMTraces);
         return new FormSession(serializableFormSession,
                 restoreFactory,
                 formSendCalloutHandler,
@@ -57,7 +62,7 @@ public class FormSessionFactory {
                 commCareSession,
                 formplayerRemoteInstanceFetcher,
                 formDefinitionService,
-                windowWidth
+                metaSessionContext
         );
     }
 }
--- a/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
@@ -91,8 +91,8 @@ public class MenuSessionFactory {
     public Screen rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper caseSearchHelper,
             boolean respectRelevancy)
             throws CommCareSessionException, RemoteInstanceFetcher.RemoteInstanceException {
-//        boolean keepAPMTraces = (boolean)menuSession.getMetaSessionContext().get("keepAPMTraces");
-//        FormplayerDatadog.handleKeepDropAPMTraces(keepAPMTraces);
+        boolean keepAPMTraces = (boolean)menuSession.getMetaSessionContext().get("keepAPMTraces");
+        FormplayerDatadog.handleKeepDropAPMTraces(keepAPMTraces);
         Vector<StackFrameStep> steps = menuSession.getSessionWrapper().getFrame().getSteps();
         List<StackFrameStep> processedSteps = new ArrayList<>();
         menuSession.resetSession();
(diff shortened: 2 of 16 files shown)


src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
```diff
@@ -87,6 +91,8 @@ public Screen rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper
     public Screen rebuildSessionFromFrame(MenuSession menuSession, CaseSearchHelper caseSearchHelper,
             boolean respectRelevancy)
             throws CommCareSessionException, RemoteInstanceFetcher.RemoteInstanceException {
+//        boolean keepAPMTraces = (boolean)menuSession.getMetaSessionContext().get("keepAPMTraces");
```
MartinRiese: delete?

src/main/java/org/commcare/formplayer/util/Constants.java
```diff
@@ -200,4 +200,8 @@ public static final class TimingCategories {
     public static final String TOGGLE_INCLUDE_STATE_HASH = "FORMPLAYER_INCLUDE_STATE_HASH";
 
     public static final String AUTHORITY_COMMCARE = "COMMCARE";
+
+    // metaSessionContext keys
+    public static final String WINDOW_WIDTH = "windowWidth";
```
MartinRiese, line 205: Can you use the constants form commcare-core?

Robert-Costello: Yes, that's a good point.

src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
```diff
@@ -39,25 +44,28 @@ public class FormSessionFactory {
...
     @NotNull
     public FormSession getFormSession(SerializableFormSession serializableFormSession,
-            @Nullable CommCareSession commCareSession, @Nullable String windowWidth) throws Exception {
+            @Nullable CommCareSession commCareSession, @Nullable String windowWidth, @Nullable boolean keepAPMTraces) throws Exception {
         FormplayerRemoteInstanceFetcher formplayerRemoteInstanceFetcher = new FormplayerRemoteInstanceFetcher(
                 runnerService.getCaseSearchHelper(),
                 virtualDataInstanceService);
+        HashMap<String, Object> metaSessionContext = new HashMap<String, Object>();
+        metaSessionContext.put(WINDOW_WIDTH, windowWidth);
+        metaSessionContext.put(KEEP_APM_TRACES, keepAPMTraces);
```
shubham1g5, line 60: probably a better design to construct the `sessionContext` from the controller itself.

src/main/java/org/commcare/formplayer/util/FormplayerDatadog.java
```diff
@@ -234,4 +237,11 @@ private String getTagValueToSend(String tagName, String tagValue) {
         }
     }
 
+    public static void handleKeepDropAPMTraces(boolean keepAPMTrace) {
+        Span span = GlobalTracer.get().activeSpan();
+        String keepDropTag;
+        keepDropTag = keepAPMTrace ? DDTags.MANUAL_KEEP : DDTags.MANUAL_DROP;
+        span.setTag(keepDropTag, true);
```
shubham1g5, line 244: this can land us in a scenario where both `MANUAL_KEEP` and `MANUAL_DROP` are set to `true` which seems a bit error-prone. I would modify this code as - 
````
span.setTag(MANUAL_KEEP, keepAPMTrace)
span.setTag(MANUAL_DROP, !keepAPMTrace)
````

File: src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java [hunk 5]
```diff
@@ -213,26 +222,33 @@ public MenuSession buildSession(String username,
...
     }
 
     @Trace
     public MenuSession buildSession(SerializableMenuSession serializableMenuSession, FormplayerConfigEngine engine,
             CommCareSession commCareSession) throws Exception {
-        return new MenuSession(serializableMenuSession, engine, commCareSession, restoreFactory,
+        MenuSession menuSession =  new MenuSession(serializableMenuSession, engine, commCareSession, restoreFactory,
                 new FormplayerRemoteInstanceFetcher(caseSearchHelper, virtualDataInstanceService),
                 storageFactory);
+        FormplayerDatadog.handleKeepDropAPMTraces((boolean)menuSession.getMetaSessionContext().get(KEEP_APM_TRACES));
```
shubham1g5, line 242: any reason we can't do this from controllers itself instead of passing around `KEEP_APM_TRACES` uptill methods in this class ?

src/main/java/org/commcare/formplayer/util/FormplayerDatadog.java
```diff
@@ -234,4 +237,11 @@ private String getTagValueToSend(String tagName, String tagValue) {
         }
     }
 
+    public static void handleKeepDropAPMTraces(boolean keepAPMTrace) {
```
shubham1g5, line 240: this only drops the traces after they have been collected, a more optimised way here might be to disable the trace collection itself. I looked around and found this (have not tested) - 

````
public static void disableTracing() {
        Tracer tracer = GlobalTracer.get();
        if (tracer instanceof datadog.opentracing.DDTracer) {
            ((datadog.opentracing.DDTracer) tracer).close();
        }
    }
````

shubham1g5 (changes requested):
It feels to me that what we need here is more of a feature toggle rather than a request parameter specially since it's a domain level property. Curious if we have considered using a feature flag instead here - `FeatureFlagChecker.isToggleEnabled`

src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java
Robert-Costello: Do you mean that setting keep/drop on the span in the controller (or disabling the trace in the controller as you describe below) will mean it's not necessary to handle traces in individual methods because that trace won't propagate down? 
That sounds reasonable to me, but I'm not sure of exactly how the traces behave.

shubham1g5: I think whatever approach we take with regards to how to sample traces in Datadog, we don't need to do this at individual method level i.e. the DataDog settings should be global at-least in the scope of a single request. Therefore an optimum design here should be able to configure this only once without interferring with other business level details. 

I would think this kind of thing is best done with the help of a `@Aspect` annotated class ([Example](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/aspects/SetBrowserValuesAspect.java)). So something like below might be worth exploring - 

````
@Aspect
@Order(x)
public class SetTracingDisabledAspect {


    @Before(value = "@annotation(org.springframework.web.bind.annotation.RequestMapping)")
    public void setValues(JoinPoint joinPoint) throws Throwable {
        Object[] args = joinPoint.getArgs();
        if (args != null && args.length > 0 && args[0] instanceof AuthenticatedRequestBean) {
             final AuthenticatedRequestBean requestBean = (AuthenticatedRequestBean) args[0];
             if(!requestBean.keepAPMTraces()) { // OR  `FeatureFlagChecker.isToggleEnabled` if this comes from a feature flag
                  FormplayerDatadog.disableTraces();
             }
        }
    }
}
````

PR #1770, opened 2026-04-16
Pass case search table name as storage cache identifier (USH-6370 fix)

MartinRiese:
## Product Description

Pass caseSearchTableName to CaseInstanceTreeElement so case search results cache under "casedb:<table_hash>" instead of plain "casedb". Prevents RecordObjectCache key collision with the user's casedb instance when both share a QueryContext during nodeset evaluation.

## Technical Summary

https://dimagi.atlassian.net/browse/USH-6370

### Safety story

The change only effects a cache key. The code that populates the cache is unchanged.
The test script to reproduce the issue ran through multiple times without reproduction after this fix was deployed to staging.

### Automated test coverage

None

This PR requires [these changes in commcare-core](https://github.com/dimagi/commcare-core/pull/1535)

### Rollback instructions

### Review


src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
```diff
@@ -118,6 +121,9 @@ public AbstractTreeElement getExternalRoot(String instanceId, ExternalDataInstan
         if (caseSearchStorage.isStorageExists()) {
             // return root as CaseInstanceTreeElement
             InstanceBase instanceBase = new InstanceBase(instanceId);
+            if (FeatureFlagChecker.isToggleEnabled(TOGGLE_CASE_SEARCH_CACHE_KEY)) {
```
shubham1g5, line 124: is this still intended to be behind a toggle ?

src/test/java/org/commcare/formplayer/tests/CaseSearchResultsInStorageTests.java
```diff
@@ -223,6 +236,253 @@ public void testCaseClaimWithResultsInStorageWithAutoLaunch() throws Exception {
...
+     * RecordObjectCache key "casedb" causes the search results instance to read from the user's
+     * casedb instead of the search table, corrupting the entity display data. With the fix,
+     * unique cache keys per storage table prevent any cross-instance contamination.
+     *
+     * Verifies that:
+     * - entity data from instance('results') is correct (would be "" without fix)
+     * - entity data from the instance('casedb') cross-reference field is also correct
+     */
+    @Test
+    public void testCaseListWithCasedbCrossReferenceInSearchDetail() throws Exception {
```
shubham1g5, line 342: does this test fail today with FF off ?

shubham1g5 (approved):

src/main/java/org/commcare/formplayer/services/CaseSearchHelper.java
MartinRiese: Yes. Even though we don't not expect any performance issues, we want to run a test on production to make sure. I'll remove this with the following deployment.

src/test/java/org/commcare/formplayer/tests/CaseSearchResultsInStorageTests.java
MartinRiese: No. I removed it and clarified the test doc. If we could set conditions to force a collision it would not happen with the flag. But the test setup does not get us there. I tried running it 5000 time and no collision.

esoergel (approved):
Approving last 3 commits since Shubham's review

src/test/java/org/commcare/formplayer/tests/CaseSearchCacheCollisionTest.java
```diff
@@ -0,0 +1,124 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.util.HashMap;
+
+/**
+ * Reproduces USH-6370: CaseInstanceTreeElement.getStorageCacheName() returns "casedb"
+ * for ALL instances, causing RecordObjectCache collisions between the user's casedb
+ * and case search results when they share a QueryContext.
+ */
+public class CaseSearchCacheCollisionTest {
```
MartinRiese, line 28: We could improve these tests by extracting the tree element creation so the name gets set automatically using the functionality from the case search helper. That way we don't rely on setting it explicitly.

PR #1791, opened 2026-07-27
Public Webforms: Authentication

nospame:
## Technical Summary
[SAAS-19925](https://dimagi.atlassian.net/browse/SAAS-19925), [SAAS-19926](https://dimagi.atlassian.net/browse/SAAS-19926), [SAAS-19927](https://dimagi.atlassian.net/browse/SAAS-19927)

- First of three PRs adding Public Web Apps Sessions (one-time links that let an
  unauthenticated recipient fill a single pre-designated form). This PR is the authentication layer
  only — recognizing and authenticating a public session, inbound and outbound.
  - `CommCareSessionAuthFilter`: a request with the `CommCare-Public-Session: true` header **and** a
    `public_form_session_key` cookie is wrapped as a `PublicSessionCredential`.
  - `HqUserDetailsService`: public sessions send the key as `publicSessionKey` (not `sessionId`) to
    `session_details`.
  - `HqUserDetailsBean`: new HMAC-authoritative `public` field; `isAuthorized` drops the
    synthetic-username check for public sessions but still requires the request domain to match.
  - `PublicFormSessionAuth`: authenticates formplayer's outbound HQ calls with the session key,
    selected in `UserRestoreAspect#getHqAuth`.
- Design decision: the `CommCare-Public-Session` header is only a **credential-routing hint** —
  trust comes solely from HQ's `public` field, returned over the HMAC-authenticated `session_details`
  call.

Code and PR description written or co-written by AI and edited by human. Review by commit.

### Safety story
Purely additive and gated behind a new session type. The existing Django-cookie and HMAC auth paths
are untouched; the new branches only fire when the public header + cookie are present **and** HQ
confirms `public`. No public session can authenticate until HQ issues one-time links. No
existing data is read or written differently for non-public requests.

### Automated test coverage
- `SessionAuthTests` — credential routing, header-vs-cookie precedence, failure when the key is absent.
- `HqUserDetailsServiceTests` — `publicSessionKey` wire format vs `sessionId`.
- `HqUserDetailsTests` — domain-only `isAuthorized` for public sessions, `public` deserialization.
- `PublicFormSessionAuthTest`, `UserRestoreAspectTest` — outbound-auth selection.

### QA Plan
Public Webforms will get end-to-end QA before its release.

Though public form sessions are not yet being created in HQ, this PR should not be deployed without changes from #1792, which add required security checks to public form sessions.

### Rollback instructions

### Review

[SAAS-19925]: https://dimagi.atlassian.net/browse/SAAS-19925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SAAS-19926]: https://dimagi.atlassian.net/browse/SAAS-19926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SAAS-19927]: https://dimagi.atlassian.net/browse/SAAS-19927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/commcare/formplayer/auth/PublicFormSessionAuth.java
+++ b/src/main/java/org/commcare/formplayer/auth/PublicFormSessionAuth.java
@@ -2,7 +2,6 @@ package org.commcare.formplayer.auth;
 
 import org.commcare.formplayer.util.Constants;
 import org.springframework.http.HttpHeaders;
-import org.springframework.util.Assert;
 
 /**
  * {@link HqAuth} for a public web apps session.
@@ -16,7 +15,6 @@ public class PublicFormSessionAuth implements HqAuth {
     private final String sessionKey;
 
     public PublicFormSessionAuth(String sessionKey) {
-        Assert.hasText(sessionKey, "A public form session key is required");
         this.sessionKey = sessionKey;
     }
 
--- a/src/main/java/org/commcare/formplayer/auth/PublicFormSessionAuth.java
+++ b/src/main/java/org/commcare/formplayer/auth/PublicFormSessionAuth.java
@@ -1,0 +1,37 @@
+package org.commcare.formplayer.auth;
+
+import org.commcare.formplayer.util.Constants;
+import org.springframework.http.HttpHeaders;
+import org.springframework.util.Assert;
+
+/**
+ * {@link HqAuth} for a public web apps session.
+ *
+ * Emits the credential pair HQ requires to recognize a public session on its receiver/restore
+ * endpoints: the {@code public_form_session_key} cookie carrying the session key together with the
+ * {@code CommCare-Public-Session: true} header.
+ */
+public class PublicFormSessionAuth implements HqAuth {
+
+    private final String sessionKey;
+
+    public PublicFormSessionAuth(String sessionKey) {
+        Assert.hasText(sessionKey, "A public form session key is required");
+        this.sessionKey = sessionKey;
+    }
+
+    @Override
+    public HttpHeaders getAuthHeaders() {
+        return new HttpHeaders() {
+            {
+                add("Cookie", Constants.PUBLIC_FORM_SESSION_COOKIE_NAME + "=" + sessionKey);
+                add(Constants.PUBLIC_FORM_SESSION_HEADER, Constants.PUBLIC_FORM_SESSION_HEADER_VALUE);
+            }
+        };
+    }
+
+    @Override
+    public String toString() {
+        return "PublicFormSessionAuth";
+    }
+}
--- a/src/main/java/org/commcare/formplayer/services/HqUserDetailsService.java
+++ b/src/main/java/org/commcare/formplayer/services/HqUserDetailsService.java
@@ -1,7 +1,9 @@
 package org.commcare.formplayer.services;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.commcare.formplayer.auth.PublicSessionCredential;
 import org.commcare.formplayer.auth.UserDomainPreAuthPrincipal;
+import org.commcare.formplayer.beans.auth.HqPublicSessionKeyBean;
 import org.commcare.formplayer.beans.auth.HqSessionKeyBean;
 import org.commcare.formplayer.beans.auth.HqUserDetailsBean;
 import org.commcare.formplayer.exceptions.SessionAuthUnavailableException;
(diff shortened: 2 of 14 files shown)


kaapstorm (dismissed):

src/main/java/org/commcare/formplayer/services/HqUserDetailsService.java
```diff
@@ -73,9 +88,14 @@ private String getHmac(String data) throws Exception {
...
+        final Object credentials = token.getCredentials();
         try {
-            HqUserDetailsBean userDetails = getUserDetails(principal.getDomain(), sessionId);
+            HqUserDetailsBean userDetails;
+            if (credentials instanceof PublicSessionCredential publicCredential) {
+                userDetails = getPublicUserDetails(principal.getDomain(), publicCredential.getSessionKey());
+            } else {
+                userDetails = getUserDetails(principal.getDomain(), (String) credentials);
+            }
             if (!userDetails.isAuthorized(principal.getDomain(), principal.getUsername())) {
```
avazirna: Nit: I believe at this point `userDetails` already knows whether this is a public session, I'd suggested having a separate method to do the domain check, rather then inside`isAuthorized`. The signature entails checking whether a user is authorized to access a domain.

src/main/java/org/commcare/formplayer/auth/PublicFormSessionAuth.java
```diff
@@ -0,0 +1,37 @@
...
+ * Emits the credential pair HQ requires to recognize a public session on its receiver/restore
+ * endpoints: the {@code public_form_session_key} cookie carrying the session key together with the
+ * {@code CommCare-Public-Session: true} header.
+ */
+public class PublicFormSessionAuth implements HqAuth {
+
+    private final String sessionKey;
+
+    public PublicFormSessionAuth(String sessionKey) {
+        Assert.hasText(sessionKey, "A public form session key is required");
```
avazirna: Good check to have, but I'd move this before we instantiate HqAuth rather than after, unless there's a specific reason it needs to live here? Also, this throws an `IllegalArgumentException`, which likely won't surface as an auth failure, did you confirm it's caught and mapped to the right error path so the user sees an appropriate error message rather than a generic exception?

avazirna (commented):
@nospame Overall looks good to me, left a couple of comments.

src/main/java/org/commcare/formplayer/services/HqUserDetailsService.java
nospame: Yep, fair point [19bbafa](https://github.com/dimagi/formplayer/pull/1791/commits/19bbafa60bcac26d6d9404cbe7210b01ac4bf9db)

src/main/java/org/commcare/formplayer/auth/PublicFormSessionAuth.java
nospame: Good call, this seems like it was not the right spot. Moved to `loadUserDetails` where we already throw similar errors with messages. [ab59764](https://github.com/dimagi/formplayer/pull/1791/commits/ab59764eb68e438dd2f1dc15381aea058e710f40)

src/test/java/org/commcare/formplayer/auth/SessionAuthTests.java
```diff
@@ -97,6 +97,47 @@ public void testEndpoint_WithFullAuth_Succeeds() throws Exception {
...
+                .cookie(new Cookie(Constants.PUBLIC_FORM_SESSION_COOKIE_NAME, "public-key-abc"));
+        this.testEndpoint(builder, status().isForbidden());
+    }
+
+    /**
+     * When both signals are present, the explicit public header routes to the public credential
+     * rather than the Django session cookie.
+     */
+    @Test
+    public void testEndpoint_PublicHeaderAndSessionCookie_PrefersPublicCredential() throws Exception {
```
MartinRiese, line 131: I think the interesting test would be supply both but only public session is valid. will it fail auth or revert to public session?

MartinRiese (approved):
All those changes make sense to me. I am not well versed in the whole Formplayer auth though.

src/test/java/org/commcare/formplayer/auth/SessionAuthTests.java
nospame: The effect would be the same as this test - the public session credential is used and the django session credential is ignored. We don't even look at the django session cookie if the request appears to be a public session - which also means we'll have to do a good job of clearing the old public session token when a session is finished, on the HQ side, otherwise it would still take precedence over the potentially-valid django session. https://github.com/dimagi/formplayer/blob/ab59764eb68e438dd2f1dc15381aea058e710f40/src/main/java/org/commcare/formplayer/auth/CommCareSessionAuthFilter.java#L52-L56

PR #1792, opened 2026-07-27
Public Webforms: Deep-link handling and security

nospame:
## Product Description
Public Web Apps Sessions let someone open a one-time link and fill a single, pre-designated form
without an HQ account. This PR is where that behavior is confined: a public session can reach only
the form-entry routes it needs, and is locked to the exact app, endpoint, and identity HQ bound the
link to — it cannot navigate to any other app or form.

## Technical Summary
[SAAS-19928](https://dimagi.atlassian.net/browse/SAAS-19928), [SAAS-19929](https://dimagi.atlassian.net/browse/SAAS-19929), [SAAS-19930](https://dimagi.atlassian.net/browse/SAAS-19930)

- Second of three public webforms PRs in this repo; builds on #1791. Adds the confinement layer:
  - `WebSecurityConfig`: `restrictPublicSession(...)` default-deny allowlist on `/validate_form` and
    `anyRequest`. A public session may reach only `get_endpoint` + the in-form actions; `new-form`
    and `validate_form` are denied.
  - `get_endpoint`/`new-repeat`: `sessionid` cookie binding relaxed to `required=false` (public
    sessions carry no Django cookie).
  - `HqUserDetailsBean`: HQ-authoritative `app_build_id` + `endpoint_id`.
  - `PublicSessionLockAspect` (ordered ahead of all other aspects): overrides client-supplied app,
    endpoint, endpoint args, and identity with HQ's values; **fails closed** if they're missing.
  - `UserRestoreAspect#configureRestoreFactory`: a public session restores strictly as its
    HQ-authenticated user (no client username / restore-as / case id).
  - `MenuSessionRunnerService`: public sessions use endpoint navigation without the
    `SESSION_ENDPOINTS` toggle - rather than building out a new method of navigation or
    requiring the toggle to be enabled in HQ.

Code and PR description written or co-written by AI and edited by human. Review by commit.

### Safety story
The core safety property is **default-deny**: `restrictPublicSession` only ever *removes* access, and
only when the principal is a public session — regular sessions get exactly their prior decision. The
lock aspect and restore pinning both fail closed on missing authoritative values, and a public
session restores under its own synthetic user, so no real user's sandbox or identity is touched.
Aspect ordering (a subtle correctness requirement) is guarded by a dedicated weaving test after a
real ordering bug was caught there. Still inert in production until HQ issues links.

### Automated test coverage
- `SessionAuthTests` — allowlist: allowed routes not 403, `new-form`/`validate_form` denied, regular
  sessions unaffected.
- `PublicSessionLockAspectTest` — app/endpoint/identity override + fail-closed cases.
- `PublicSessionLockAspectWeavingTest` — proves the aspect actually weaves and is correctly ordered.
- `EndpointLaunchTest` — public launch without the `SESSION_ENDPOINTS` toggle.
- `HqUserDetailsTests` + test harness (`@WithHqUser` public flag).

### QA Plan
Public Webforms will get end-to-end QA before its release.

- **Cross-repo dependency:** relies on HQ's `session_details` returning `app_build_id` + `endpoint_id`
  for public sessions. Safe to deploy formplayer first (feature stays dark until HQ issues links), but
  the feature must not be enabled on HQ until this is deployed.

### Rollback instructions

### Review

[SAAS-19928]: https://dimagi.atlassian.net/browse/SAAS-19928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SAAS-19929]: https://dimagi.atlassian.net/browse/SAAS-19929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SAAS-19930]: https://dimagi.atlassian.net/browse/SAAS-19930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
+++ b/src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
@@ -1,5 +1,7 @@
 package org.commcare.formplayer.application;
 
+import org.commcare.formplayer.beans.auth.HqUserDetailsBean;
+import org.commcare.formplayer.exceptions.FormNotFoundException;
 import org.commcare.formplayer.objects.SerializableFormSession;
 import org.commcare.formplayer.services.FormDefinitionService;
 import org.commcare.formplayer.services.FormplayerRemoteInstanceFetcher;
@@ -8,12 +10,17 @@ import org.commcare.formplayer.services.MenuSessionRunnerService;
 import org.commcare.formplayer.services.RestoreFactory;
 import org.commcare.formplayer.services.VirtualDataInstanceService;
 import org.commcare.formplayer.session.FormSession;
+import org.commcare.formplayer.util.RequestUtils;
+import org.commcare.modern.database.TableBuilder;
 import org.commcare.session.CommCareSession;
 import org.javarosa.core.model.actions.FormSendCalloutHandler;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Component;
+
+import java.util.Objects;
+import java.util.Optional;
 
 @Component
 public class FormSessionFactory {
@@ -47,6 +54,7 @@ public class FormSessionFactory {
     @NotNull
     public FormSession getFormSession(SerializableFormSession serializableFormSession,
             @Nullable CommCareSession commCareSession, @Nullable String windowWidth) throws Exception {
+        verifyPublicSessionOwnership(serializableFormSession);
         FormplayerRemoteInstanceFetcher formplayerRemoteInstanceFetcher = new FormplayerRemoteInstanceFetcher(
                 runnerService.getCaseSearchHelper(),
                 virtualDataInstanceService);
@@ -59,5 +67,25 @@ public class FormSessionFactory {
                 formDefinitionService,
                 windowWidth
         );
+    }
+
+    /**
+     * A public web apps session may only operate on the form session that its own one-time link
+     * created. The session stores the scrubbed authoritative username, so any session bound to a
+     * different user/domain is rejected. No-op for non-public sessions and non-request contexts.
+     *
+     * Package-private for testing.
+     */
+    void verifyPublicSessionOwnership(SerializableFormSession session) {
+        Optional<HqUserDetailsBean> userDetails = RequestUtils.getUserDetails();
+        if (userDetails.isEmpty() || !userDetails.get().isPublicSession()) {
+            return;
+        }
+        HqUserDetailsBean details = userDetails.get();
+        boolean owned = Objects.equals(session.getDomain(), details.getDomain())
+                && Objects.equals(session.getUsername(), TableBuilder.scrubName(details.getUsername()));
+        if (!owned) {
+            throw new FormNotFoundException(session.getId());
+        }
     }
 }
--- a/src/main/java/org/commcare/formplayer/aspects/PublicSessionLockAspect.java
+++ b/src/main/java/org/commcare/formplayer/aspects/PublicSessionLockAspect.java
@@ -1,0 +1,80 @@
+package org.commcare.formplayer.aspects;
+
+import org.aspectj.lang.JoinPoint;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.annotation.Before;
+import org.commcare.formplayer.beans.InstallRequestBean;
+import org.commcare.formplayer.beans.SessionNavigationBean;
+import org.commcare.formplayer.beans.auth.HqUserDetailsBean;
+import org.commcare.formplayer.util.RequestUtils;
+import org.springframework.core.annotation.Order;
+import org.springframework.util.StringUtils;
+
+import java.util.Objects;
+import java.util.Optional;
+
+import lombok.extern.java.Log;
+
+/**
+ * Locks a public web apps session to the app, session endpoint, and identity that HQ bound its
+ * one-time link to. For a public session these values carried in the request are replaced with the
+ * HMAC-authenticated values from HQ's session_details response, so a recipient cannot navigate to
+ * any other app/form or key storage off an arbitrary user (all are otherwise client-supplied).
+ *
+ * Runs before {@link AppInstallAspect}, which keys the sandbox DB off the request's app id and
+ * username, so the storage factory, the MenuSession build, and endpoint navigation all see the
+ * authoritative values. Ordered ahead of every other formplayer aspect.
+ */
+@Aspect
+@Order(0)
+@Log
+public class PublicSessionLockAspect {
+
+    @Before(value = "@annotation(org.commcare.formplayer.annotations.AppInstall)")
+    public void lockToPublicApp(JoinPoint joinPoint) {
+        Optional<HqUserDetailsBean> userDetails = RequestUtils.getUserDetails();
+        if (userDetails.isEmpty() || !userDetails.get().isPublicSession()) {
+            return;
+        }
+        HqUserDetailsBean details = userDetails.get();
+        Object[] args = joinPoint.getArgs();
+        if (args.length == 0 || !(args[0] instanceof InstallRequestBean requestBean)) {
+            return;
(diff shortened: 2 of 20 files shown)


src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
```diff
@@ -47,6 +54,7 @@ public FormSession getFormSession(SerializableFormSession serializableFormSessio
     @NotNull
     public FormSession getFormSession(SerializableFormSession serializableFormSession,
             @Nullable CommCareSession commCareSession, @Nullable String windowWidth) throws Exception {
+        verifyPublicSessionOwnership(serializableFormSession);
```
kaapstorm, line 58: When a caller calls `getFormSession()` with two arguments, it calls `getCommCareSession()` before calling this line. `getCommCareSession()` loads the target menu session and runs `installService.configureApplication(...)` against a DB. A public session POSTing `/answer` with someone else's `sessionId` opens the victim's app DB, and if re-init throws, `InstallService` calls `sqliteDB.deleteDatabaseFile()` on it, before `FormNotFoundException`.

Move this check to the 2-arg entry point above (line 50) before `commCareSession` is set, or into an aspect.

src/main/java/org/commcare/formplayer/aspects/PublicSessionLockAspect.java
```diff
@@ -0,0 +1,84 @@
...
+ * Runs before {@link AppInstallAspect}, which keys the sandbox DB off the request's app id and
+ * username, so the storage factory, the MenuSession build, and endpoint navigation all see the
+ * authoritative values. Ordered ahead of every other formplayer aspect.
+ */
+@Aspect
+@Order(0)
+@Log
+public class PublicSessionLockAspect {
+
+    @Before(value = "@annotation(org.commcare.formplayer.annotations.AppInstall)")
```
kaapstorm, line 52: Identity pinning covers only `@AppInstall`. The other 12 allowlisted public routes (`answer`, `next_index`, `submit-all`, …) aren't `@AppInstall`, so `username`/`restoreAs` stay client-supplied and feed `LockAspect.getLockKeyForAuthenticatedBean`. A link holder can send `/answer` with `username: "realuser@domain"` and repeatedly hold that user's `FormplayerLockRegistry` lock, resulting in `LockError`/423 for the real user. Pin `username`/`domain`/`restoreAs` for every public request, not just installs.

src/main/java/org/commcare/formplayer/aspects/PublicSessionLockAspect.java
```diff
@@ -0,0 +1,84 @@
...
+            throw new IllegalStateException(
+                    "Public web apps session is missing an authoritative app id from HQ");
+        }
+        if (!Objects.equals(requestBean.getAppId(), details.getPublicAppId())) {
+            log.warning("Public session request app id did not match the authoritative value; "
+                    + "using the authoritative app id");
+        }
+        requestBean.setAppId(details.getPublicAppId());
+        requestBean.setUsername(details.getUsername());
+        requestBean.setDomain(details.getDomain());
```
kaapstorm: Add `requestBean.setPreview(false)`.

With default `preview: true`, `initFromArchive` skips the `jr://archive/<appId>/` cache and re-downloads the CCZ from HQ every request, and `MenuSession` skips `performTimedSync()`. A loop of `get_endpoint` with `preview:true` is CCZ-download amplification against HQ and enters the form with an unsynced sandbox.

src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
```diff
@@ -60,4 +68,24 @@ public FormSession getFormSession(SerializableFormSession serializableFormSessio
...
+
+    /**
+     * A public web apps session may only operate on the form session that its own one-time link
+     * created. The session stores the scrubbed authoritative username, so any session bound to a
+     * different user/domain is rejected. No-op for non-public sessions and non-request contexts.
+     *
+     * Package-private for testing.
+     */
+    void verifyPublicSessionOwnership(SerializableFormSession session) {
+        Optional<HqUserDetailsBean> userDetails = RequestUtils.getUserDetails();
```
kaapstorm, line 81: If HQ mints a shared public username per app/domain (which the bean's own `isAuthorized` comment implies — "no real HQ account, so the per-session username is not a meaningful check"), holder A can read/answer/submit holder B's session given its id, and all concurrent public sessions serialize on one `@UserLock` key. Worth confirming HQ issues a per-link username, or binding to the link's session key instead.

src/main/java/org/commcare/formplayer/aspects/PublicSessionLockAspect.java
```diff
@@ -0,0 +1,84 @@
...
+                    || (navigationBean.getEndpointArgs() != null
+                    && !navigationBean.getEndpointArgs().isEmpty());
+            if (clientDiffers) {
+                log.warning("Public session request endpoint/args did not match the authoritative "
+                        + "endpoint; using the authoritative endpoint with no args");
+            }
+            navigationBean.setEndpointId(details.getPublicEndpointId());
+            // Public sessions have an empty restore, so endpoint args (e.g. case ids) cannot
+            // resolve; the designated public endpoint must take no required arguments.
+            navigationBean.setEndpointArgs(null);
```
kaapstorm, line 97: Forcing `setEndpointArgs(null)` skips the `endpointArgs != null` block in `MenuSessionRunnerService.java:780-808`, which is where `Endpoint.populateEndpointArgumentsToEvaluationContext` validates missing/unexpected args. A public link bound to an endpoint with a required argument fails with an opaque XPath/500 instead of "Missing arguments: …". Consider validating the endpoint takes no required args and failing closed.

kaapstorm (commented):
I'm not familiar with Formplayer, but I am passing on findings from Claude.

src/main/java/org/commcare/formplayer/aspects/PublicSessionLockAspect.java
nospame: Yep, good point. Addressed here (along with setting preview: false as a related property) https://github.com/dimagi/formplayer/pull/1792/commits/eca1e14db20675ccee564100796a7f5cd92a514a

src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
nospame: This is a good callout. Added to the 2-arg version here, but also kept in the 3-arg version because that is called directly with `submit-all`. [d6942fb](https://github.com/dimagi/formplayer/pull/1792/commits/d6942fb51975fa2324c79d0a5bfd9e5fe53a9efc)

src/main/java/org/commcare/formplayer/application/FormSessionFactory.java
nospame: HQ doesn't mint a shared public username per app or domain, public session usernames are guaranteed unique, based on the model's own unique `id` field. https://github.com/dimagi/commcare-hq/blob/b766337a10696df02f2a3e89a6badc255099c404/corehq/apps/public_webforms/models.py#L147

src/main/java/org/commcare/formplayer/aspects/PublicSessionLockAspect.java
nospame: You're right this would have thrown an unhelpful XPath error - that is handled by the GlobalDefaultExceptionHandler and shouldn't 500, but it's not really what we want. Added validation here: https://github.com/dimagi/formplayer/pull/1792/commits/230343cfa691bfff0f034c6555d5609a2a5f3094

kaapstorm (approved):
Looks good to me.

